PR #272, opened 2023-09-14
Update stopevents_gen.py

MatheusFarias03:
Updated the python code using the enumerate function, which will eliminate the need for an explicit index variable "i". It also uses f-strings for cleaner string formatting.


akorotkov:
@MatheusFarias03, see the test failures.
You need to also adjust the line, which generates STOPEVENTS_COUNT macro.

PR #312, opened 2023-11-22
Supabase compat

pashkinelfe:
Orioledb sort compatibility & bump patches versions (that implement main burden of compatibility at PG side)


pashkinelfe:
Already merged into 15-16. 
(14- support is deprecated)

PR #374, opened 2024-10-21
feature: Put into S3 unchanged PostgreSQL files only

za-arthur:
### Hash files

`orioledb_data/pg_files.crc`, `orioledb_data/pg_small_files.crc` files contain hashes of changed and unchanged PostgreSQL files. To be able to read them easily by external tools, like `orioledb_s3_loader.py`, they are written as text files.
Each line of the files contains string in the following format:

```
FILE: <filename>, HASH: <hash>, CHECKPOINT: <checkpoint>
```

Where `<checkpoint>` is the latest checkpoint when the file was changed.

### S3 Workers

Each of S3 worker during processing a PG file checks its hash and accumulates them into a `orioledb_data/pg_files.crc.<worker_num>` temporary file.

### Checkpointer

The checkpointer (within `s3_perform_backup()`) waits until all PG files processed and merges all temporary hash files into the `orioledb_data/pg_files.crc` file.

The checkpointer separately processed small PG files, checks their hash and stores hashes into the `orioledb_data/pg_small_files.crc`.

Both `orioledb_data/pg_files.crc`, `orioledb_data/pg_small_files.crc` files are put into a S3 bucket.

### orioledb_s3_loader.py

`orioledb_s3_loader.py` can download unchanged files from previous checkpoints using `pg_files.crc`, `pg_small_files.crc` files.

## Tests

I added new tests into `s3_test.py`.

### DIFF
--- a/include/s3/checksum.h
+++ b/include/s3/checksum.h
@@ -1,0 +1,53 @@
+/*-------------------------------------------------------------------------
+ *
+ * checksum.h
+ * 		Declarations for calculating checksums of S3-specific data.
+ *
+ * Copyright (c) 2024, Oriole DB Inc.
+ *
+ * IDENTIFICATION
+ *	  contrib/orioledb/include/utils/checksum.h
+ *
+ *-------------------------------------------------------------------------
+ */
+#ifndef __S3_CHECKSUM_H__
+#define __S3_CHECKSUM_H__
+
+#include "utils/hsearch.h"
+
+#include "openssl/sha.h"
+
+#define O_SHA256_DIGEST_STRING_LENGTH	(SHA256_DIGEST_LENGTH * 2 + 1)
+
+typedef struct S3FileChecksum
+{
+	char		filename[MAXPGPATH];
+	char		checksum[O_SHA256_DIGEST_STRING_LENGTH];
+
+	bool		changed;		/* true if the checksum changed since last
+								 * checkpoint */
+	uint32		checkpointNumber;
+} S3FileChecksum;
+
+typedef struct S3ChecksumState
+{
+	HTAB	   *hashTable;
+	uint32		checkpointNumber;
+
+	S3FileChecksum *fileChecksums;	/* Buffer of S3FileChecksum entries */
+	uint32		fileChecksumsMaxLen;
+	uint32		fileChecksumsLen;
+} S3ChecksumState;
+
+extern S3ChecksumState *makeS3ChecksumState(uint32 checkpointNumber,
+											S3FileChecksum *fileChecksums,
+											uint32 fileChecksumsMaxLen,
+											const char *filename);
+extern void freeS3ChecksumState(S3ChecksumState *state);
+extern void flushS3ChecksumState(S3ChecksumState *state, const char *filename);
+
+extern S3FileChecksum *getS3FileChecksum(S3ChecksumState *state,
+										 const char *filename,
+										 Pointer data, uint64 size);
+
+#endif							/* __S3_CHECKSUM_H__ */
--- a/src/s3/checkpoint.c
+++ b/src/s3/checkpoint.c
@@ -17,7 +17,9 @@
 
 #include "checkpoint/checkpoint.h"
 #include "s3/checkpoint.h"
+#include "s3/checksum.h"
 #include "s3/headers.h"
+#include "s3/requests.h"
 #include "s3/worker.h"
 
 #include "utils/wait_event.h"
@@ -199,11 +201,16 @@ typedef struct
 	List	   *smallFileSizes;
 	int			smallFilesTotalSize;
 	int			smallFilesNum;
+	S3ChecksumState *checksumState;
 	uint32		chkpNum;
 } S3BackupState;
 
 #define SMALL_FILE_THRESHOLD		0x10000
 #define SMALL_FILES_TOTAL_THRESHOLD 0x100000
+
+#define FILE_CHECKSUMS_MAX_LEN	100
+#define SMALL_FILE_CHECKSUMS_FILENAME	ORIOLEDB_DATA_DIR "/small_file_checksums"
+#define SMALL_FILE_CHECKSUMS_TMP_FILENAME	SMALL_FILE_CHECKSUMS_FILENAME ".0"
 
 static S3TaskLocation flush_small_files(S3BackupState *state);
 static S3TaskLocation accumulate_small_file(S3BackupState *state,
@@ -228,6 +235,7 @@ s3_perform_backup(int flags, S3TaskLocation maxLocation)
 	StringInfoData tablespaceMapData;
 	ListCell   *lc;
 	tablespaceinfo *newti;
+	S3FileChecksum *fileChecksums;
 	S3TaskLocation location;
 
 	backup_started_in_recovery = RecoveryInProgress();
@@ -236,18 +244,29 @@ s3_perform_backup(int flags, S3TaskLocation maxLocation)
 
 	s3_headers_sync();
 
+	s3_workers_checkpoint_init();
+
+	/* Just in case delete a leftover file */
+	unlink(SMALL_FILE_CHECKSUMS_TMP_FILENAME);
+
 	initStringInfo(&tablespaceMapData);
 	state.tablespaces = get_tablespaces(&tablespaceMapData);
 
 	/* Add a node for the base directory at the end */
 	newti = palloc0(sizeof(tablespaceinfo));
 	newti->size = -1;
+
+	fileChecksums = palloc(sizeof(S3FileChecksum) * FILE_CHECKSUMS_MAX_LEN);
+
 	state.tablespaces = lappend(state.tablespaces, newti);
 	state.chkpNum = chkpNum;
 	state.smallFileNames = NIL;
 	state.smallFileSizes = NIL;
 	state.smallFilesTotalSize = sizeof(int);
 	state.smallFilesNum = 0;
+	state.checksumState = makeS3ChecksumState(chkpNum, fileChecksums,
+											  FILE_CHECKSUMS_MAX_LEN,
+											  SMALL_FILE_CHECKSUMS_FILENAME);
 
 	/* Send off our tablespaces one by one */
 	foreach(lc, state.tablespaces)
@@ -280,7 +299,21 @@ s3_perform_backup(int flags, S3TaskLocation maxLocation)
 	location = flush_small_files(&state);
 	maxLocation = Max(maxLocation, location);
 
+	freeS3ChecksumState(state.checksumState);
+	list_free_deep(state.tablespaces);
 	pfree(tablespaceMapData.data);
+	s3_queue_wait_for_location(maxLocation);
+
+	/* Wait until all S3 workers finish flushing and compact checksum files */
+	s3_workers_checkpoint_finish();
+
+	location = s3_schedule_file_write(chkpNum, FILE_CHECKSUMS_FILENAME, false);
+	maxLocation = Max(maxLocation, location);
+
+	durable_rename(SMALL_FILE_CHECKSUMS_TMP_FILENAME, SMALL_FILE_CHECKSUMS_FILENAME, ERROR);
+	location = s3_schedule_file_write(chkpNum, SMALL_FILE_CHECKSUMS_FILENAME, false);
+	maxLocation = Max(maxLocation, location);
+
 	s3_queue_wait_for_location(maxLocation);
 }
 
@@ -575,7 +608,7 @@ s3_backup_scan_dir(S3BackupState *state, const char *path,
 			if (statbuf.st_size < SMALL_FILE_THRESHOLD)
 				location = accumulate_small_file(state, pathbuf, statbuf.st_size);
 			else
-				location = s3_schedule_file_write(state->chkpNum, pathbuf, false);
+				location = s3_schedule_pg_file_write(state->chkpNum, pathbuf);
 			maxLocation = Max(maxLocation, location);
 		}
 		else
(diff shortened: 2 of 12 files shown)


src/s3/worker.c
```diff
@@ -26,6 +27,7 @@
 #include "postmaster/bgworker.h"
 #include "postmaster/bgwriter.h"
 #include "storage/bufmgr.h"
+#include "storage/condition_variable.h"
```
pashkinelfe: Looks like redundant include

za-arthur: Removed the include

src/s3/checkpoint.c
```diff
@@ -806,6 +838,36 @@ accumulate_small_file(S3BackupState *state, const char *path, int size)
...
+	uint64		dataSize;
+
+	/* First check if the file changed */
+	data = read_file(path, &dataSize);
+
+	if (data != NULL)
+	{
+		S3FileHash *entry;
+
+		if (state->hashState->pgFilesLen == state->hashState->pgFilesSize)
```
pashkinelfe: Propose rename: pgFilesSize -> bufferSize, pgFilesLen -> currentEntry (currentSize?)

src/s3/worker.c
```diff
@@ -89,6 +132,148 @@ register_s3worker(int num)
...
+
+		snprintf(worker_filename, sizeof(worker_filename), "%s.%d",
+				 PGFILES_CRC_FILENAME, i);
+
+		unlink(worker_filename);
+	}
+}
+
+static S3FileHash *
+get_worker_pgfiles(int worker_num)
```
pashkinelfe: int worker_num is not needed in signature. there is static worker_num

src/s3/worker.c
```diff
@@ -89,6 +132,148 @@ register_s3worker(int num)
...
+{
+	return workers_pgfiles + WORKERS_PGFILES_SIZE * worker_num;
+}
+
+static void
+flush_pgfiles_hash(int worker_num)
```
pashkinelfe: int worker_num is not needed in signature. there is static worker_num

src/s3/requests.c
```diff
@@ -498,11 +503,17 @@ s3_put_object_with_contents(char *objectname, Pointer data, uint64 dataSize,
 	char	   *tmp;
 	int			sc;
 	StringInfoData buf;
-	unsigned char hash[32];
 	long		http_code = 0;
 
-	(void) SHA256((unsigned char *) data, dataSize, hash);
-	contenthash = hex_string((Pointer) hash, sizeof(hash));
+	if (dataHash == NULL)
```
pashkinelfe: maybe if (dataHash == NULL && data != NULL) to not calculating hash of directories

include/s3/hash.h
```diff
@@ -0,0 +1,45 @@
+/*-------------------------------------------------------------------------
+ *
+ * hash.h
```
pashkinelfe: maybe checksum.h/checksum.c is better to split semantically hashing (HTAB) and file checksums ? The same is for variables: S3FileHash -> S3FileChecksum.

I'm not sure about it, because it's already called hash in the previous code. But e.g code of getS3PGFileHash is derived from SendBackupManifest, so it's worth to stick to "checksum-style" naming ( like checksumbuf, checksumstringbuf)

src/s3/hash.c
```diff
@@ -0,0 +1,245 @@
...
+	newEntry = (S3FileHash *) palloc0(sizeof(S3FileHash));
+
+	strlcpy(newEntry->filename, filename, sizeof(newEntry->filename));
+	strlcpy(newEntry->hash, hashstringbuf, sizeof(newEntry->hash));
+
+	newEntry->changed = ((prevEntry == NULL) ||
+						 (strncmp(prevEntry->hash, newEntry->hash, sizeof(newEntry->hash)) != 0)) ?
+		1 : 0;
+	newEntry->checkpointNumber = (newEntry->changed) ?
+		state->checkpointNumber : prevEntry->checkpointNumber;
```
pashkinelfe: Maybe split this block to if(prevEntry == NULL){} else {} to increase readability and to be clear that we never dereference NULL ptr (in fact we don't) ?

src/s3/hash.c
```diff
@@ -0,0 +1,245 @@
...
+		ereport(ERROR,
+				(errcode_for_file_access(),
+				 errmsg("could not write file \"%s\": %m",
+						filename)));
+
+	state->pgFilesLen = 0;
+}
+
+/*
+ * Check if a PGDATA file changed since last checkpoint and return PGFileHash.
```
pashkinelfe: and update S3HashState

src/s3/checkpoint.c
za-arthur: I think that they can be named similar to `StringInfoData`:
https://github.com/postgres/postgres/blob/REL_16_STABLE/src/include/lib/stringinfo.h#L40
i.e. in the case `pgFilesSize` might become `pgFilesMaxLen`.

include/s3/hash.h
za-arthur: Indeed. I agree that "hash" might sound a bit ambiguous. I'll try to use "checksum".

src/s3/requests.c
```diff
@@ -498,11 +503,17 @@ s3_put_object_with_contents(char *objectname, Pointer data, uint64 dataSize,
...
+	{
+		unsigned char hash[PG_SHA256_DIGEST_STRING_LENGTH];
```
pashkinelfe: Should be PG_SHA256_DIGEST_LENGTH not PG_SHA256_DIGEST_STRING_LENGTH
This can influence hex_string() as this size is transferred to it. (Maybe it'll fix valgrind test)

za-arthur:
Thank you @pashkinelfe for the review! I pushed new fixes and I merged changed from the `main` branch.

src/s3/checkpoint.c
za-arthur: I renamed it to `pgFilesMaxLen`.

src/s3/worker.c
za-arthur: Removed the argument.

src/s3/worker.c
za-arthur: Removed the argument.

src/s3/requests.c
za-arthur: For now I'm not sure that checking `data != NULL` is necessary, I left same behavior as it was before:
https://github.com/orioledb/orioledb/blob/3a75cb0bad0804f3aa6748c15f094bf34f4a93bd/src/s3/requests.c#L504
And my experiments show that `SHA256` returns a specific value if `data` is `NULL` and `dataSize` is `0`:
`e3b0c44298fc1c149afbf4c8996fb92427ae41e4649b934ca495991b7852b855`.

include/s3/hash.h
za-arthur: I renamed the files and the structs.

src/s3/hash.c
za-arthur: I rewrote to make it more readable.

src/s3/hash.c
za-arthur: Done

src/s3/requests.c
za-arthur: I replaced it by `SHA256_DIGEST_LENGTH` here and other places where `PG_SHA256_DIGEST_STRING_LENGTH` was used mistakenly.
I added `O_SHA256_DIGEST_STRING_LENGTH` to remove dependency from `common/sha2.h`. We anyway don't use PostgreSQL API to calculate checksum and therefore it would make sense to try to use `openssl` definitions.

src/s3/requests.c
pashkinelfe: Ok for same behavior as previous.
I also thought of setting SHA256(NULL) as defined constant, and encoded version as a constant too. Maybe apart from this PR.

src/s3/checksum.c
```diff
@@ -0,0 +1,254 @@
...
+
+	hash_destroy(state->hashTable);
+	pfree(state);
+}
+
+/*
+ * Initialize a hash table to store checksums of PGDATA files.
+ */
+static void
+initHashTable(S3ChecksumState *state, const char *filename)
```
pashkinelfe, line 69: Function not only inits but also inserts. Provided it's called only once, maybe make its code a part of makeS3ChecksumState.

src/s3/checksum.c
```diff
@@ -0,0 +1,254 @@
...
+#include "common/string.h"
+#include "storage/fd.h"
+#include "utils/hsearch.h"
+
+#include "openssl/sha.h"
+
+static void initHashTable(S3ChecksumState *state, const char *filename);
+
+/*
+ * Allocate S3ChecksumState and initalize hashTable by reading hashFilename.  A
```
pashkinelfe: ..and inserts hash entry..

src/s3/checksum.c
za-arthur: I think having `initHashTable()` as a separate function makes `makeS3ChecksumState()` a bit more clear and hides initialization details from it.
"init" here means not only that the hash table is created but also initialized with initial data.

src/s3/worker.c
```diff
@@ -89,6 +133,148 @@ register_s3worker(int num)
...
+
+		snprintf(worker_filename, sizeof(worker_filename), "%s.%d",
+				 PGFILES_CRC_FILENAME, i);
+
+		unlink(worker_filename);
+	}
+}
+
+static S3FileChecksum *
+get_worker_pgfiles(void)
```
pashkinelfe: One-line function called only once. Code could be inlined in the caller.

include/s3/checksum.h
```diff
@@ -0,0 +1,52 @@
...
+	bool		changed;		/* true if crc changed since last checkpoint */
+	uint32		checkpointNumber;
+} S3FileChecksum;
+
+typedef struct S3ChecksumState
+{
+	HTAB	   *hashTable;
+	uint32		checkpointNumber;
+
+	S3FileChecksum *pgFiles;		/* Buffer of S3FileChecksum entries */
```
pashkinelfe: Maybe renane pgFiles -> fileChecksums/pgFileChecksums
Same for pgFilesMaxLen, pgFilesLen,  WORKERS_PGFILES_MAX_LEN, pgFilesCnt, pgFilesFlushedCV, PGFILES_MAX_LEN, worker_pgfiles etc.

src/s3/checksum.c
```diff
@@ -0,0 +1,254 @@
...
+	if (ferror(file))
+		ereport(ERROR,
+				(errcode_for_file_access(),
+				 errmsg("could not read file \"%s\": %m", filename)));
+
+	FreeFile(file);
+}
+
+/*
+ * Save current workers' pgFiles hash table into a temporary file.
```
pashkinelfe: hash -> checksums

src/s3/requests.c
```diff
@@ -481,11 +483,13 @@ write_file(const char *filename, Pointer data, uint64 size)
...
  *
+ * If dataHash is NULL the function calculates hash of the content.
+ *
  * Returns HTTP status code.
  */
-static long
+long
 s3_put_object_with_contents(char *objectname, Pointer data, uint64 dataSize,
-							bool ifNoneMatch)
+							char *dataHash, bool ifNoneMatch)
```
pashkinelfe: dataHash -> dataChecksum ?

src/s3/requests.c
```diff
@@ -498,11 +502,17 @@ s3_put_object_with_contents(char *objectname, Pointer data, uint64 dataSize,
...
 	long		http_code = 0;
 
-	(void) SHA256((unsigned char *) data, dataSize, hash);
-	contenthash = hex_string((Pointer) hash, sizeof(hash));
+	if (dataHash == NULL)
+	{
+		unsigned char hash[SHA256_DIGEST_LENGTH];
+
+		(void) SHA256((unsigned char *) data, dataSize, hash);
+		contenthash = hex_string((Pointer) hash, sizeof(hash));
```
pashkinelfe: hash->checksumbuf
contenthash->checksumstringbuf (naming from SendBackupManifest())

src/s3/worker.c
```diff
@@ -700,14 +968,31 @@ s3worker_main(Datum main_arg)
...
+				workers_locations[worker_num] = taskLocation;
 				s3process_task(taskLocation);
-				workers_locations[num] = InvalidS3TaskLocation;
+				workers_locations[worker_num] = InvalidS3TaskLocation;
+			}
+
+			if (!pg_atomic_unlocked_test_flag(&workers_ctl->workersInProgress[worker_num]) &&
+				pg_atomic_read_u32(&workers_ctl->pgFilesCnt) == 0)
+			{
+				/* hash_state might be NULL if the worker restarted */
```
pashkinelfe: -> checksum_state

src/s3/checksum.c
```diff
@@ -0,0 +1,254 @@
...
+static void initHashTable(S3ChecksumState *state, const char *filename);
+
+/*
+ * Allocate S3ChecksumState and initalize hashTable by reading hashFilename.  A
+ * caller should prepare a buffer for S3FileChecksum entries, the caller is
+ * responsible to release the buffer.
+ */
+S3ChecksumState *
+makeS3ChecksumState(uint32 checkpointNumber, S3FileChecksum *pgFiles,
+					uint32 pgFilesMaxLen, const char *hashFilename)
```
pashkinelfe: -> checksumFilename

src/s3/requests.c
```diff
@@ -293,7 +294,7 @@ s3_delete_object(char *objectname)
 	char	   *tmp;
 	int			sc;
 	StringInfoData buf;
-	unsigned char hash[32];
+	unsigned char hash[SHA256_DIGEST_LENGTH];
```
pashkinelfe: -> checksumbuf

src/s3/requests.c
```diff
@@ -293,7 +294,7 @@ s3_delete_object(char *objectname)
...
 	char	   *contenthash;
```
pashkinelfe: checksumstring/checksumstringptr (?)

src/s3/worker.c
za-arthur: I think having a separate function improves readability, hides pointer arithmetic and adds possibility to use it later if necessary.

include/s3/checksum.h
za-arthur: Done

src/s3/requests.c
za-arthur: Renamed

PR #381, opened 2024-11-02
Docker/check improvements

ImreSamu:
(Updated version)

### Key changes:
- Added `orioledb-nightly` images with tags. ( alpine, ubuntu, debian ) ( from dockertest.yml )
- Implemented Docker-based test suite running:  `regresscheck isolationcheck testgrescheck`  ( alpine, ubuntu, debian )
- Enhanced `docker.yml` workflow; Full AMD64 image testing before push

Notes:
This PR contains substantial changes to demonstrate my vision for the project, particularly focusing on future extensibility (e.g., PostGIS support). 
I've taken a comprehensive approach to show the full potential, but I understand that:
- Some components might not be immediately necessary
- Alternative approaches may be preferred
- We can implement changes incrementally

So please feel free to:
- Select specific components for implementation
- Suggest alternative approaches
- Prioritize what's most valuable now

------------

#### 1. Docker Testing Strategy: Post-Build vs Build-Time

Chose "post-build testing" over "build-time testing" because:
- Tests run against final production image
- Better control over runtime capabilities ( `--privileged`, `--cap-add=` )
- Simpler debugging and development
- Avoids build-time cleanup complexities

Trade-off: Requires test dependency installation in test environment.

IMHO: Can switch to build-time testing if preferred.

####  2. Dockerfile Platform Support

- `Dockerfile.ubuntu` supports both Ubuntu and Debian builds - with minimal changes.
- Implemented dual parameter handling ("old" and "new" style)
- "Old style" parameters increase code complexity 

Decision needed:
- Keep backward compatibility OR Remove old parameter style ( ~ breaking change)

Need guidance on preferred approach.

####  3. OroleDB Regression Testing Strategy

Current implementation challenge:
- make regresscheck isolationcheck testgrescheck triggers reinstallation ( as I see .. )
- This conflicts with my need to test specific installations without modifications
I modified the `Makefile` to prevent reinstallation, though the solution may need review.

~ Seeking feedback on Makefile modifications.

------------

####  `.github/workflows/docker-nightly.yml`  ( old dockertest.yml )

- build 11 test combinations ( only amd64 )
- test with official docker-library check
- test with the orioledb official test targets ( ~ "regresscheck isolationcheck testgrescheck" )
- if refs/heads/main - then push orioledb-nightly:<tags>
  -  if secrets.DOCKERHUB_USERNAME exists --> push to docker registry
     - my test: https://hub.docker.com/r/imresamu/orioledb-nightly/tags
   - if secrets.GHCR_USERNAME exists --> push to ghcr.io ( github container registry ) 
      - my test: https://github.com/ImreSamu/orioledb/pkgs/container/orioledb-nightly

And the ~11 versions: (alpine, debian, ubuntu ) 
```
docker pull imresamu/orioledb-nightly:16-clang-alpine-3.21
docker pull imresamu/orioledb-nightly:16-clang-debian-bookworm-slim
docker pull imresamu/orioledb-nightly:16-clang-ubuntu-focal
docker pull imresamu/orioledb-nightly:16-clang-ubuntu-noble
docker pull imresamu/orioledb-nightly:17-clang-alpine-3.21
docker pull imresamu/orioledb-nightly:17-clang-alpine-edge
docker pull imresamu/orioledb-nightly:17-clang-debian-bookworm-slim
docker pull imresamu/orioledb-nightly:17-clang-ubuntu-focal
docker pull imresamu/orioledb-nightly:17-clang-ubuntu-noble
docker pull imresamu/orioledb-nightly:17-gcc-alpine-3.21
docker pull imresamu/orioledb-nightly:17-gcc-ubuntu-noble
```

#### `.github/workflows/docker.yml`

- test with official docker-library check
- test with the orioledb official test targets ( ~ "regresscheck isolationcheck testgrescheck" )

my test: https://hub.docker.com/r/imresamu/orioledb/tags

#### `Dockerfile`   ( alpine )

- adding alpine 3.21  as a default.
- adding new style parameters.
```
ARG BASE_IMAGE=alpine
ARG BASE_VERSION=3.21
```

#### `Dockerfile.ubuntu`   ( debian + ubuntu ) 

- add debian support.
- new style parameters
```
ARG BASE_IMAGE=ubuntu
ARG BASE_VERSION=noble
```


Dockerfile
```diff
@@ -1,11 +1,29 @@
...
-# set ALPINE_VERSION= [ edge 3.20 3.19 3.18 3.17 3.16 3.15 3.14 ]
-ARG ALPINE_VERSION=3.20
-FROM alpine:${ALPINE_VERSION}
+# Base image configuration
+# BASE_IMAGE=alpine
+# BASE_VERSION= [ edge 3.20 3.19 3.18 3.17 3.16 3.15 3.14 3.13]
+# New recommended way:
+#   docker build --build-arg BASE_IMAGE=alpine --build-arg BASE_VERSION=3.20 -t orioletest:alpine-3.20 .
+ARG BASE_IMAGE=alpine
+ARG BASE_VERSION=3.20
```
za-arthur: It seems that default values of variables are not used. Because of that the following command doesn't work anymore (it is mentioned [in the documentation)](https://github.com/orioledb/orioledb/blob/2cf58ec61ac0cba14212edc14ed567fdce51e0f1/doc/contributing/docker-builds.mdx?plain=1#L24):
```
docker build -t orioletest:17 --pull --network=host --progress=plain --build-arg PG_MAJOR="17" .

#0 building with "orbstack" instance using docker driver

#1 [internal] load build definition from Dockerfile
#1 transferring dockerfile: 11.76kB done
#1 WARN: InvalidDefaultArgInFrom: Default value for ARG ${BASE_IMAGE}:${BASE_VERSION} results in empty or invalid base image name (line 19)
#1 DONE 0.0s
Dockerfile:19
--------------------
  17 |     ARG BASE_VERSION=${ALPINE_VERSION:+${ALPINE_VERSION}}
  18 |
  19 | >>> FROM ${BASE_IMAGE}:${BASE_VERSION}
  20 |
  21 |     # Re-declare ARGs after FROM for use in subsequent instructions
--------------------
ERROR: failed to solve: failed to parse stage name ":": invalid reference format
```
Do you know if it is possible to use the default version?

za-arthur (commented):
Thank you for the another PR! I didn't check whole PR yet but noticed that default values for base image don't work anymore.

Makefile
```diff
@@ -196,21 +196,30 @@ endif
 
 PG_REGRESS_ARGS=--no-locale
 
-regresscheck: | install
+# Control installation dependency and wrapper based on NO_INSTALL
+# in the docker environment, we don't need to install the extension
+ifdef NO_INSTALL
```
za-arthur: I feel that `NO_INSTALL` adds additional complexity to the Makefile. Isn't it possible to run `make <target>` from docker containers with `install`?

ci/check_docker.sh
za-arthur, line 1: Is it necessary to have a separate `check_docker.sh` script which is essentially a wrapper around the `Makefile`? It seems it is possible to run `make` without the script.

The script installs dependencies. I think dependencies are already installed by the `Dockerfile`:
https://github.com/orioledb/orioledb/blob/2cf58ec61ac0cba14212edc14ed567fdce51e0f1/Dockerfile#L90-L124

But they are deleted afterwards:
https://github.com/orioledb/orioledb/blob/2cf58ec61ac0cba14212edc14ed567fdce51e0f1/Dockerfile#L215

Makefile
```diff
@@ -314,6 +323,21 @@ yapf:
 	yapf -i t/*.py
 	yapf -i *.py
 
+show-build-env:
```
za-arthur: Probably `show-build-env` might be unnecessary to have in the Makefile. I think version information can be get differently depending on the environment and OS, therefore might be good to avoid having environment specific commands in the Makefile.

za-arthur (commented):
@ImreSamu I added more comments below. My main concern would be that some of the changes add complexity. And it seems that during testing dependencies would be installed two times.

Dockerfile
ImreSamu: thank you - fixed!

Makefile
ImreSamu: For post-build testing, I need functionality similar to (postgresql)  `make installcheck` - specifically to run tests against an existing installation without triggering a new install.

My current approach to modifying the `Makefile` may not be optimal for this use case. Open to suggestions for better solutions.

ci/check_docker.sh
ImreSamu: > The script installs dependencies. I think dependencies are already installed by the Dockerfile:

I selected "post-build testing" over "build-time testing" - so I need to install .

Makefile
ImreSamu: > Probably show-build-env might be unnecessary to have in the Makefile. 

I added `show-build-env` to provide a no-op option in the CI/CD matrix for cases where we want to skip full regression testing (regresscheck isolationcheck testgrescheck) 
for specific Docker image variants. 
This was intended to prevent parameter-related errors when testing is disabled.

However, I agree this addition is unnecessary and can be removed.

ImreSamu:
@za-arthur :  Ready for review - with lot of changes.  

* new PR message !  ( with architectural decisions , ... )  Please read !  

--------------

My Main branch at https://github.com/ImreSamu/orioledb is now synchronized with the improvements from PR https://github.com/ImreSamu/orioledb/tree/docker/check_improvements_2024nov
and Docker images have been published to: 
- Docker Hub
   -  https://hub.docker.com/r/imresamu/orioledb/tags
   -  https://hub.docker.com/r/imresamu/orioledb-nightly/tags
- GitHub Container Registry (ghcr.io)
  -  https://github.com/ImreSamu/orioledb/pkgs/container/orioledb-nightly

Makefile
za-arthur: What if we rename `NO_INSTALL` into `IN_DOCKER`/`IS_DOCKER`? It think this might be more universal in future and it might change other variables as well.

Makefile
za-arthur: I think it isn't necessary to have this in the Makefile and can be removed.

ImreSamu: Ok ; I will remove.

Makefile
ImreSamu: imho; this functionality is valuable in several non-Docker scenarios; 
like verifying orioledb extensions installed through package managers (apt, yum, etc.)
or running regression tests against production installations.

Therefore, I'd suggest using a more general parameter name that better reflects its purpose.
* `TEST_INSTALLED=1`  ~ we're testing an already installed version
* `SKIP_INSTALL=1`
* `SYSTEM_TEST=1`
* `USE_EXISTING=1`
* `NO_BUILD_INSTALL=1`
* ....

akorotkov:
Hey @ImreSamu.
Thank you for the update. Is it ready for review?

ImreSamu:
A newer and improved version is available in https://github.com/orioledb/orioledb/pull/543
so I am closing this PR.

PR #386, opened 2024-11-11
Abstract page chunk API

za-arthur:
Add page chunk API to manipulate and read chunks, it allows to use different chunk implementations.

- `include/btree/chunk_ops.h` - defines abstract page chunk API
- `src/btree/tuple_chunk.c` - tuple chunks implementation of the chunk API for leaf, internal and hikey chunks
- `src/btree/tuple_chunk_test.c`, `test/sql/tuple_chunk.sql` - unit tests for tuple chunks implementation, needs https://github.com/orioledb/orioledb/pull/400 to be merged first

### DIFF
--- a/include/btree/chunk_ops.h
+++ b/include/btree/chunk_ops.h
@@ -1,0 +1,173 @@
+/*-------------------------------------------------------------------------
+ *
+ * chunk_ops.h
+ *		OrioleDB abstract page chunk access API.
+ *
+ * Copyright (c) 2025, Oriole DB Inc.
+ *
+ * IDENTIFICATION
+ *	  contrib/orioledb/include/btree/chunk_ops.h
+ *
+ *-------------------------------------------------------------------------
+ */
+#ifndef __BTREE_CHUNK_OPS_H__
+#define __BTREE_CHUNK_OPS_H__
+
+#include "btree/btree.h"
+#include "btree/page_contents.h"
+
+typedef enum BTreeChunkOperationType
+{
+	BTreeChunkOperationInsert,
+	BTreeChunkOperationUpdate,
+	BTreeChunkOperationDelete
+} BTreeChunkOperationType;
+
+typedef struct PartialPageState PartialPageState;
+
+typedef struct BTreeChunkOps BTreeChunkOps;
+
+typedef struct BTreeChunkBuilderItem
+{
+	Pointer		data;
+	uint16		size;
+	uint8		flags;
+} BTreeChunkBuilderItem;
+
+typedef struct BTreeChunkBuilder
+{
+	BTreeDescr *treeDesc;
+	const BTreeChunkOps *const ops;
+
+	/* Array of offsets of chunk items */
+	BTreeChunkBuilderItem chunkItems[BTREE_PAGE_MAX_CHUNK_ITEMS];
+	/* Number of chunk items */
+	uint16		chunkItemsCount;
+} BTreeChunkBuilder;
+
+typedef struct BTreeChunkDesc
+{
+	BTreeDescr *treeDesc;
+	const BTreeChunkOps *const ops;
+
+	/* Page which contains the chunk */
+	Page		page;
+	/* Offset of the chunk within the page */
+	OffsetNumber chunkOffset;
+} BTreeChunkDesc;
+
+typedef struct BTreeChunkOps
+{
+	uint16		itemHeaderSize;
+	BTreeKeyType itemKeyType;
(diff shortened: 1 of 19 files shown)


src/btree/tuple_chunk.c
```diff
@@ -0,0 +1,521 @@
...
+static uint16 get_tuple_chunk_header_size(BTreeChunkType chunkType);
+static uint16 get_tuple_chunk_tuple_size(BTreeDescr *treeDesc,
+										 BTreeChunkType chunkType, OTuple tuple);
+
+static void add_builder_item(BTreeChunkBuilderData *builderData,
+							 Pointer tupleHeader, uint16 tupleHeaderSize,
+							 OTuple tuple, uint16 tupleSize);
+
+int32
+tuple_chunk_estimate_change(BTreeChunkDesc *desc, OffsetNumber itemOffset,
```
akorotkov: We need to also estimate the size of the new BTreePageChunk.items element.

include/btree/chunk_ops.h
```diff
@@ -0,0 +1,129 @@
...
+	 * and shrinking of the chunk should be performed after "perform_change".
+	 */
+	void		(*perform_change) (BTreeChunkDesc *desc, OffsetNumber itemOffset,
+								   BTreeChunkOperationType operation,
+								   Pointer tupleHeader, OTuple tuple);
+
+	/*
+	 * Compare given key to the chunk tuple.
+	 */
+	OBTreeKeyCmp keycmp;
```
akorotkov: There should be a different signature. This function should compare a tuple stored in the chunk (at a given offset) to the given tuple.

include/btree/chunk_ops.h
```diff
@@ -0,0 +1,129 @@
...
+} BTreeChunkOperationType;
+
+typedef enum BTreeChunkType
+{
+	BTreeChunkHiKeys,
+	BTreeChunkNonLeaf,
+	BTreeChunkLeaf
+} BTreeChunkType;
+
+typedef struct BTreeChunkBuilderData
```
akorotkov: I think we need a pre-allocated array of items (like BTreePageChunk.items) here. First, fill them with offsets of tuples from the beginning of the first tuple. On build finish, advance them with the size of the items array itself.

src/btree/tuple_chunk.c
```diff
@@ -0,0 +1,521 @@
...
+{
+	builderData->treeDesc = treeDesc;
+	builderData->builderData = palloc(size);
+	builderData->builderMaxSize = size;
+	builderData->builderSize = 0;
+	builderData->chunkType = chunkType;
+}
+
+int32
+tuple_chunk_builder_estimate(BTreeChunkBuilderData *builderData,
```
akorotkov: Also, need to take into account the size of new items[] element.

include/btree/chunk_ops.h
```diff
@@ -0,0 +1,129 @@
...
+	 */
+	bool		(*search) (BTreeChunkDesc *desc, PartialPageState *partial,
+						   Page page, void *key, BTreeKeyType keyType,
+						   OffsetNumber *itemOffset);
+
+	/*
+	 * Read the tuple from the chunk.  Return false if the chunk was changed
+	 * concurrently.
+	 */
+	bool		(*read_tuple) (BTreeChunkDesc *desc, PartialPageState *partial,
```
akorotkov: We need another parameter, `bool *needsFree`, which indicates if the returned tuple is a separate copy that needs to be freed. In the current implementation, it should always be set to false. But in the future, we may have implementations that could return a tuple copy without copying a full chunk into a partial state.

src/btree/tuple_chunk.c
za-arthur: Thanks. I added the size of LocationIndex shift to the calculation.

include/btree/chunk_ops.h
za-arthur: Added a separate implementation of tuple comparison within the chunk.

include/btree/chunk_ops.h
za-arthur: Added the argument.

include/btree/chunk_ops.h
za-arthur: Added new fields `chunkItems` and `chunkItemsCount`.

src/btree/tuple_chunk.c
za-arthur: Fixed the function `tuple_chunk_builder_estimate`.

Issue #395, opened 2024-12-10
warning - s3 - alpine 


ImreSamu:
* alpine/v3.21 ;  clang19 (19.1.4-r0)
* the new : docker 17-clang-alpine-3.21 test cases 
   *  https://github.com/orioledb/orioledb/actions/runs/12245250718/job/34158755362

```
2024-12-09T22:18:59.7990392Z #12 336.3 src/s3/control.c:16:
2024-12-09T22:18:59.7991183Z #12 336.3 /usr/include/sys/fcntl.h:1:2: warning: redirecting incorrect #include <sys/fcntl.h> to <fcntl.h> [-W#warnings]
2024-12-09T22:18:59.7992138Z #12 336.3     1 | #warning redirecting incorrect #include <sys/fcntl.h> to <fcntl.h>
2024-12-09T22:18:59.7992691Z #12 336.3       |  ^
```

```
2024-12-09T22:19:00.6673525Z #12 337.0 In file included from src/s3/worker.c:42:
2024-12-09T22:19:00.6674484Z #12 337.0 /usr/include/sys/fcntl.h:1:2: warning: redirecting incorrect #include <sys/fcntl.h> to <fcntl.h> [-W#warnings]
2024-12-09T22:19:00.6675816Z #12 337.0     1 | #warning redirecting incorrect #include <sys/fcntl.h> to <fcntl.h>
2024-12-09T22:19:00.6676380Z #12 337.0       |  ^
```

or in the current  main [cfdf353](https://github.com/orioledb/orioledb/commit/cfdf353daece27313c5f3fc1cdd3626f302764b1) test
*  "[dockerTEST](https://github.com/orioledb/orioledb/actions/workflows/dockertest.yml)"   docker 17-clang-alpine-3.20

za-arthur:
Fixed by https://github.com/orioledb/orioledb/pull/397

PR #397, opened 2024-12-12
fix: use POSIX <fcntl.h> instead of <sys/fcntl.h> 

ImreSamu:
fix alpine warning ...  https://github.com/orioledb/orioledb/issues/395


za-arthur (approved):
Thank you for the fix. It looks good to me and tests pass.

PR #449, opened 2025-03-20
'index_bridging' option improving

homper:
- Added 'index_bridging' option to non-btree indices
- Not allowing changing 'index_bridging' option on the fly
- fixed order in all checks "(index->rd_rel->relam != BTREE_AM_OID || (options && options->index_bridging))"

### DIFF
--- a/include/indexam/handler.h
+++ b/include/indexam/handler.h
@@ -17,7 +17,6 @@
 
 extern IndexAmRoutine *orioledb_indexam_routine_hook(Oid tamoid,
 													 Oid amhandler);
-extern void o_add_bridged_option_to_ams(void);
 
 extern Relation o_current_index;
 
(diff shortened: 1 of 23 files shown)


akorotkov:
I see the pull request redefines options methods for all built-in AMs.  Would this work for an index AM defined in an extension?

homper:
Updated branch so now:
- Tables still have 'index_bridging' option which can be set during CREATE TABLE or automatically when first bridged index added
- Non-btree indices now don't have special options
- Btree index now have 'orioledb_index', which should be set to false to test btree bridged indices
- Not allowing changing 'index_bridging' and 'orioledb_index' options on the fly
---

Possible improvements and questions:
- Do different names for options needed
- Do we need to remove bridge when last bridged index removed
- Now I rebuild all indices during adding bridge, it could be optimized by rebuilding only primary and adding bridge_ctid as last column not as first/second one

PR #462, opened 2025-04-25
Enable postgresql regression tests in CI

e-ivkov:
Depends on https://github.com/orioledb/postgres/pull/13

For now the check job will fail, as we need to merge the changes to postgres fork first and tag them for 16 and 17 versions

### DIFF
--- a/.github/workflows/check.yml
+++ b/.github/workflows/check.yml
@@ -14,7 +14,7 @@ jobs:
         pg_version: [16, 17]
         compiler: [clang, gcc]
         check_type:
-          [normal, debug, sanitize, check_page, valgrind_1, valgrind_2]
+          [normal, debug, sanitize, check_page, valgrind_1, valgrind_2, pg_tests]
     env:
       LLVM_VER: 18
       CPU: ${{ matrix.cpu }}
--- a/ci/check.sh
+++ b/ci/check.sh
@@ -38,6 +38,20 @@ elif [ $CHECK_TYPE = "sanitize" ]; then
 	END
 	) \
 		make USE_PGXS=1 IS_DEV=1 installcheck -j $(nproc) || status=$?
+elif [ $CHECK_TYPE = "pg_tests" ]; then
+    cd ../postgresql
+    # Get float tests patch
+    wget -O float-patch.diff "https://git.postgresql.org/gitweb/?p=postgresql.git;a=patch;h=da83b1ea10c2b7937d4c9e922465321749c6785b"
+    git apply float-patch.diff
+    # Apply test setup SQL patches to enable OrioleDB
+    git apply src/test/regress/oriole-patches/test_setup_enable_oriole.diff
+    # Initialize data directory and set OrioleDB as default AM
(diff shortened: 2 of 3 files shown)


pashkinelfe (changes requested):
Should the file from https://git.postgresql.org/gitweb/?p=postgresql.git;a=patch;h=da83b1ea10c2b7937d4c9e922465321749c6785b has  *.patch name instead of diff? Does it apply to PG16/PG17 without conflicts?

Diff files from https://github.com/orioledb/postgres/pull/13/files#diff-2826bb004f96a894afaa97339f01b36facf9b5cc997e407dbda49afb42ddbc22 should have *.diff or *.patch extension (not *.sql).

ci/check_output.sh
```diff
@@ -15,7 +15,7 @@ if [ $oomcount != $oomsbefore ]; then
 fi
 
 # show diff if it exists
-for f in ` find ./orioledb/test -name regression.diffs ` ; do
+for f in ` find ./orioledb/test ./pgsql/test -name regression.diffs ` ; do
```
pashkinelfe, line 18: Will this work for all check types?

ci/check.sh
```diff
@@ -38,6 +38,23 @@ elif [ $CHECK_TYPE = "sanitize" ]; then
...
+    # Get float tests patch
+    wget -O float-patch.diff "https://git.postgresql.org/gitweb/?p=postgresql.git;a=patch;h=da83b1ea10c2b7937d4c9e922465321749c6785b"
+    git apply float-patch.diff
+    cd src/test/regress
+    # Apply test setup SQL patches to enable OrioleDB
+    git apply sql/oriole-patches/test_setup.sql
+    git apply expected/oriole-patches/test_setup.patch
+    # Initialize data directory and set OrioleDB as default AM
+    initdb --locale=C -D $GITHUB_WORKSPACE/pgsql/pgdata
+    sed -i "s/^#*default_table_access_method.*/default_table_access_method = 'orioledb'/" $GITHUB_WORKSPACE/pgsql/pgdata/postgresql.conf
```
pashkinelfe, line 52: Does this work? 
```
sed -i "s/^#*default_table_access_method.*/default_table_access_method = 'orioledb'/" postgresql.conf
sed: 1: "postgresql.conf": extra characters at the end of p command
```

.github/workflows/check.yml
```diff
@@ -14,7 +14,15 @@ jobs:
         pg_version: [16, 17]
         compiler: [clang, gcc]
         check_type:
-          [normal, debug, sanitize, check_page, valgrind_1, valgrind_2]
+          [
```
pashkinelfe: One-line was ok

ci/check_output.sh
e-ivkov: I assume yes, I just added another directory to the search, this directory seems to always exist

e-ivkov:
> Does it apply to PG16/PG17 without conflicts?

Yes, it applies cleanly to both patches16 and patches17 branches

pashkinelfe:
Pushed with minor changes

PR #505, opened 2025-07-11
Add tests of parallel and serial index recovery

za-arthur:
- Add additional `DEBUG1` messages to see parallel and serial index is recovery
- Add additional checks into `ReplicationTest.test_replication_table_rewrite` to check that index was rebuilt in parallel.
- Add test `ReplicationTest.test_replication_serial_index_build` to check that index was built with a single worker


akorotkov:
Already merged

PR #517, opened 2025-08-04
Add orioledb_zstd_version() function

ImreSamu:
Introduce a new SQL function 
- `orioledb_zstd_version()` 

that returns the version of the `zstd` library used by orioledb. 

This helps with diagnostics and compatibility checks when users 
or tests need to know the zstd compression library version.

The function returns the version string as text

Example:

```sql
SELECT orioledb_zstd_version();
 orioledb_zstd_version
-----------------------
 1.5.5
 ```


ImreSamu:
Ready for review.

-----

Motivation / Problem:

* I need this information for the Orioledb ZSTD-compression benchmark.
* PostGIS offers many similar functions; a comparable feature would be useful here.
   * `postgis_geos_version()` ; `postgis_geos_compiled_version()` ; `postgis_proj_version()` ; ....

Urgency:

* For me, this is not critical, not urgent, and unblocking ( so no rush ) 

Disclaimer:

* C programming isn’t my strong suit, so I may have misplaced some definitions.

pashkinelfe:
zstd in Postgres is used without checking it's version. The only check at build time is that it's over version 1.4.0
It relies on zstd, is conservative and backwards-compatible (Pg commit 479b69a4a551)
I propose to stick to this and not introducing extra SQL functions for this.

PR #551, opened 2025-09-04
Don't fsync relnode truncate and create for temp tables

e-ivkov:


akorotkov:
I think you could handle `add_undo_drop_relnode()` the same way.

e-ivkov:
add_undo_drop_relnode already calls add_undo_relnode with fsync=false:
```c
void
add_undo_drop_relnode(ORelOids oids, ORelOids *treeOids, int numTreeOids)
{
	ORelOids	invalid = {InvalidOid, InvalidOid, InvalidOid};

	Assert(ORelOidsIsValid(oids));
	add_undo_relnode(oids, treeOids, numTreeOids, invalid, NULL, 0, /*fsync=*/ false);
}
```

PR #555, opened 2025-09-08
Add specs test for assertion failed - non-NULL prevHiKey on LEFTMOST page - on sequential scanning while merge

GremSnoort:
**Problem**

Assertion failed `O_TUPLE_IS_NULL(prevHiKey)` for LEFTMOST internal page (means that there is a valid non-NULL previous high key for LEFTMOST internal page).
During a sequential scanning on btree internal pages (for example, on query: `DELETE FROM <table>`) invalid tuple location on a current loaded page triggers a next page loading which results in a sequence of loaded pages: `LEFTMOST -> LEFTMOST` or `middle -> LEFTMOST`. The next loaded page is LEFTMOST whereas there was a previous page.

**Root Cause**

Sequential scanning occurs concurrently with internal pages merge (on checkpoint or on eviction).
Model:
- backend 0 (seq scan): loads LEFTMOST page X
- backend 0 (seq scan): loads next page Y
- backend 1 (checkpoint or eviction): merges right page Y to the left page X
- backend 0 (seq scan): loads next page LEFTMOST X (merged) when there is a valid non-NULL previous high key from Y

**Solution**

Remove assert or fix it to allow prev high key on LEFTMOST.

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -132,6 +132,7 @@ REGRESSCHECKS = btree_sys_check \
 ISOLATIONCHECKS = bitmap_hist_scan \
 				  btree_iterate \
 				  btree_print_backend_id \
+				  btree_scan_on_merge \
 				  btree_scan \
 				  concurrent_update_delete \
 				  fkeys \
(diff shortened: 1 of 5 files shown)


akorotkov:
Merged

Issue #561, opened 2025-09-12
Missed TRUNCATE operation on replica


ereutt:
Hi!

I've catch a bug with TRUNCATE on replication

Steps to reproduce:
1. Startup master and replica
2. Master node:
- Create an orioledb extension on master
`CREATE EXTENSION orioledb;`
- Create table with explicit primary key and TAM orioledb
`CREATE TABLE test(data int primary key) USING orioledb;`
- Insert data into created table
`INSERT INTO test SELECT x FROM generate_series(1, 1000) x;`
3. Replica node:
- Ensure replica is synchronized by selecting data from table
`SELECT COUNT(*) FROM test;`
1000
4. Master node:
- Execute truncate operation
`TRUNCATE TABLE test;`
-Ensure truncate is done by selecting data from table
`SELECT COUNT(*) FROM test;`
0
5. Replica node:
- Ensure replica is synchronized
`SELECT pg_last_wal_replay_lsn();` it should be the same as master `SELECT pg_current_wal_lsn();`
- Check truncation result
`SELECT COUNT(*) FROM test;`
1000

Expected: Table is empty
Actual: Table containing data inserted on step 2

PR #562, opened 2025-09-12
fix missing truncate on replication

ereutt:
Fix issue #561 

There is no need synchronize with old refilenode on truncate table, because this synchronization cause index rebuilding from old refilenode containing truncated data.

### DIFF
--- a/src/tableam/handler.c
+++ b/src/tableam/handler.c
@@ -791,7 +791,7 @@ orioledb_relation_set_new_filenode(Relation rel,
 		 */
 		o_tables_drop_by_oids(old_oids, oxid, COMMITSEQNO_INPROGRESS);
 		o_tables_add(new_o_table, oxid, oSnapshot.csn);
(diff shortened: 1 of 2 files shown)


ereutt:
What I need to do in order to push my PR? Has anybody reviewed it? 
If I wrong I need feedback to correct my fix. 
I really want to help make Oriole more stable:)

akorotkov:
Hello, @ereutt!
Thank you so much for your PR.  I actually was going to rewrite it.
I think TRUNCATE needs its own WAL record.  This WAL record should facilitate the deletion of non-relevant relnodes on commit/abort (I haven't checked, but I believe your fix doesn't provide this).  If you feed power to re-implement it this way, please, go ahead.

akorotkov:
Sorry, my previous message seems wrong.  I think it's possible to fix the issue as you intended.
I've slightly modified it: 2077099971da.
I'll merge if CI tests pass.

PR #568, opened 2025-09-16
CI: increase 'check' timeout from 20 to 25 minutes

ImreSamu:
fixing: 'check (arm64, 17, gcc, debug) '  -  "Error: The action 'Check' has timed out after 20 minutes.''
https://github.com/orioledb/orioledb/actions/runs/17644307390/job/50138457788?pr=439#step:8:789

As the number of tests increases, 
it has become more common for the 'check' job to stop due to the 20-minute timeout. 
To avoid such premature failures, the limit is raised to 25 minutes. 
The valgrind case remains unchanged at 150 minutes.

### DIFF
--- a/.github/workflows/check.yml
+++ b/.github/workflows/check.yml
@@ -85,7 +85,7 @@ jobs:
       - name: Install post build prerequisites
         run: bash ./orioledb/ci/post_build_prerequisites.sh
       - name: Check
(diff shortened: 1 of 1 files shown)


akorotkov:
I don't think the timeout is the reason. Typically, tests pass in 6-7 minutes. More than 20 minutes is a clear anomaly.  One of the tests hangs.
BTW, your pull request is not the reason for that, for sure.

PR #593, opened 2025-10-08
Multiple fixes related to pg regression tests

homper:
- Fix testgres 1.11.1 compatibility
- Do not clean o_saved_relrewrite until the end of refresh mat view
- Fixes REINDEX SCHEMA for temp tables with ON COMMIT DELETE ROWS
- copied duplicates field from oIndex to OIndexDescr to make duplicate column logic less complex

### DIFF
--- a/src/catalog/ddl.c
+++ b/src/catalog/ddl.c
@@ -1183,8 +1183,7 @@ orioledb_utility_command(PlannedStmt *pstmt,
 			case REINDEX_OBJECT_SCHEMA:
 			case REINDEX_OBJECT_SYSTEM:
 			case REINDEX_OBJECT_DATABASE:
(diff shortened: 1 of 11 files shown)


src/catalog/ddl.c
```diff
@@ -514,6 +515,30 @@ check_multiple_tables(const char *objectName, ReindexObjectType objectKind, bool
 	 * by caller. At the same time do permission checks that need different
 	 * processing depending on the object type.
 	 */
+#if PG_VERSION_NUM >= 170006
```
pashkinelfe, line 518: Versions between 17000 and 17006 could not run with this commit.
Versions before 17000 can. So checking >= 17000 should be enough

PR #604, opened 2025-10-10
page inspection function

ereutt:
I would like suggest page inspection function analog of _bt_page_items_ from _pageinspect_ extension.
Currently it works only with "in memory" pages. In future we can add option to look into "on disk" pages as well.

### DIFF
--- a/include/recovery/wal.h
+++ b/include/recovery/wal.h
@@ -33,6 +33,11 @@
 #define WAL_REC_TRUNCATE	(13)
 #define WAL_REC_BRIDGE_ERASE (14)
 
+#define FIRST_WAL_VERSION (16)
+
+/* Bump it when WAL format changes */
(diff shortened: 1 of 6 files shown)


pashkinelfe (changes requested):
We'll probably need to bump Orioledb version if we add a new sql function e.g. to version 1.6 ```[sql/orioledb--1.5--1.6_prod.sql]```

I haven't reviewed the code, though.

homper:
Yes, it looks really cool as idea
Can this pr add some tests for function?
And in your opinion then, do we need to deprecate and remove orioledb_tbl_bin_structure?
Since we don't use it anywhere, but I was adding it with relatively similar intention
IMO, yes, conforming with pageinspect interface is better

homper:
We will look at this later, I think
Closing for now

Issue #588, opened 2025-10-06
`TRAP: failed Assert("!ctx->fast_forward")`


za-arthur:
Command to run:

```bash
make installcheck TEMP_CONFIG=/Users/artur/sources/supabase/orioledb/test/orioledb_tap.conf PG_TEST_INITDB_EXTRA_OPTS="--locale=C" PROVE_TESTS=t/006_logical_decoding.pl
```

`./orioledb_tap.conf` is:

```
shared_preload_libraries = 'orioledb'
default_table_access_method = 'orioledb'
orioledb.debug_disable_bgwriter = true
orioledb.use_sparse_files = true
checkpoint_timeout = 9000
max_wal_size = 5GB
```

From `tmp_check/log/regress_log_006_logical_decoding`:

```
[11:24:44.575](0.011s) ok 13 - logical slot was actually dropped with DB
psql:<stdin>:1: server closed the connection unexpectedly
	This probably means the server terminated abnormally
	before or while processing the request.
psql:<stdin>:1: error: connection to server was lost
[11:24:44.620](0.045s) not ok 14 - slot advancing with logical slot
```

From `tmp_check/log/006_logical_decoding_primary.log`:

```
TRAP: failed Assert("!ctx->fast_forward"), File: "logical.c", Line: 860, PID: 79867
0   postgres                            0x0000000102d2d088 ExceptionalCondition + 108
1   postgres                            0x0000000102b78790 change_cb_wrapper + 0
2   postgres                            0x0000000102b82864 ReorderBufferProcessTXN + 1004
3   postgres                            0x0000000102b802b8 ReorderBufferCommit + 156
4   orioledb.dylib                      0x0000000104c2ffc8 orioledb_decode + 1372
5   postgres                            0x0000000102b738bc LogicalDecodingProcessRecord + 128
6   postgres                            0x0000000102b785d4 LogicalSlotAdvanceAndCheckSnapState + 308
7   postgres                            0x0000000102b97194 pg_replication_slot_advance + 216
8   postgres                            0x0000000102a2002c ExecInterpExpr + 1468
9   postgres                            0x0000000102a591fc ExecResult + 324
10  postgres                            0x0000000102a29000 standard_ExecutorRun + 344
11  postgres                            0x0000000102bfb380 PortalRunSelect + 232
12  postgres                            0x0000000102bfafac PortalRun + 492
13  postgres                            0x0000000102bfa084 exec_simple_query + 1236
[...]

PR #616, opened 2025-10-20
Skip decoding of WAL records in fast_forward mode

za-arthur:
In fast_forward mode Postgres needs to skip doing unnecessary work and in some of its functions it has assert checks that `fast_forward` is `false`.

Issue: https://github.com/orioledb/orioledb/issues/588

### DIFF
--- a/src/recovery/logical.c
+++ b/src/recovery/logical.c
@@ -493,7 +493,6 @@ orioledb_decode(LogicalDecodingContext *ctx, XLogRecordBuffer *buf)
 	Pointer		ptr = startPtr;
 	OTableDescr *descr = NULL;
 	OIndexDescr *indexDescr = NULL;
-	int			sys_tree_num = -1;
 	ORelOids	cur_oids = {0, 0, 0};
 	OXid		oxid = InvalidOXid;
 	TransactionId logicalXid = InvalidTransactionId;
@@ -577,25 +576,46 @@ orioledb_decode(LogicalDecodingContext *ctx, XLogRecordBuffer *buf)
 				if (txn->toptxn)
 					txn = txn->toptxn;
 
-				dlist_foreach(cur_txn_i, &txn->subtxns)
+				if (rec_type == WAL_REC_COMMIT)
 				{
-					ReorderBufferTXN *cur_txn;
+					/*
+					 * SnapBuildXactNeedsSkip() does strict comparison.  So,
+					 * subtract endXLogPtr by one to fit snapshot just taken
+					 * after commit.
+					 */
+					if (SnapBuildXactNeedsSkip(ctx->snapshot_builder,
+											   endXLogPtr - 1) ||
+						ctx->fast_forward)
+					{
+						dlist_foreach(cur_txn_i, &txn->subtxns)
+						{
+							ReorderBufferTXN *cur_txn;
 
-					cur_txn = dlist_container(ReorderBufferTXN, node, cur_txn_i.cur);
+							cur_txn = dlist_container(ReorderBufferTXN, node, cur_txn_i.cur);
+							ReorderBufferForget(ctx->reorder, cur_txn->xid,
+												startXLogPtr);
+						}
+						ReorderBufferForget(ctx->reorder, txn->xid,
+											startXLogPtr);
 
-					ReorderBufferCommitChild(ctx->reorder, txn->xid, cur_txn->xid,
-											 startXLogPtr, endXLogPtr);
+						oxid = InvalidOXid;
+						logicalXid = InvalidTransactionId;
 
-				}
+						/* Skip processing of this transaction */
+						continue;
+					}
 
-				/*
-				 * SnapBuildXactNeedsSkip() does strict comparison.  So,
-				 * subtract endXLogPtr by one to fit snapshot just taken after
-				 * commit.
-				 */
-				if (rec_type == WAL_REC_COMMIT &&
-					!SnapBuildXactNeedsSkip(ctx->snapshot_builder, endXLogPtr - 1))
-				{
+					/*
+					 * Proceed to commit this transaction and subtransactions
+					 */
+					dlist_foreach(cur_txn_i, &txn->subtxns)
+					{
+						ReorderBufferTXN *cur_txn;
+
+						cur_txn = dlist_container(ReorderBufferTXN, node, cur_txn_i.cur);
+						ReorderBufferCommitChild(ctx->reorder, txn->xid, cur_txn->xid,
+												 startXLogPtr, endXLogPtr);
+					}
 					ReorderBufferCommit(ctx->reorder, logicalXid,
 										startXLogPtr, endXLogPtr,
 										0, XLogRecGetOrigin(buf->record),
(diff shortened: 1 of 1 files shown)


src/recovery/logical.c
```diff
@@ -423,6 +444,11 @@ orioledb_decode(LogicalDecodingContext *ctx, XLogRecordBuffer *buf)
 			ptr += sizeof(Oid);
 
 			elog(DEBUG4, "WAL_REC_RELATION");
+
+			/* Skip actual relation processing in fast_forward mode */
+			if (ctx->fast_forward)
+				continue;
```
pashkinelfe, line 703: Maybe it's worth to also invalidate cur_oids here?

za-arthur: It seems it isn't invalidated anywhere in the function anyway.

za-arthur: I'm not sure that it would be correct to invalidate it here, since it might be used later.

pashkinelfe: You are right that we don't invalidate cur_oids.
IMO all this works because we add WAL_REC_RELATION before WAL_REC_MODIFY around this place https://github.com/orioledb/orioledb/blob/f43fc6611ab414ca021c67137d5556f62a52011e/src/recovery/wal.c#L153 when necessary. However can't we skip this WAL_REC_RELATION because of fast_forward, but not skip WAL_REC_MODIFY (having cur_oids from previous relation)? Invalidation and checking that cur_oids are valid before processing WAL_REC_MODIFY could give some warranty against this scenario.

pashkinelfe: If fast_forward can't change for all duration of orioledb_decode, maybe invalidation of  oxid and logical_xid is also not needed in  fast_forward cases.

src/recovery/logical.c
```diff
@@ -328,6 +327,16 @@ orioledb_decode(LogicalDecodingContext *ctx, XLogRecordBuffer *buf)
 			if (!set_snapshot(ctx, logicalXid, changeXLogPtr))
 				continue;
 
+			/* Skip actual commit processing in fast_forward mode */
+			if (ctx->fast_forward)
+			{
+				UpdateDecodingStats(ctx);
```
pashkinelfe: UpdateDecodingStats() is not called from PG DecodeCommit() in case of fast_forward https://github.com/orioledb/postgres/blob/3a15cc3bd1c970a352082e35a7385150c368f117/src/backend/replication/logical/decode.c#L709

src/recovery/logical.c
pashkinelfe: I'd suggest using 
```
if (ctx->fast_forward)
          continue;
```
in all occurences in this PR (without invalidation of local vars and without calling UpdateDecodingStats)

za-arthur: > However can't we skip this WAL_REC_RELATION because of fast_forward, but not skip WAL_REC_MODIFY (having cur_oids from previous relation)? Invalidation and checking that cur_oids are valid before processing WAL_REC_MODIFY could give some warranty against this scenario.

Sorry I don't understand your point. Can you elaborate it?

pashkinelfe: > Sorry I don't understand your point. Can you elaborate it?

@za-arthur please just skip this comment. Based on my analysis in the more recent comments, fast_forward can't change for the duration of orioledb_decode and invalidating of all local variables is not needed. 

https://github.com/orioledb/orioledb/pull/616#discussion_r2447814457

https://github.com/orioledb/orioledb/pull/616#discussion_r2447799509

src/recovery/logical.c
za-arthur: Pushed changes

src/recovery/logical.c
```diff
@@ -577,32 +576,45 @@ orioledb_decode(LogicalDecodingContext *ctx, XLogRecordBuffer *buf)
...
-				if (rec_type == WAL_REC_COMMIT &&
-					!SnapBuildXactNeedsSkip(ctx->snapshot_builder, endXLogPtr - 1))
+				if (rec_type == WAL_REC_COMMIT)
 				{
+					/*
+					 * SnapBuildXactNeedsSkip() does strict comparison.  So,
+					 * subtract endXLogPtr by one to fit snapshot just taken
+					 * after commit.
+					 */
+					if (SnapBuildXactNeedsSkip(ctx->snapshot_builder,
```
pashkinelfe, line 586: Do we need check these additional conditions for skipping ?https://github.com/orioledb/postgres/blob/3a15cc3bd1c970a352082e35a7385150c368f117/src/backend/replication/logical/decode.c#L1302

src/recovery/logical.c
```diff
@@ -577,32 +576,45 @@ orioledb_decode(LogicalDecodingContext *ctx, XLogRecordBuffer *buf)
...
+											   endXLogPtr - 1) ||
+						ctx->fast_forward)
+					{
+						ReorderBufferForget(ctx->reorder, txn->xid,
```
pashkinelfe, line 598: In PG ReorderBufferForget is called in this state also for subxacts: https://github.com/orioledb/postgres/blob/3a15cc3bd1c970a352082e35a7385150c368f117/src/backend/replication/logical/decode.c#L705

src/recovery/logical.c
```diff
@@ -577,32 +576,45 @@ orioledb_decode(LogicalDecodingContext *ctx, XLogRecordBuffer *buf)
...
 					ReorderBufferCommit(ctx->reorder, logicalXid,
 										startXLogPtr, endXLogPtr,
 										0, XLogRecGetOrigin(buf->record),
 										buf->origptr);
 				}
 				else
 				{
+					Assert(rec_type == WAL_REC_ROLLBACK);
+
 					ReorderBufferAbort(ctx->reorder, logicalXid,
```
pashkinelfe, line 636: In PG ReorderBufferAbort is called in this state also for subxacts: https://github.com/orioledb/postgres/blob/3a15cc3bd1c970a352082e35a7385150c368f117/src/backend/replication/logical/decode.c#L874

src/recovery/logical.c
```diff
@@ -635,25 +647,28 @@ orioledb_decode(LogicalDecodingContext *ctx, XLogRecordBuffer *buf)
 			if (!set_snapshot(ctx, logicalXid, changeXLogPtr))
 				continue;
 
-			if (!SnapBuildXactNeedsSkip(ctx->snapshot_builder, endXLogPtr))
+			/* Skip actual commit processing */
+			if (SnapBuildXactNeedsSkip(ctx->snapshot_builder, endXLogPtr - 1) ||
+				ctx->fast_forward)
+			{
+				ReorderBufferAbort(ctx->reorder, logicalXid, startXLogPtr, 0);
```
pashkinelfe: Is right action here is ReorderBufferAbort or ReorderBufferForget ?

src/recovery/logical.c
za-arthur: Not sure if we should add all those conditions and it seems we don't have DB id, but maybe can parse it from `XLogReaderState`. BTW up-to-date version is:
https://github.com/orioledb/postgres/blob/patches17_14/src/backend/replication/logical/decode.c#L1312

src/recovery/logical.c
za-arthur: TBH I'm not sure how it should behave here and if we can use `logicalXid` instead of `xid`.

src/recovery/logical.c
za-arthur: I'd leave other cases for now and handle them later separately.

PR #627, opened 2025-10-22
GitHub issue 571: New mechanics for logical xid assignment and switch xid for correct logical decoding

GremSnoort:
New mechanics for logical xid assignment in a case when both Heap and Oriole apply changes within a single transaction (mixed transaction). For mixed transaction, Oriole's part acts as a sub-transaction for Heap.

New WAL record type (switch xid) added to determine a connection between Oriole's sub-transaction xid and a xid of the top Heap transaction.

Otherwise, without this connection, main transaction suddenly becomes splitted into two independent parts. From logical decoder's point of view this looks like two independent transactions but in fact internally related to each other. This situation outcomes in troubles for logical decoder with visibility of heap modifications in Oriole's sub-part due to incorrect state of the MVCC-historical snapshot.

### DIFF
--- a/include/recovery/wal.h
+++ b/include/recovery/wal.h
@@ -36,6 +36,20 @@
 #define WAL_REC_REPLAY_FEEDBACK	(16)
 
 /*
+ * Record type for a case when both heap and Oriole apply changes within a single transaction,
+ * so one logical xid is assigned by heap, and the other is assigned by Oriole.
+ *
+ * This record defines a connection between Oriole's sub-transaction xid and a xid of the top heap transaction
+ * which is needed for logical decoder.
+ *
+ * Otherwise, without this connection, main transaction suddenly becomes splitted into two independent parts.
+ * From logical decoder's point of view this looks like two independent transactions but in fact internally related to each other.
+ * This situation outcomes in troubles for logical decoder with visibility of heap modifications
+ * in Oriole's sub-part due to incorrect state of the MVCC-historical snapshot.
+ */
+#define WAL_REC_SWITCH_LOGICAL_XID	(17)
+
+/*
  * Value has been fixed at the moment of introducing WAL versioning.
  * Now, when we have WAL version in the beginning of each container
  * we should never change this value.
@@ -48,7 +62,7 @@
  * This is unlike ORIOLEDB_DATA_VERSION, ORIOLEDB_PAGE_VERSION and ORIOLEDB_COMPRESS_VERSION,
  * that make sense only inside one ORIOLEDB_BINARY_VERSION.
  */
-#define ORIOLEDB_WAL_VERSION (16)
+#define ORIOLEDB_WAL_VERSION (17)
 
 /* Constants for commitInProgressXlogLocation */
 #define OWalTmpCommitPos			(0)
@@ -68,7 +82,16 @@ typedef struct
 	uint8		recType;
 	uint8		oxid[sizeof(OXid)];
 	uint8		logicalXid[sizeof(TransactionId)];
+	/* Since ORIOLEDB_WAL_VERSION = 17 */
+	uint8		heapXid[sizeof(TransactionId)];
 } WALRecXid;
+
+typedef struct
+{
+	uint8		recType;
+	uint8		topXid[sizeof(TransactionId)];
+	uint8		subXid[sizeof(TransactionId)];
+} WALRecSwitchLogicalXid;
 
 typedef struct
 {
@@ -121,6 +144,8 @@ typedef struct
 {
 	uint8		recType;
 	uint8		parentSubid[sizeof(SubTransactionId)];
+	uint8		xmin[sizeof(OXid)]; /* Since ORIOLEDB_WAL_VERSION = 17 */
+	uint8		csn[sizeof(CommitSeqNo)];	/* Since ORIOLEDB_WAL_VERSION = 17 */
 } WALRecRollbackToSavepoint;
 
 typedef struct
@@ -156,11 +181,46 @@ typedef struct
 #define ORIOLEDB_WAL_PREFIX	"o_wal"
 #define ORIOLEDB_WAL_PREFIX_SIZE (5)
 
+extern const char *wal_record_type_to_string(int wal_record);
+
+/* API for parsing WAL-records */
+
+/* Parser for WAL_REC_XID */
+extern Pointer wal_parse_rec_xid(Pointer ptr, OXid *oxid, TransactionId *logicalXid, TransactionId *heapXid, uint16 wal_version);
+
+/* Parser for WAL_REC_COMMIT and WAL_REC_ROLLBACK */
+extern Pointer wal_parse_rec_finish(Pointer ptr, OXid *xmin, CommitSeqNo *csn);
+
+/* Parser for WAL_REC_JOINT_COMMIT */
+extern Pointer wal_parse_rec_joint_commit(Pointer ptr, TransactionId *xid, OXid *xmin, CommitSeqNo *csn);
+
+/* Parser for WAL_REC_RELATION */
+extern Pointer wal_parse_rec_relation(Pointer ptr, uint8 *treeType, ORelOids *oids);
+
+/* Parser for WAL_REC_O_TABLES_META_UNLOCK */
+extern Pointer wal_parse_rec_o_tables_meta_unlock(Pointer ptr, ORelOids *oids, Oid *old_relnode);
+
+/* Parser for WAL_REC_SAVEPOINT */
+extern Pointer wal_parse_rec_savepoint(Pointer ptr, SubTransactionId *parentSubid, TransactionId *logicalXid, TransactionId *parentLogicalXid);
+
+/* Parser for WAL_REC_ROLLBACK_TO_SAVEPOINT */
+extern Pointer wal_parse_rec_rollback_to_savepoint(Pointer ptr, SubTransactionId *parentSubid, OXid *xmin, CommitSeqNo *csn, uint16 wal_version);
+
+/* Parser for WAL_REC_TRUNCATE */
+extern Pointer wal_parse_rec_truncate(Pointer ptr, ORelOids *oids);
+
+/* Parser for WAL_REC_BRIDGE_ERASE */
+extern Pointer wal_parse_rec_bridge_erase(Pointer ptr, ItemPointerData *iptr);
+
+/* Parser for WAL_REC_SWITCH_LOGICAL_XID */
+extern Pointer wal_parse_rec_switch_logical_xid(Pointer ptr, TransactionId *topXid, TransactionId *subXid);
+
 extern void add_modify_wal_record(uint8 rec_type, BTreeDescr *desc,
 								  OTuple tuple, OffsetNumber length);
 extern void add_bridge_erase_wal_record(BTreeDescr *desc, ItemPointer iptr);
 extern void add_o_tables_meta_lock_wal_record(void);
 extern void add_o_tables_meta_unlock_wal_record(ORelOids oids, Oid oldRelnode);
+extern void add_switch_logical_xid_wal_record(TransactionId logicalXid_top, TransactionId logicalXid_sub);
 extern void add_savepoint_wal_record(SubTransactionId parentSubid,
 									 TransactionId prentLogicalXid);
 extern void add_rollback_to_savepoint_wal_record(SubTransactionId parentSubid);
--- a/include/transam/oxid.h
+++ b/include/transam/oxid.h
@@ -43,6 +43,15 @@ typedef struct
 
 extern XidMeta *xid_meta;
 
+typedef struct
+{
+	TransactionId xid;			/* a 32-bit transaction id to be used during
+								 * logical decoding */
+	bool		useHeap;		/* flag indicates if current logical xid was
+								 * allocated when heap xid has been already
+								 * set */
+} LogicalXidCtx;
+
 typedef struct OSnapshot
 {
 	CommitSeqNo csn;
@@ -84,6 +93,8 @@ o_check_isolation_level(void)
 
 #define XLOG_PTR_ALIGN(ptr) ((ptr) + ((ptr) & 1))
 
+extern void oxid_subxact_callback(SubXactEvent event, SubTransactionId mySubid, SubTransactionId parentSubid, void *arg);
+
 extern Size oxid_shmem_needs(void);
 extern void oxid_init_shmem(Pointer ptr, bool found);
 extern bool wait_for_oxid(OXid oxid, bool errorOk);
@@ -95,11 +106,12 @@ extern void assign_subtransaction_logical_xid(void);
 extern void set_oxid_csn(OXid oxid, CommitSeqNo csn);
 extern void set_oxid_xlog_ptr(OXid oxid, XLogRecPtr ptr);
 extern void set_current_oxid(OXid oxid);
-extern void set_current_logical_xid(TransactionId xid);
+extern void set_current_logical_xid(LogicalXidCtx *in);
 extern void parallel_worker_set_oxid(void);
 extern void reset_current_oxid(void);
 extern OXid get_current_oxid_if_any(void);
 extern TransactionId get_current_logical_xid(void);
+extern void get_current_logical_xid_ctx(LogicalXidCtx *output);
 extern void current_oxid_precommit(void);
 extern void current_oxid_xlog_precommit(void);
 extern void current_oxid_commit(CommitSeqNo csn);
--- a/include/transam/undo.h
+++ b/include/transam/undo.h
@@ -270,7 +270,7 @@ typedef struct
 	bool		has_retained_undo_location[(int) UndoLogsCount];
 	bool		local_wal_has_material_changes;
 	OXid		oxid;
-	TransactionId logicalXid;
+	LogicalXidCtx logicalXidContext;
 } OAutonomousTxState;
 
 /*
--- a/sql/orioledb--1.5--1.6_dev.sql
+++ b/sql/orioledb--1.5--1.6_dev.sql
@@ -1,0 +1,16 @@
+/* contrib/orioledb/sql/orioledb--1.5--1.6_dev.sql */
+
+-- complain if script is sourced in psql, rather than via CREATE EXTENSION
+\echo Use "ALTER EXTENSION orioledb UPDATE TO '1.6'" to load this file. \quit
+
+-- Get current logical xid to remember it
+CREATE FUNCTION orioledb_get_current_logical_xid()
+RETURNS int8
+AS 'MODULE_PATHNAME'
+VOLATILE LANGUAGE C;
+
+-- Get current heap xid to remember it
+CREATE FUNCTION orioledb_get_current_heap_xid()
+RETURNS int8
+AS 'MODULE_PATHNAME'
+VOLATILE LANGUAGE C;
--- a/src/recovery/logical.c
+++ b/src/recovery/logical.c
@@ -502,7 +502,8 @@ orioledb_decode(LogicalDecodingContext *ctx, XLogRecordBuffer *buf)
 	OIndexDescr *indexDescr = NULL;
 	ORelOids	cur_oids = {0, 0, 0};
 	OXid		oxid = InvalidOXid;
-	TransactionId logicalXid = InvalidTransactionId;
+	TransactionId logicalXid = InvalidTransactionId,
+				heapXid = InvalidTransactionId;
 	uint8		rec_type;
 	OIndexType	ix_type = oIndexInvalid;
 	TupleDescData *o_toast_tupDesc = NULL;
@@ -518,41 +519,61 @@ orioledb_decode(LogicalDecodingContext *ctx, XLogRecordBuffer *buf)
 		return;
 	}
 
+	elog(DEBUG4, "OrioleDB decode started startXLogPtr %X/%X endXLogPtr %X/%X", LSN_FORMAT_ARGS(startXLogPtr), LSN_FORMAT_ARGS(endXLogPtr));
+
 	while (ptr < endPtr)
 	{
+		const char *rec_type_str;
 		XLogRecPtr	changeXLogPtr = startXLogPtr + (ptr - startPtr);
 
 		rec_type = *ptr;
 		ptr++;
 
-		elog(DEBUG4, "RECEIVE: %s", rec_type == WAL_REC_XID ? "XID" :
-			 rec_type == WAL_REC_COMMIT ? "COMMIT" :
-			 rec_type == WAL_REC_ROLLBACK ? "ROLLBACK" :
-			 rec_type == WAL_REC_JOINT_COMMIT ? "JOINT COMMIT" :
-			 rec_type == WAL_REC_RELATION ? "RELATION" :
-			 rec_type == WAL_REC_O_TABLES_META_LOCK ? "META LOCK" :
-			 rec_type == WAL_REC_O_TABLES_META_UNLOCK ? "META_UNLOCK" :
-			 rec_type == WAL_REC_TRUNCATE ? "TRUNCATE" :
-			 rec_type == WAL_REC_SAVEPOINT ? "SAVEPOINT" :
-			 rec_type == WAL_REC_ROLLBACK_TO_SAVEPOINT ? "ROLLBACK TO SAVEPOINT" :
-			 rec_type == WAL_REC_INSERT ? "INSERT" :
-			 rec_type == WAL_REC_UPDATE ? "UPDATE" :
-			 rec_type == WAL_REC_DELETE ? "DELETE" :
-			 rec_type == WAL_REC_REINSERT ? "REINSERT" :
-			 rec_type == WAL_REC_REPLAY_FEEDBACK ? "FEEDBACK" : "_UNKNOWN");
+		rec_type_str = wal_record_type_to_string(rec_type);
+
+		elog(DEBUG4, "RECEIVE record type %d (%s)", rec_type, rec_type_str);
 
 		if (rec_type == WAL_REC_XID)
 		{
-			CSNSnapshotData *csnSnapshot;
+			ptr = wal_parse_rec_xid(ptr, &oxid, &logicalXid, &heapXid, wal_version);
 
-			memcpy(&oxid, ptr, sizeof(oxid));
-			ptr += sizeof(oxid);
+			if (TransactionIdIsValid(logicalXid))
+			{
+				CSNSnapshotData *csnSnapshot;
 
-			memcpy(&logicalXid, ptr, sizeof(TransactionId));
-			ptr += sizeof(TransactionId);
+				elog(DEBUG4, "RECEIVE record type %d (%s) oxid %lu logicalXId %u heapXid %u", rec_type, rec_type_str, oxid, logicalXid, heapXid);
 
-			csnSnapshot = SnapBuildGetCSNSnaphot(ctx->snapshot_builder);
-			csnSnapshot->nextXid = Max(csnSnapshot->nextXid, oxid);
+				csnSnapshot = SnapBuildGetCSNSnaphot(ctx->snapshot_builder);
+				csnSnapshot->nextXid = Max(csnSnapshot->nextXid, oxid);
+			}
+			else
+			{
+				/*
+				 * Oriole logical xid stays invalid for an autonomous
+				 * transactions (internal Oriole's mechanism intended for
+				 * independed commit of Oriole system catalog changes)
+				 */
+				elog(DEBUG4, "IGNORED record type %d (%s) invalid logicalXid for oxid %lu", rec_type, rec_type_str, oxid);
+			}
+		}
+		else if (rec_type == WAL_REC_SWITCH_LOGICAL_XID)
+		{
+			TransactionId topXid,
+						subXid;
+
+			ptr = wal_parse_rec_switch_logical_xid(ptr, &topXid, &subXid);
+
+			elog(DEBUG4, "RECEIVE record type %d (%s) %u=>%u oxid %lu logicalXId %u heapXid %u changeXLogPtr %X/%X",
+				 rec_type, rec_type_str,
+				 topXid, subXid,
+				 oxid, logicalXid, heapXid, LSN_FORMAT_ARGS(changeXLogPtr));
+
+			Assert(TransactionIdIsValid(topXid));
+			Assert(TransactionIdIsValid(subXid));
+
+			ReorderBufferAssignChild(ctx->reorder,
+									 topXid, subXid,
+									 changeXLogPtr);
 		}
 		else if (rec_type == WAL_REC_COMMIT || rec_type == WAL_REC_ROLLBACK)
 		{
@@ -562,11 +583,22 @@ orioledb_decode(LogicalDecodingContext *ctx, XLogRecordBuffer *buf)
 			CommitSeqNo csn;
 			CSNSnapshotData *csnSnapshot;
 
-			elog(DEBUG4, "WAL_REC_COMMIT");
-			memcpy(&xmin, ptr, sizeof(xmin));
-			ptr += sizeof(xmin);
-			memcpy(&csn, ptr, sizeof(csn));
-			ptr += sizeof(csn);
+			ptr = wal_parse_rec_finish(ptr, &xmin, &csn);
+
+			elog(DEBUG4, "RECEIVE record type %d (%s) oxid %lu logicalXId %u heapXid %u",
+				 rec_type, rec_type_str, oxid, logicalXid, heapXid);
+
+			if (!TransactionIdIsValid(logicalXid))
+			{
+				/*
+				 * Oriole logical xid stays invalid for an autonomous
+				 * transactions
+				 */
+				elog(DEBUG4, "IGNORED record type %d (%s) invalid logicalXid for oxid %lu", rec_type, rec_type_str, oxid);
+
+				UpdateDecodingStats(ctx);
+				continue;
+			}
 
 			csnSnapshot = SnapBuildGetCSNSnaphot(ctx->snapshot_builder);
 			csnSnapshot->snapshotcsn = csn;
@@ -578,11 +610,16 @@ orioledb_decode(LogicalDecodingContext *ctx, XLogRecordBuffer *buf)
 
 			txn = get_reorder_buffer_txn(ctx->reorder, logicalXid);
 
+			if (txn == NULL)
+			{
+				elog(DEBUG4, "RECEIVE record type %d (%s) oxid %lu logicalXId %u :: unknown transaction, nothing to replay",
+					 rec_type, rec_type_str, oxid, logicalXid);
+			}
+
 			/* unknown transaction, nothing to replay */
 			if (txn != NULL)
 			{
-				if (txn->toptxn)
-					txn = txn->toptxn;
+				Assert(TransactionIdIsValid(logicalXid));
 
 				if (rec_type == WAL_REC_COMMIT)
 				{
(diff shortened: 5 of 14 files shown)


sql/orioledb--1.4--1.5_prod.sql
```diff
@@ -27,6 +27,18 @@ RETURNS bigint
 AS 'MODULE_PATHNAME'
 VOLATILE LANGUAGE C;
 
+-- Get current logical xid to remember it
```
pashkinelfe: Could you please move these sql functions to the newly created orioledb--1.5--1.6_prod.sql [1] or orioledb--1.5--1.6_dev.sql [2] and bump current sql version to 1.6?
That's because we already have released Oriole release with version 1.4 and can't add new sql functions to existing releases already in use.
If these sql functions are intended only for testing and not to end-users, please use [2] otherwise [1]

src/rewind/rewind.c
```diff
@@ -89,6 +91,18 @@ orioledb_get_current_oxid(PG_FUNCTION_ARGS)
 	PG_RETURN_DATUM(get_current_oxid());
 }
 
+Datum
```
pashkinelfe: Could you please move these functions to oxid.c or something like that? It's because rewind is a separate narrow feature and your changes are more universal and not connected to rewind. If you use some of the functions from rewind.c in general replication tests, please feel free also to move them outside this file.

include/recovery/wal.h
```diff
@@ -25,21 +25,24 @@
...
+#define WAL_REC_TRUNCATE				(13)
+#define WAL_REC_BRIDGE_ERASE			(14)
+#define WAL_REC_REINSERT				(15)	/* UPDATE with changed pkey
+												 * represented as DELETE +
+												 * INSERT in OrioleDB but
+												 * externally exported as an
+												 * UPDATE in logical decoding */
+#define WAL_REC_SWITCH_LOGICAL_XID		(16)
+
+#define FIRST_WAL_VERSION (17)
```
pashkinelfe: FIRST_WAL_VERSION should never be changed since https://github.com/orioledb/orioledb/commit/51e6f150040eaefe19c95752cc149186b72ebdf5

include/recovery/wal.h
```diff
@@ -25,21 +25,24 @@
...
 
 /* Bump it when WAL format changes */
-#define CURRENT_WAL_VERSION (16)
+#define CURRENT_WAL_VERSION (17)
```
pashkinelfe: Please don't bump CURRENT_WAL_VERSION in this commit. Bumping is needed iff we already released previous version. There were no release with version 16 yet. We'd better leave CURRENT_WAL_VERSION (16)

include/recovery/wal.h
GremSnoort: restored `FIRST_WAL_VERSION` back to 16

include/recovery/wal.h
GremSnoort: restored `CURRENT_WAL_VERSION` back to 16

sql/orioledb--1.4--1.5_prod.sql
GremSnoort: new functions moved to a new file `orioledb--1.5--1.6_dev.sql`

src/rewind/rewind.c
GremSnoort: new added sql functions related to xids are moved to a file `transam/oxid.c`

src/recovery/logical.c
```diff
@@ -513,38 +517,56 @@ orioledb_decode(LogicalDecodingContext *ctx, XLogRecordBuffer *buf)
...
+
+			elog(DEBUG4, "RECEIVE record type %d (%s) %u=>%u oxid %lu logicalXId %u heapXid %u",
+				 rec_type, rec_type_str,
+				 oldXid, newXid,
+				 oxid, logicalXid, heapXid);
+
+			Assert(TransactionIdIsValid(oldXid));
+			Assert(TransactionIdIsValid(newXid));
+			Assert(logicalXid == oldXid);
+			Assert(heapXid == newXid);
```
akorotkov: I don't think this makes sense.  WAL_REC_SWITCH_LOGICAL_XID is especially for the case we didn't know heapXid in advance.

src/recovery/logical.c
```diff
@@ -513,38 +517,56 @@ orioledb_decode(LogicalDecodingContext *ctx, XLogRecordBuffer *buf)
...
+
+			ReorderBufferAssignChild(ctx->reorder,
+									 heapXid, logicalXid,
+									 buf->origptr);
+
+			logicalXid = newXid;
```
akorotkov: I think we also must set heapXid = newXid;

sql/orioledb--1.5--1.6_dev.sql
```diff
@@ -0,0 +1,16 @@
...
+\echo Use "ALTER EXTENSION orioledb UPDATE TO '1.6'" to load this file. \quit
+
+-- Get current logical xid to remember it
+CREATE FUNCTION orioledb_get_current_logical_xid()
+RETURNS int
+AS 'MODULE_PATHNAME'
+VOLATILE LANGUAGE C;
+
+-- Get current heap xid to remember it
+CREATE FUNCTION orioledb_get_current_heap_xid()
```
akorotkov, line 13: Why not just use select pg_current_xact_id_if_assigned()::xid ?

test/t/logical_xid_subxacts_test.py
```diff
@@ -0,0 +1,220 @@
...
+            CREATE TABLE {self.newtable} (a int PRIMARY KEY, b text);
+            {self.check_xact}
+        """)
+
+		con.commit()
+
+		self.assertEqual(output[0][0], '"0"/"0"')
+		self.assertEqual(output[2][0], '"0"/"0"')
+		self.assertEqual(output[3][0], '"0"/"0"')
+		self.assertEqual(output[4][0], '"740"/"0"')
```
akorotkov: Hardcoding particular numeric values looks fragile. Can we just assume any non-zero value instead?

sql/orioledb--1.5--1.6_dev.sql
GremSnoort: `pg_current_xact_id_if_assigned` returns top transaction ID, but for check sub-transactions we need to know current transaction ID

include/recovery/wal.h
```diff
@@ -63,8 +64,16 @@ typedef struct
 	uint8		recType;
 	uint8		oxid[sizeof(OXid)];
 	uint8		logicalXid[sizeof(TransactionId)];
+	uint8		heapXid[sizeof(TransactionId)];
```
pashkinelfe, line 86: Can break compatibility with existing WAL without heapXid in WALRecXid. We need to bump CURRENT_WAL_VERSION with this change and add a conversion routine for reading WALRecXid from WAL of a previous version without this change (16) https://github.com/orioledb/orioledb/releases/tag/beta13

include/recovery/wal.h
```diff
@@ -63,8 +64,16 @@ typedef struct
...
 } WALRecXid;
 
+typedef struct
+{
+	uint8		recType;
+	uint8		topXid[sizeof(TransactionId)];
+	uint8		subXid[sizeof(TransactionId)];
+} WALRecSwitchLogicalXid;
```
pashkinelfe, line 94: Adding this needs bumping ORIOLEDB_WAL_VERSION from the currently released version (16)  (No conversion is needed for reading newly added WALRec types) . https://github.com/orioledb/orioledb/releases/tag/beta13

pashkinelfe (changes requested):

.github/workflows/check.yml
```diff
@@ -25,8 +25,7 @@ jobs:
               pg_version = [16, 17],
               compiler   = ["clang", "gcc"],
               check_type = [
-                  "normal", "debug", "sanitize", "check_page",
-                  "valgrind_1", "valgrind_2", "pg_tests",
+                  "pg_tests",
```
pashkinelfe: Revert this temporary change before merge

GremSnoort: this temporary change was reverted

src/recovery/logical.c
GremSnoort: This change is outdated. Currently, on switch xid record, simply reads topXid and subXid, then checks that they are both valid and establishes a connection between them using ReorderBufferAssignChild.

src/recovery/logical.c
GremSnoort: This change is outdated. Currently, on switch xid record, simply reads topXid and subXid, then checks that they are both valid and establishes a connection between them using ReorderBufferAssignChild.

include/recovery/wal.h
GremSnoort: `ORIOLEDB_WAL_VERSION` set up to 17

src/recovery/logical.c
```diff
@@ -765,27 +828,15 @@ orioledb_decode(LogicalDecodingContext *ctx, XLogRecordBuffer *buf)
...
 		{
 			ORelOids	oids;
 
-			memcpy(&oids.datoid, ptr, sizeof(Oid));
-			ptr += sizeof(Oid);
-			memcpy(&oids.reloid, ptr, sizeof(Oid));
-			ptr += sizeof(Oid);
-			memcpy(&oids.relnode, ptr, sizeof(Oid));
-			ptr += sizeof(Oid);
+			ptr = WAL_PARSE_REC_TRUNCATE(ptr, oids);
```
pashkinelfe: I propose to call wal_parse_rec_truncate directly and not define macro WAL_PARSE_REC_TRUNCATE. (I.e. doing similar to calling wal_parse_rec_savepoint, wal_parse_rec_rollback_to_savepoint etc.)

src/recovery/wal.c
```diff
@@ -45,6 +46,192 @@ static void add_modify_wal_record_extended(uint8 rec_type, BTreeDescr *desc,
...
+}
+
+/* Parser for WAL_REC_ROLLBACK_TO_SAVEPOINT */
+Pointer
+wal_parse_rec_rollback_to_savepoint(Pointer ptr, SubTransactionId *parentSubid, OXid *xmin, CommitSeqNo *csn)
+{
+	Assert(ptr);
+
+	PARSE(ptr, parentSubid);
+	PARSE(ptr, xmin);
```
pashkinelfe: Should be done only if (ORIOLEDB_WAL_VERSION >=17). Otherwise, this breaks the whole container parsing for ORIOLEDB_WAL_VERSION == 16

src/recovery/wal.c
```diff
@@ -45,6 +46,192 @@ static void add_modify_wal_record_extended(uint8 rec_type, BTreeDescr *desc,
...
+	PARSE(ptr, csn);
```
pashkinelfe: Should be done only if (ORIOLEDB_WAL_VERSION >=17) Otherwise, this breaks the whole container parsing for ORIOLEDB_WAL_VERSION == 16

include/recovery/wal.h
```diff
@@ -121,6 +144,10 @@ typedef struct
 {
 	uint8		recType;
 	uint8		parentSubid[sizeof(SubTransactionId)];
+#if ORIOLEDB_WAL_VERSION >= (17)
```
za-arthur: I'm not sure if it is beneficial to have this condition since I don't think the code will be compiled with `ORIOLEDB_WAL_VERSION = 16`.

include/recovery/wal.h
```diff
@@ -156,11 +183,51 @@ typedef struct
...
+extern Pointer wal_parse_rec_bridge_erase(Pointer ptr, ItemPointerData *iptr);
+
+/* Parser for WAL_REC_SWITCH_LOGICAL_XID */
+extern Pointer wal_parse_rec_switch_logical_xid(Pointer ptr, TransactionId *topXid, TransactionId *subXid);
+
+/* Parser for WAL_REC_INSERT */
+/* Parser for WAL_REC_UPDATE */
+/* Parser for WAL_REC_DELETE */
+/* Parser for WAL_REC_REINSERT */
+
```
za-arthur:
```suggestion
```

src/recovery/wal.c
```diff
@@ -45,6 +46,194 @@ static void add_modify_wal_record_extended(uint8 rec_type, BTreeDescr *desc,
...
+	return "UNKNOWN";
+}
+
+#define PARSE(ptr, valptr) \
+{ \
+	if (valptr) \
+	{ \
+		memcpy(valptr, ptr, sizeof(*(valptr))); \
+	} \
+	ptr += sizeof(*(valptr)); \
```
za-arthur, line 103: I think this macro is bogus since in this line there is not check for NULL.

src/recovery/wal.c
```diff
@@ -45,6 +46,194 @@ static void add_modify_wal_record_extended(uint8 rec_type, BTreeDescr *desc,
...
+}
```
za-arthur: It might be better to use `do ... while(0)` statement here.
```suggestion
do
{ \
	if (valptr) \
	{ \
		memcpy(valptr, ptr, sizeof(*(valptr))); \
	} \
	ptr += sizeof(*(valptr)); \
} while (0)
```

include/transam/oxid.h
```diff
@@ -43,6 +43,16 @@ typedef struct
...
 extern XidMeta *xid_meta;
 
+typedef struct
+{
+	TransactionId xid;			/* a 32-bit transaction id to be used during
+								 * logical decoding */
+	bool		useHeap;		/* flag indicates if current logical xid was
+								 * allocated when heap xid has been already
+								 * set */
+
```
za-arthur:
```suggestion
```

include/transam/oxid.h
```diff
@@ -84,6 +94,10 @@ o_check_isolation_level(void)
 
 #define XLOG_PTR_ALIGN(ptr) ((ptr) + ((ptr) & 1))
 
+extern void oxid_subxact_callback(
+								  SubXactEvent event, SubTransactionId mySubid,
```
za-arthur:
```suggestion
extern void oxid_subxact_callback(SubXactEvent event, SubTransactionId mySubid,
```

include/transam/undo.h
```diff
@@ -392,6 +392,8 @@ extern void o_add_rewind_relfilenode_undo_item(RelFileNode *onCommit,
 											   RelFileNode *onAbort,
 											   int nOnCommit, int nOnAbort);
 
+#define TRANSACTION_HAS_UNDO_CHANGES() (have_retained_undo_location())
```
za-arthur: Why just not use the function `have_retained_undo_location()` instead of the macro?

src/recovery/logical.c
```diff
@@ -571,11 +603,19 @@ orioledb_decode(LogicalDecodingContext *ctx, XLogRecordBuffer *buf)
...
+			}
+
 			/* unknown transaction, nothing to replay */
 			if (txn != NULL)
 			{
-				if (txn->toptxn)
-					txn = txn->toptxn;
+				/* if (txn->toptxn) */
+				/* txn = txn->toptxn; */
+
```
za-arthur:
```suggestion
```

src/recovery/logical.c
```diff
@@ -670,32 +722,43 @@ orioledb_decode(LogicalDecodingContext *ctx, XLogRecordBuffer *buf)
...
+				 * ReorderBufferCommitChild
+				 */
+				elog(DEBUG4, "SKIP_COMMIT_CHILD record type %d (%s) oxid %lu logicalXid %u", rec_type, rec_type_str, oxid, logicalXid);
+
 				UpdateDecodingStats(ctx);
 			}
 
 			oxid = InvalidOXid;
 			logicalXid = InvalidTransactionId;
+
```
za-arthur:
```suggestion
```

src/recovery/logical.c
```diff
@@ -809,11 +857,63 @@ orioledb_decode(LogicalDecodingContext *ctx, XLogRecordBuffer *buf)
...
+			OXid		xmin;
+			dlist_iter	cur_txn_i;
+			ReorderBufferTXN *txn;
+			CommitSeqNo csn;
+			CSNSnapshotData *csnSnapshot;
 
-			memcpy(&parentSubid, ptr, sizeof(SubTransactionId));
-			ptr += sizeof(SubTransactionId);
+			xmin = InvalidOXid;
+			csn = 0;
```
za-arthur:
```suggestion
```

src/recovery/logical.c
```diff
@@ -809,11 +857,63 @@ orioledb_decode(LogicalDecodingContext *ctx, XLogRecordBuffer *buf)
...
 
+			ptr = wal_parse_rec_rollback_to_savepoint(ptr, &parentSubid, &xmin, &csn);
+
+#if ORIOLEDB_WAL_VERSION < (17)
 			/* Skip */
+			continue;
+#endif
+
```
za-arthur:
```suggestion
```

include/recovery/wal.h
```diff
@@ -121,6 +144,10 @@ typedef struct
 {
 	uint8		recType;
 	uint8		parentSubid[sizeof(SubTransactionId)];
+#if ORIOLEDB_WAL_VERSION >= (17)
+	uint8		xmin[sizeof(OXid)];
+	uint8		csn[sizeof(CommitSeqNo)];
+#endif
```
za-arthur:
```suggestion
	uint8		xmin[sizeof(OXid)];
	uint8		csn[sizeof(CommitSeqNo)];
```

src/recovery/logical.c
```diff
@@ -809,11 +857,63 @@ orioledb_decode(LogicalDecodingContext *ctx, XLogRecordBuffer *buf)
...
+				}
+				elog(DEBUG4, "ABORT record type %d (%s) oxid %lu logicalXid %u", rec_type, rec_type_str, oxid, logicalXid);
+				ReorderBufferAbort(ctx->reorder, logicalXid,
+								   startXLogPtr, 0);
+			}
+
+			UpdateDecodingStats(ctx);
+
+			oxid = InvalidOXid;
+			logicalXid = InvalidTransactionId;
```
za-arthur, line 936: Is it correct to invalidate `oxid` and `logicalXid` here? As far as I understand we just rollback to savepoint here but the transaction itself might be still open.

src/recovery/wal.c
```diff
@@ -45,6 +46,194 @@ static void add_modify_wal_record_extended(uint8 rec_type, BTreeDescr *desc,
...
+Pointer
+wal_parse_rec_rollback_to_savepoint(Pointer ptr, SubTransactionId *parentSubid, OXid *xmin, CommitSeqNo *csn)
+{
+	Assert(ptr);
+
+	PARSE(ptr, parentSubid);
+#if ORIOLEDB_WAL_VERSION >= (17)
+	PARSE(ptr, xmin);
+	PARSE(ptr, csn);
+#endif
```
za-arthur:
```suggestion
	PARSE(ptr, xmin);
	PARSE(ptr, csn);
```

src/recovery/wal.c
```diff
@@ -553,18 +790,29 @@ void
 add_rollback_to_savepoint_wal_record(SubTransactionId parentSubid)
 {
 	WALRecRollbackToSavepoint *rec;
+#if ORIOLEDB_WAL_VERSION >= (17)
+	OXid		runXmin;
+	CommitSeqNo csn;
+#endif
```
za-arthur:
```suggestion
	OXid		runXmin;
	CommitSeqNo csn;
```

src/recovery/wal.c
```diff
@@ -553,18 +790,29 @@ void
...
 	rec = (WALRecRollbackToSavepoint *) (&local_wal_buffer[local_wal_buffer_offset]);
 
 	rec->recType = WAL_REC_ROLLBACK_TO_SAVEPOINT;
 	memcpy(rec->parentSubid, &parentSubid, sizeof(SubTransactionId));
+#if ORIOLEDB_WAL_VERSION >= (17)
+	runXmin = pg_atomic_read_u64(&xid_meta->runXmin);
+	memcpy(rec->xmin, &runXmin, sizeof(runXmin));
+	csn = pg_atomic_read_u64(&TRANSAM_VARIABLES->nextCommitSeqNo);
+	memcpy(rec->csn, &csn, sizeof(csn));
+#endif
```
za-arthur:
```suggestion
	runXmin = pg_atomic_read_u64(&xid_meta->runXmin);
	memcpy(rec->xmin, &runXmin, sizeof(runXmin));
	csn = pg_atomic_read_u64(&TRANSAM_VARIABLES->nextCommitSeqNo);
	memcpy(rec->csn, &csn, sizeof(csn));
```

src/recovery/wal.c
GremSnoort: Changes of `WALRecRollbackToSavepoint` are hidden under `#if ORIOLEDB_WAL_VERSION >= (17)` check

src/recovery/wal.c
GremSnoort: Changes of `WALRecRollbackToSavepoint` are hidden under `#if ORIOLEDB_WAL_VERSION >= (17)` check

test/t/logical_test.py
```diff
@@ -53,10 +85,680 @@ def test_simple(self):
...
+#			CREATE TABLE o_data(id serial primary key, data text) USING orioledb;
+#			INSERT INTO o_data(data) VALUES('10');
+#			ABORT;
+#		''')
+#
+#	result = self.squashLogicalChanges(
+#	    node.execute( # TRAP: failed Assert("descr != NULL") because there is no relation `o_data` in orioledb_table
+#	        "SELECT * FROM pg_logical_slot_get_changes('regression_slot', NULL, NULL);"
+#	    ))
+#	#print(result)
```
za-arthur, line 166: If this are a temporary tests might be better to delete them.
```suggestion
```

za-arthur (changes requested):
Thank you for the work on this issue! I especially like the new WAL records parse functions.
I have few suggestions below.

src/recovery/wal.c
za-arthur: I think @pashkinelfe meant runtime check rather than compile time check

pashkinelfe: Exactly. Need to compare to wal_version that's defined in orioledb_decode() and read there from wal container
e.g.
if (wal_version >=17)

include/recovery/wal.h
pashkinelfe: +1

src/recovery/wal.c
pashkinelfe: +1

include/recovery/wal.h
pashkinelfe: +1

src/recovery/logical.c
pashkinelfe: +1

include/recovery/wal.h
pashkinelfe: +1

src/recovery/wal.c
pashkinelfe: +1

src/recovery/wal.c
GremSnoort: Done. Runtime check added in parser for `WAL_REC_ROLLBACK_TO_SAVEPOINT`

src/recovery/wal.c
GremSnoort: Done. Runtime check added in parser for `WAL_REC_ROLLBACK_TO_SAVEPOINT`

include/recovery/wal.h
GremSnoort: Done

include/recovery/wal.h
GremSnoort: Static check removed

src/recovery/wal.c
GremSnoort: Done

src/recovery/wal.c
GremSnoort: Done

src/recovery/wal.c
GremSnoort: Done

src/recovery/wal.c
GremSnoort: Changed this to runtime check `if (wal_version >= 17)`

include/recovery/wal.h
GremSnoort: Done

include/transam/oxid.h
GremSnoort: Done

include/transam/oxid.h
GremSnoort: Done

src/recovery/logical.c
GremSnoort: Done

src/recovery/logical.c
GremSnoort: Change this to runtime check `if (wal_version < 17)` - then skip and continue

test/t/logical_test.py
GremSnoort: These tests reproduce an **existing issue**: incorrect state of OrioleDB system catalogs during logical decoding. This problem arises because changes to Oriole sys trees are not included in MVCC-historical snapshot and are not applied on replaying changes from the reorder buffer.
During logical decoding, when processing each command, we observe a final state of the Oriole sys catalogs rather than some intermediate state that was relevant at the time when the current command has been executed within transaction.
These tests should be enabled but only after this issue has been resolved.

pashkinelfe: Could you add an explanation of when to enable this test in the comments before the commented-out test? And leave it commented now.

include/transam/undo.h
pashkinelfe: +1

pashkinelfe (changes requested):

include/recovery/wal.h
```diff
@@ -35,6 +35,22 @@
...
+ * In this case, the main transaction consists of two parts: heap and Oriole.
+ * Oriole's modifications come with their own Oriole logical xid as a sub-transaction to the top heap transaction.
+ *
+ * This WAL record type is intended to define the connection between top heap xid and child xid Oriole.
+ *
+ * Otherwise, without this connection, main transaction suddenly becomes splitted into two independent parts.
+ * From logical decoder's point of view this looks like two independent transactions but in fact internally related to each other.
+ * This situation outcomes in troubles for logical decoder with visibility of heap modifications
+ * in Oriole's sub-part due to incorrect state of the MVCC-historical snapshot.
+ */
```
pashkinelfe, line 49: I propose to shorten like this:
"Record type for a case when both heap and Oriole apply changes within a single transaction,
so one logical xid is assigned by heap, and the other is assigned by Oriole. This record defines a connection between Oriole a sub-transaction xid and a xid of to the top heap transaction which is needed for logical decoder."

src/transam/oxid.c
```diff
@@ -1142,7 +1354,32 @@ get_current_oxid_if_any(void)
...
 get_current_logical_xid(void)
 {
-	return logicalXid;
+	TransactionId heapXid;
+
+	heapXid = GetTopTransactionIdIfAny();
+	if (TransactionIdIsValid(heapXid) && TransactionIdIsValid(logicalXidMeta.xid) && !logicalXidMeta.useHeap)
+	{
+		elog(DEBUG4, "[%s] SWITCH_LOGICAL_XID O2H heap xid %u -> oriole xid %u", __func__, heapXid, logicalXidMeta.xid);
+		add_switch_logical_xid_wal_record(heapXid, logicalXidMeta.xid);
```
pashkinelfe: Should this be called when we call get_current_logical_xid() from:
undo_xact_callback(), 
undo_subxact_callback(), 
start_autonomous_transaction(), 
abort_autonomous_transaction(), 
finish_autonomous_transaction()

?

src/transam/oxid.c
```diff
@@ -1142,7 +1354,32 @@ get_current_oxid_if_any(void)
...
+}
+
+TransactionId
+get_current_logical_xid_if_any(void)
+{
+	return logicalXidMeta.xid;
+}
+
+void
+retrieve_current_logical_xid(LogicalXid *output)
```
pashkinelfe: Could we add some semantic difference to this function name? Looks similar to get_current_logical_xid() Do we need both variants anyway?

src/recovery/wal.c
GremSnoort: Here `valptr` is checked for the `memcpy` operation, so we may pass NULL.
Expression `sizeof(*(valptr))` is used intentionally, because it is correct and this won't cause any problems if `valptr` is NULL.
- https://stackoverflow.com/questions/19785518/is-dereferencing-null-pointer-valid-in-sizeof-operation
- https://www.open-std.org/jtc1/sc22/wg14/www/docs/n1256.pdf
6.5.3.4 The sizeof operator
Expression `double *dp = alloc(sizeof *dp);` is also correct, according to std doc.
- https://en.wikipedia.org/wiki/Sizeof
- https://sourceforge.net/p/cppcheck/discussion/general/thread/56b5b41a08/

Should i add check for `ptr`?

src/recovery/logical.c
```diff
@@ -829,10 +935,20 @@ orioledb_decode(LogicalDecodingContext *ctx, XLogRecordBuffer *buf)
...
 			ReorderBufferTXN *txn;
 
-			ReorderBufferProcessXid(ctx->reorder, logicalXid, changeXLogPtr);
-
 			read_modify_wal_tuples(rec_type, &ptr, &tuple1, &tuple2, &debug_length);
 
+			if (!TransactionIdIsValid(logicalXid))
+			{
+				/*
+				 * Oriole logical xid stays invalid for an autonomous
```
pashkinelfe, line 953: I think autonomous transactions are already rejected by checking (cur_oids.datoid == ctx->slot->data.database) below. Worth to investigate this.

include/transam/undo.h
GremSnoort: I wanted to make the code easier to understand.
In the current context, checking `have_retained_undo_location` means that the current transaction is making modifications, i.e., it is not read-only. It makes sense to log a transaction if it's not read-only.
May i keep this macro but rename it to `TRANSACTION_IS_READ_ONLY` or smth like that?

src/recovery/wal.c
pashkinelfe: I agree on sizeof validity. Memcpy has UB with null pointer, but this haven't changed in this PR. We shouldn't parse anything without a valid pointer. So it could be left as it is now.

src/recovery/logical.c
GremSnoort: `logicalXid` behaves same as native pg heap xid for now: a new xid is allocated for each sub-txn if it's not read-only (for Oriole, if `have_retained_undo_location`); this allows to correctly build up a reorder buffer with one top-txn and many sub-txns the same way as native pg does it.
Each rollback to savepoint WAL record forces adding refreshed xid record on future changes going after it (`add_rollback_to_savepoint_wal_record`, set `local_wal_contains_xid = false;`).
On a sub-txn commit/rollback, we restore prevoius `logicalXid` from pre-saved list `prevLogicalXids`.

src/recovery/wal.c
pashkinelfe: Resolved after discussion off this thread

include/recovery/wal.h
GremSnoort: Comment is updated

test/t/logical_test.py
GremSnoort: Explanation added

include/transam/undo.h
za-arthur: I think `have_retained_undo_location()` is already used in a few places.  If the transaction doesn't have undo changes, it probably doesn't necessarily mean that it is read only.

GremSnoort: Removed this macro

src/transam/oxid.c
GremSnoort: found a better way to guarantee that switch xid record will be added: see latest commit

src/transam/oxid.c
```diff
@@ -63,13 +64,60 @@
...
+ *     - oriole write       - oriole xid - COMMIT oriole xact
+ *     - heap->oriole write - SWITCH_LOGICAL_XID H2O: \
+ *                                                     Oriole txn acts as a sub-txn of a top heap txn
+ *     - oriole->heap write - SWITCH_LOGICAL_XID O2H: /
  */
-static TransactionId logicalXid = InvalidTransactionId;
-static List *prevLogicalXids = NIL;
+static OXid curOxid = InvalidOXid;	/* a 64-bit OrioleDB oxid */
+
+static LogicalXid logicalXidMeta = {InvalidTransactionId, false};
```
pashkinelfe: I propose renaming type LogicalXid for not having the similar name to TransactionId logicalXid

src/recovery/logical.c
GremSnoort: Done

src/recovery/logical.c
GremSnoort: Done

src/transam/oxid.c
GremSnoort: Renamed: `retrieve_current_logical_xid` to `get_current_logical_xid_ctx`.
In many places we need to get just xid without its meta info, so the previous version `get_current_logical_xid` was retained.
The new version `get_current_logical_xid_ctx` is intended to use when we need to know if the heap xid was allocated already at the moment of Oriole xid assignment.

src/transam/oxid.c
GremSnoort: Renamed to `LogicalXidCtx`

include/orioledb.h
```diff
@@ -398,6 +398,9 @@ extern bool orioledb_strict_mode;
 	(AssertMacro(MYPROCNUMBER >= 0 && \
 				 MYPROCNUMBER < max_procs), \
 	 &oProcData[MYPROCNUMBER])
+#define O_GET_AUTONOMOUS_NESTING_LEVEL() (GET_CUR_PROCDATA()->autonomousNestingLevel)
+#define O_IS_TRANSACTION_AUTONOMOUS() \
```
pashkinelfe: I propose to delete these two macro as unused

src/transam/oxid.c
```diff
@@ -63,13 +64,60 @@
...
+ *     - oriole->heap write - SWITCH_LOGICAL_XID O2H: /
  */
-static TransactionId logicalXid = InvalidTransactionId;
-static List *prevLogicalXids = NIL;
+static OXid curOxid = InvalidOXid;	/* a 64-bit OrioleDB oxid */
+
+static LogicalXidCtx logicalXidContext = {InvalidTransactionId, false};
+
+static inline void
+reset_logical_xid_meta(void)
```
pashkinelfe: Propose to rename to reset_logical_xid_ctx

src/transam/oxid.c
```diff
@@ -63,13 +64,60 @@
...
+{
+	logicalXidContext.xid = InvalidTransactionId;
+	logicalXidContext.useHeap = false;
+}
+
+static inline LogicalXidCtx *
+clone_logical_xid_meta(void)
```
pashkinelfe: Propose to renane to clone_logical_xid_ctx

src/transam/oxid.c
```diff
@@ -301,40 +360,176 @@ acquire_logical_xid(void)
...
 		MemoryContextSwitchTo(mcxt);
 	}
 
-	logicalXid = nextLogicalXid;
+	logicalXidContext.xid = nextLogicalXid;
+	logicalXidContext.useHeap = isValidHeapXid;
+}
+
+static void
+setup_prev_logical_xid_meta(void)
```
pashkinelfe: propose to rename to setup_prev_logical_xid_ctx

pashkinelfe (changes requested):

include/orioledb.h
GremSnoort: Done

src/transam/oxid.c
GremSnoort: Done

src/transam/oxid.c
GremSnoort: Done

src/transam/oxid.c
GremSnoort: Done

Issue #609, opened 2025-10-16
Can't stop replica node


ereutt:
Hi!

I catch situation when after I try to stop replica node it actually not total stoped. `ps aux `shows that postmaster and checkpointer still alive. Postmaster waits chekpointer and but checkpointer is waiting wal receiver which has already gone. Via gdb I find infinity loop which catch it
```
#0  0x00007f8d27b0fdf5 in nanosleep () from /lib64/libpthread.so.0
#1  0x0000000000bbf611 in pg_usleep (microsec=10000) at pgsleep.c:50
#2  0x00007f8d1735e1c7 in finish_write_xids (chkpnum=5) at src/checkpoint/checkpoint.c:858
#3  0x00007f8d1735ec29 in o_perform_checkpoint (redo_pos=1846705968, flags=5) at src/checkpoint/checkpoint.c:1221
#4  0x000000000056f687 in CheckPointGuts (checkPointRedo=1846705968, flags=5) at xlog.c:7540
#5  0x000000000056fbd9 in CreateRestartPoint (flags=5) at xlog.c:7715
#6  0x000000000056dc49 in ShutdownXLOG (code=0, arg=0) at xlog.c:6622
#7  0x00000000008c8bf9 in HandleCheckpointerInterrupts () at checkpointer.c:625
#8  0x00000000008c8aaa in CheckpointerMain (startup_data=0x0, startup_data_len=0) at checkpointer.c:541
#9  0x00000000008c9ce6 in postmaster_child_launch (child_type=B_CHECKPOINTER, startup_data=0x0, startup_data_len=0, client_sock=0x0) at launch_backend.c:277
#10 0x00000000008cf8e0 in StartChildProcess (type=B_CHECKPOINTER) at postmaster.c:3898
#11 0x00000000008cc461 in PostmasterMain (argc=3, argv=0x187cb90) at postmaster.c:1342
#12 0x000000000079c68d in main (argc=3, argv=0x187cb90) at main.c:199 
```

Actually I can't introduce concrete step to reproduce it with 100% guarantee :(

I just discover that it happens if:

1. in replica log I see message about "restartpoint starting:"
2. stop master firstly
3. try to stop replica

This happens also if I just load orioleDB, and create heap relation.

I've use oriole main branch commit dcdea4c034cba2a67d9caf50e2e901a52944fa7f

Fill free to ask any additional information

akorotkov:
[...]

PR #644, opened 2025-11-06
Save recovery state to temp file on shutdown

e-ivkov:
Closes https://github.com/orioledb/orioledb/issues/609

### DIFF
--- a/include/recovery/internal.h
+++ b/include/recovery/internal.h
@@ -134,6 +134,7 @@ typedef struct
 	pg_atomic_uint64 commitPtr;
 	pg_atomic_uint64 retainPtr;
 	uint32		flushedUndoLocCompletedCheckpointNumber;
+	pg_atomic_flag has_temp_file;
 } RecoveryWorkerPtrs;
 
 typedef struct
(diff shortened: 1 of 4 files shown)


src/checkpoint/checkpoint.c
```diff
@@ -841,6 +842,16 @@ finish_write_xids(uint32 chkpnum)
 		LWLockRelease(&oProcData[i].undoStackLocationsFlushLock);
 	}
 
+	/* Check for exited recovery workers with temp files */
+	for (i = 0; i < total_recovery_workers; i++)
+	{
+		if (pg_atomic_unlocked_test_flag(&worker_ptrs[i].has_temp_file))
+		{
+			recovery_load_and_process_state_from_temp_file(i, chkpnum);
+			pg_atomic_clear_flag(&worker_ptrs[i].has_temp_file);
```
akorotkov: Why do we clear a flag here?  It means we would be able to use the temp file only once. The idea of this fix is to be able to use this file in any number of checkpoints (we yet know they can be 2).

include/recovery/internal.h
```diff
@@ -134,6 +134,7 @@ typedef struct
 	pg_atomic_uint64 commitPtr;
 	pg_atomic_uint64 retainPtr;
 	uint32		flushedUndoLocCompletedCheckpointNumber;
+	pg_atomic_flag has_temp_file;
```
za-arthur:
```suggestion
	pg_atomic_flag hasTempFile;
```

PR #645, opened 2025-11-07
Add new WAL record `WAL_REC_XACT_TIME`

za-arthur:
This will allow to do proper PITR using `recovery_target_time`. It is implemented by adding new WAL record of `WAL_REC_XACT_TIME` type and new hook `RecoveryStopsBeforeHook`.

Issue: https://github.com/orioledb/orioledb/issues/597
Relevant PRs:
- https://github.com/orioledb/postgres/pull/31
- https://github.com/orioledb/postgres/pull/32


src/recovery/recovery.c
```diff
@@ -842,6 +842,58 @@ recovery_get_effective_replay_ptr(void)
...
+		return false;
+
+	/* If for some reason data is empty just exit */
+	if (XLogRecGetDataLen(record) == 0)
+		return false;
+
+	ptr = wal_container_read_header(ptr, &wal_version, &wal_flags);
+	/* Unexpected new WAL record which we cannot read */
+	if (wal_version > ORIOLEDB_WAL_VERSION)
+		elog(ERROR, "cannot read WAL record of version %u newer than supported %u",
```
pashkinelfe: What will happen to the recovery after this error? Should we call abort_recovery()?

za-arthur: I replaced the message with `elog(PANIC, ...)` to crash the instance completely.

Issue #639, opened 2025-10-29
"Invalid Datum Pointer" Error issues


raikamil07:
Hello everyone,

I’d like to share an issue I encountered while developing Odoo 18 Community using PostgreSQL with the OrioleDB extension as the database backend.

When I tried to create a new database for Odoo, I got the following error in the Odoo logs:

> ERROR: invalid Datum pointer
...
psycopg2.errors.DataException: invalid Datum pointer`

Here’s a more complete trace from the Odoo log:

> 2025-10-20 07:15:02,131 2904996 INFO ? werkzeug: 19x.16x.19x.17x - - [20/Oct/2025 07:15:02] "GET /websocket?version=saas-18.3-4 HTTP/1.1" 404 - 0 0.000 0.051
2025-10-20 07:15:05,576 2904994 ERROR None odoo.sql_db: bad query: b' UPDATE "ir_act_client"\n                    SET "params_store" = "__tmp"."params_store"::bytea, "write_date" = "__tmp"."write_date"::timestamp, "write_uid" = "__tmp"."write_uid"::int4\n                    FROM (VALUES (3, \'\\x7b276d656e755f6964273a20317d\'::bytea, \'2025-10-20T07:12:19.908661\'::timestamp, 1)) AS "__tmp"("id", "params_store", "write_date", "write_uid")\n                    WHERE "ir_act_client"."id" = "__tmp"."id"\n                '
ERROR: invalid Datum pointer
2025-10-20 07:15:05,588 2904994 INFO None odoo.modules.loading: skipping reset_modules_state, ir_module_module table does not exists 
2025-10-20 07:15:05,590 2904994 ERROR None odoo.registry: Failed to load registry 
2025-10-20 07:15:05,590 2904994 ERROR None odoo.service.db: CREATE DATABASE failed: 
Traceback (most recent call last):
  File "/opt/odoo18/odoo/odoo/service/db.py", line 66, in _initialize_db
    registry = odoo.modules.registry.Registry.new(db_name, update_module=True, new_db_demo=demo)
               ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
  File "/opt/odoo18/venv/lib/python3.12/site-packages/decorator.py", line 232, in fun
    return caller(func, *(extras + args), **kw)
[...]

PR #652, opened 2025-11-21
Don't pass null vals to datumIsEqual

e-ivkov:
Closes https://github.com/orioledb/orioledb/issues/639

The function `tts_orioledb_modified` incorrectly calls `datumIsEqual()` when both old and new values are NULL. The code checks if values differ when one is NULL and the other isn't, but fails to skip comparison when both are NULL, leading to `datumIsEqual()` being called with NULL pointers for pass-by-reference types.

Now the code skips datumIsEqual check when both values are NULL

### DIFF
--- a/src/tuple/slot.c
+++ b/src/tuple/slot.c
@@ -1834,6 +1834,22 @@ tts_orioledb_update_toast_values(TupleTableSlot *oldSlot,
 	return result;
 }
 
+/*
+ * tts_orioledb_modified - Check if specified attributes were modified between two tuples
+ *
+ * Compares the values of specific attributes between an old and new tuple slot
+ * to determine if any modifications have occurred. This is primarily used during
+ * UPDATE operations to distinguish between key and non-key updates.
+ *
+ * Parameters:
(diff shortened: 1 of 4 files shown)


src/tuple/slot.c
```diff
@@ -1866,11 +1882,11 @@ tts_orioledb_modified(TupleTableSlot *oldSlot,
...
-			if (isnull1 || isnull2)
-			{
-				if (isnull1 != isnull2)
-					return true;
-			}
+			if (isnull1 != isnull2)
+				return true;
+
+			if (isnull1)
+				return true;
```
pashkinelfe: I think this is wrong and should be ```continue;``` (to next attr). (This is the case when  isnull1 == true, and isnull2 == true) Test should also fail on this, but they don't.

e-ivkov: Thanks! I mistyped, will correct now

e-ivkov: I don't think tests could have detected this erroneous return as tts_orioledb_modified return value only influences locking strategy

PR #654, opened 2025-11-24
Fix index-only scan costing and optimize tuple filling performance

Serge-sudo:
OrioleDB doesn't use PostgreSQL's visibility map (VM) fork, causing all relation pages to be considered "not all visible". This leads the cost estimator to prefer bitmap scans over index-only scans, as it assumes all tuples will need to be fetched from the heap and assigns higher costs to index-only scans.

Since OrioleDB uses MVCC-aware indexes, no heap fetches are required during index-only scans. The VM fork in PostgreSQL is primarily used to determine when index-only scans are beneficial. In OrioleDB's case, when all required attributes are available in the index, index-only scans are always more efficient than bitmap scans since no heap fetches are needed.

Performance profiling revealed that approximately 50% of execution time was spent in the fill_itup function. This commit includes several optimizations for this bottleneck, which are documented in the code comments.


src/tableam/handler.c
```diff
@@ -1344,39 +1343,50 @@ orioledb_estimate_rel_size(Relation rel, int32 *attr_widths,
...
+		/*
+		 * Without reltuples/relpages, we also need to consider fillfactor.
+		 * The other branch considers it implicitly by calculating density
+		 * from actual relpages/reltuples statistics.
+		 */
+		fillfactor = RelationGetFillFactor(rel, HEAP_DEFAULT_FILLFACTOR);
 
 		tuple_width = get_rel_data_width(rel, attr_widths);
-		tuple_width += MAXALIGN(SizeOfOTupleHeader);
+		tuple_width += HEAP_OVERHEAD_BYTES_PER_TUPLE;
```
Serge-sudo, line 1370: refreshed function with pg17 implementation

pashkinelfe:
Likely stale PR, not relevant after 9c831a5f809e2c8dbef1f0741f4eecae3e4c9fb9 and 8bd7373748747a767e71907fd488667776bdc687

PR #656, opened 2025-11-26
Restart PostgreSQL instance when rewind is done 

pgnickb:


src/rewind/rewind.c
```diff
@@ -785,9 +785,71 @@ orioledb_rewind_internal(int rewind_mode, int rewind_time, OXid rewind_oxid, Tra
...
+			char cmd[MAXPGPATH + 16];
+			char *lastslash;
+
+			strlcpy(bindir, my_exec_path, MAXPGPATH);
+
+			lastslash = strrchr(bindir, '/');
+			Assert(lastslash != NULL);
+			*lastslash = '\0';
+			sprintf(cmd, "%s/pg_ctl", bindir);
+			execl(cmd, cmd, "restart", "-D", DataDir, (char *)NULL);
```
za-arthur: This probably might be a problem if some environment will use different and additional options

pgnickb: I also was concerned about that. Fortunately `pg_ctl restart` actually relies on `postmaster.opts` to use the exact same options that were used to start the cluster originally.

The only case where `postmaster.opts` are ignored is when user specifies additional options so we should be careful not to do that.

Do you think this is something we should cover in tests?

za-arthur: Good to know! I think it is okay to not cover this case with tests.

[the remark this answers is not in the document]
pgnickb: Ok this is a rather confusing comment but I think I understand what is meant. The code shown is a bit outdated.

src/rewind/rewind.c
```diff
@@ -785,9 +785,71 @@ orioledb_rewind_internal(int rewind_mode, int rewind_time, OXid rewind_oxid, Tra
...
+		  /* fork succeeded, in child */
+
+#ifdef HAVE_SETSID
+		  /* if we have setsid(), the easiest way to detach is to just call it
+		   */
+		  if (setsid() < 0) {
+			/* We can do better, but bail for now */
+			elog(LOG, "Attempt to restart failed. Stopping instead");
+			(void)kill(PostmasterPid, SIGTERM);
+			return; /* silence stupider compilers */
```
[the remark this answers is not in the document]
pgnickb: This is an interesting point. I thought about it, but since `setsid` didn't succeed I  expected the child to still belong to the `postmaster` session and thus the `SIGTERM` should clean it up for us.
I might be wrong though, since technically our child process didn't fork from postmaster, but from our random backend. I will look into this.

src/rewind/rewind.c
```diff
@@ -785,9 +785,71 @@ orioledb_rewind_internal(int rewind_mode, int rewind_time, OXid rewind_oxid, Tra
...
+#else
+		{
+		  pid_t pid;
+
+		  pid = fork();
+
+		  if (pid < 0) {
+			/* fork failed */
+			elog(LOG, "Attempt to restart failed. Stopping instead");
+			(void)kill(PostmasterPid, SIGTERM);
```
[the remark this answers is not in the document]
pgnickb: This is a valid concern. I don't actually know what is the proper way to handle it off the top of my head.
Intuitively `exit` should be called, but that also comes with implications. I'll look into this.

test/t/rewind_xid_test.py
```diff
@@ -82,10 +82,10 @@ def test_rewind_xid_oriole(self):
 		node.safe_psql(
 		    'postgres', "select orioledb_rewind_to_transaction(%d,%ld);\n" %
 		    (invalidxid, oxid))
-		time.sleep(1)
+		time.sleep(10)
```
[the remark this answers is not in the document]
pgnickb: Very helpful I was planning to write something like this later on. Although I suspect that `tesgres` already has something for this. Anyway I'll mark this as resolved.

src/rewind/rewind.c
```diff
@@ -1689,3 +1692,103 @@ get_rewind_run_xmin(void)
...
+	if (devnull >= 0)
+	{
+		/* We can't just close stderr/stdout fds, so redir*/
+		dup2(devnull, fileno(stderr));
+		dup2(devnull, fileno(stdout));
+
+		/* Be paranoid: we don't want to ever close stderr/stdout */
+		if (devnull > fileno(stdout) && devnull > fileno(stderr))
+			close(devnull);
+	}
```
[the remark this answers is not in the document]
pgnickb: Nothing we can do about this

src/rewind/rewind.c
```diff
@@ -1689,3 +1691,111 @@ get_rewind_run_xmin(void)
...
+ * On systems lacking these primitives (e.g. Windows), this is currently
+ * not supported, so we shutdown instead.
+ */
+static void
+try_restart_pg(void)
+{
+	/* Flush stdio channels just before fork, to avoid double-output problems */
+	fflush(NULL);
+	/* WIN32 doesn't support fork */
+#ifdef WIN32
```
pashkinelfe: I'd propose to disable (turning on) the rewind on these platforms at all. It's better to know that rewind is not supported, than to have non-standard behavior after rewind.

src/rewind/rewind.c
```diff
@@ -1689,3 +1691,111 @@ get_rewind_run_xmin(void)
...
+	elog(LOG, "Restart not supported on Windows. Stopping instead");
+	goto bail_and_shutdown;
+#else							/* !WIN32 */
+#ifndef HAVE_SETSID
+	/* We can do better, but bail for now */
```
pashkinelfe: I'd propose to disable (turning on) the rewind on these platforms at all. It's better to know that rewind is not supported, than to have non-standard behavior after rewind.

src/rewind/rewind.c
```diff
@@ -1689,3 +1691,111 @@ get_rewind_run_xmin(void)
...
+#else							/* HAVE_SETSID */
+	{
+		pid_t		pid;
+
+		pid = fork();
+
+		if (pid < 0)
+		{
+			/* fork failed */
+			elog(LOG, "Attempt to restart failed. Stopping instead");
```
pashkinelfe: Could we remove this as a duplicate to WARNING in bail_and_shutdown section? Maybe it's worth to have some specific message here under DEBUG loglevel

src/rewind/rewind.c
```diff
@@ -1689,3 +1691,111 @@ get_rewind_run_xmin(void)
...
+		{
+			/* fork successful, in parent: return to client */
+			return;
+		}
+
+		/* fork succeeded, in child */
+		if (setsid() < 0)
+		{
+			/* setsid failed, bail for now */
+			elog(LOG, "setsid(2) failed with error: %m\n"
```
pashkinelfe: I'd propose to hide it from the user with DEBUG loglevel.

src/rewind/rewind.c
```diff
@@ -1689,3 +1691,111 @@ get_rewind_run_xmin(void)
...
+			*lastslash = '\0';
+			snprintf(cmd, sizeof(cmd), "%s/pg_ctl", bindir);
+
+			/* Do a little dance with fds to make logger shutdown cleanly */
+			cleanup_fds();
+			/* Sleep 0.5s to let the parent return */
+			pg_usleep(500000);
+			execl(cmd, cmd, "restart", "-D", DataDir, (char *) NULL);
+			/* If we got here, means execl failed, bail */
+			elog(LOG, "Attempt to restart failed with error: %m");
```
pashkinelfe: I'd propose to convert this to some specific message under DEBUG loglevel

src/rewind/rewind.c
```diff
@@ -1689,3 +1691,111 @@ get_rewind_run_xmin(void)
...
+			goto bail_and_shutdown;
+		}
+#endif							/* !HAVE_SETSID */
+	}
+#endif							/* !WIN32 */
+bail_and_shutdown:
+	elog(WARNING, "Restart failed. Shutting down instead!");
```
pashkinelfe: I'd propose a mesage that reflects that rewind is complete but the cluster failed to restart, which is recommended to do manually

src/rewind/rewind.c
```diff
@@ -784,10 +787,9 @@ orioledb_rewind_internal(int rewind_mode, int rewind_time, OXid rewind_oxid, Tra
 	do_rewind(rewind_mode, rewind_time, rewindStartTimeStamp, rewind_oxid, rewind_xid, rewind_timestamp);
 
 	LWLockRelease(&rewindMeta->rewindEvictLock);
-	elog(LOG, "Rewind complete");
-	/* Restart Postgres */
-	(void) kill(PostmasterPid, SIGTERM);
-	return;
+	elog(LOG, "Rewind complete. Will attempt to restart");
```
pashkinelfe: This message could be hidden from the user under DEBUG loglevel. Instead
I'd propose to add a success message after a restart like: Rewind complete, system restarted.

test/t/rewind_xid_evict_large_test.py
```diff
@@ -23,6 +23,12 @@ class RewindXidTest(BaseTest):
 	def wait_shutdown_and_start(self, node):
 		super().wait_shutdown_and_start(node)
 
+	def wait_restart(self, node):
```
pashkinelfe: Could we do this like wait_shutdown_and_start() and not duplicating wait_restart code in 3 test files?

test/t/rewind_time_test.py
```diff
@@ -20,6 +20,12 @@ class RewindTest(BaseTest):
 	def wait_shutdown_and_start(self, node):
 		super().wait_shutdown_and_start(node)
 
+	def wait_restart(self, node):
+		time.sleep(5)
```
pashkinelfe: Could old node hang for 5 secs and still be running after that? (Before we had cases that checkpointer under valgrind took several seconds and just delay wasn't enough to be sure the cluster stopped)

pgnickb: I agree this is rather arbitrary.
Perhaps we should check server log for a specific message? That should be a better solution.

src/rewind/rewind.c
pgnickb: In my opinion there is no strong reason for this. So it'd mean doing extra work to forbid rewind to users who could otherwise use it, although with minor limitations.

src/rewind/rewind.c
```diff
@@ -1689,3 +1693,144 @@ get_rewind_run_xmin(void)
...
+
+	if (pid < 0)
+	{
+		/* fork failed, restore signals and bail */
+		sigprocmask(SIG_SETMASK, &old_sigset, NULL);
+
+		elog(DEBUG3, "fork failed while attempting to restart,"
+			 " stopping instead");
+
+		ereport(WARNING,
```
pashkinelfe, line 1735: Do we need the same message in emergency_shutdown section?

src/rewind/rewind.c
```diff
@@ -1689,3 +1693,144 @@ get_rewind_run_xmin(void)
...
+				(errmsg("could not restart instance"),
+				 errdetail("Sending shutdown request to postmaster.")));
+		(void) kill(PostmasterPid, SIGTERM);
+		return;
```
pashkinelfe, line 1739: Do we need exit(71) in emergency_shutdown but not here?

src/rewind/rewind.c
pgnickb: Logging isn't safe when in `emergency_shutdown`. It _can_ be done sometimes, but there are associated risks. Sometimes it can't be done. See the comment at the `emergency_shutdown`.

src/rewind/rewind.c
pgnickb: Yes. Emergency shutdown handles the forked backend which is inherently complicated to shutdown. This codepath is for the "parent" backend which is a normal pg backend and can exit cleanly.
We don't do `exit` here because we want the function to return to the user, which imo is a slightly better user experience indicating there was no (breaking) error.
If we disconnect from the user without returning from the function it might be unclear whether the rewind will persist or not and the user will be left guessing.

src/rewind/rewind.c
pashkinelfe: It's better to have a specific message when we can't restart due to setsid failed. But IMO if it's not possible, it's acceptable.

PR #658, opened 2025-11-26
Fixes related to running postgres regress tests with orioledb

homper:
- Added support for exclusion constraints
- Added support for deferrable constraints
- Fixes for testgres 1.11
- ...

Depends on:
https://github.com/orioledb/postgres/tree/patches16_exclude
https://github.com/orioledb/postgres/tree/patches17_exclude

TODO:

- [x] fix some simple bugs in our regress tests for regresscheck to pass
- [ ] add tests for recovery/replication/logical
- [x] unrevert commits from pr https://github.com/orioledb/orioledb/pull/593
- [ ] enable all non failing tests in parallel_schedule_oriole for postgres


include/orioledb.h
```diff
@@ -128,10 +128,27 @@ typedef enum
 	oIndexToast = 1,
 	oIndexBridge = 2,
 	oIndexPrimary = 3,
-	oIndexUnique = 4,
-	oIndexRegular = 5,
+	oIndexExclusion = 4,
```
akorotkov: Is there a specific reason why oIndexExclusion should come before oIndexUnique?  I think this complicates the binary update of SYS_TREES_O_INDICES.  If we even need to have this specific order for checkpointing, I think it would be better to add a new value to the end of the enum and adjust the comparison function.

src/btree/modify.c
```diff
@@ -1153,9 +1166,10 @@ o_btree_insert_unique(BTreeDescr *desc, OTuple tuple, BTreeKeyType tupleType,
 	OInMemoryBlkno blkno;
 	uint32		pageChangeCount;
 	LWLock	   *uniqueLock;
-	OBTreeModifyResult result;
+	OBTreeModifyResult result = 456;
```
akorotkov: What is this magic value of 456 doing here?

src/btree/modify.c
```diff
@@ -1144,7 +1156,8 @@ o_btree_insert_unique(BTreeDescr *desc, OTuple tuple, BTreeKeyType tupleType,
 					  Pointer key, BTreeKeyType keyType,
 					  OXid opOxid, CommitSeqNo opCsn,
 					  RowLockMode lockMode, BTreeLocationHint *hint,
-					  BTreeModifyCallbackInfo *callbackInfo)
+					  BTreeModifyCallbackInfo *callbackInfo,
+					  IndexUniqueCheck checkUnique)
```
akorotkov, line 1166: Which values of IndexUniqueCheck do we support here? I guess this is probably just UNIQUE_CHECK_YES and UNIQUE_CHECK_EXISTING.  If so, it needs to be documented and asserted.

sql/orioledb--1.4--1.5_dev.sql
za-arthur, line 1: Current orioledb version is 1.6 and therefore the file needs to be modified is `orioledb--1.5--1.6_dev.sql`.

src/catalog/ddl.c
```diff
@@ -1188,8 +1188,7 @@ orioledb_utility_command(PlannedStmt *pstmt,
 			case REINDEX_OBJECT_SCHEMA:
 			case REINDEX_OBJECT_SYSTEM:
 			case REINDEX_OBJECT_DATABASE:
-				if (concurrently)
-					has_orioledb = check_multiple_tables(stmt->name, stmt->kind, concurrently);
+				has_orioledb = check_multiple_tables(stmt->name, stmt->kind, concurrently);
```
za-arthur, line 1206: You deleted the check, but I guess this error is not valid anymore?
https://github.com/orioledb/orioledb/blob/8f2b34115c87e41fd8493f3b8aa90faca58eec2c/src/catalog/ddl.c#L507-L510

za-arthur: Although it seems it is still valid since `index_create()` function of Postgres will raise an error in case of concurrent reindex. I suppose it is necessary to check the `concurrently` argument.

PR #660, opened 2025-11-27
Support logical replication with REPLICA IDENTITY FULL

pashkinelfe:


src/catalog/ddl.c
```diff
@@ -1399,6 +1404,29 @@ orioledb_utility_command(PlannedStmt *pstmt,
 	free_parsestate(pstate);
 }
 
+static void
+o_validate_replica_identity(Relation rel, ReplicaIdentityStmt *stmt)
+{
+	elog(LOG, "Current replident %c, setting replident %c", rel->rd_rel->relreplident, stmt->identity_type);
```
za-arthur: Might be better to log this message in DEBUG level instead of LOG

```suggestion
	elog(DEBUG4, "Current replident %c, setting replident %c", rel->rd_rel->relreplident, stmt->identity_type);
```

za-arthur (changes requested):

akorotkov:
A couple of questions from my side:
1. Why do we need REORDER_BUFFER_TUPLE_TYPE macro?  Do we realistically expect it to be changed from HeapTuple?
2. Do I get correctly relreplident_ix_oid have no use, but reserved for future?  If so, I think this deserves comment.  Also, it would be good to assert that relreplident_ix_oid == InvalidOid during decoding.

Issue #450, opened 2025-03-24
Performance problems with sysbench benchmarks


mdcallag:
How I run sysbench is [explained here](https://smalldatum.blogspot.com/2017/02/using-modern-sysbench-to-compare.html). In this case I used a server with 48 cores, AMD SMT disabled, 128G of RAM and ran sysbench for Postgres 17.2, OrioleDB at patches17_5 and OrioleDB at patches17_6.

I run ~27 microbenchmarks. Relative to Postgres 17.2, OrioleDB did better on the write-heavy microbenchmarks than on the read-heavy. But OrioleDB was a lot slower than Postgres 17.2 on a few of the microbenchmarks and I focus on those here. I repeated the benchmark for 1, 10, 20 and 40 clients (threads).

Below I show relative QPS which is (QPS for OrioleDB / QPS for Postgres 17.2). When that > 1.0 then OrioleDB is faster than PG 17.2, when it is < 1.0 then OrioleDB is slower than PG 17.2

Note that some of these might be suffering from https://github.com/orioledb/orioledb/issues/392.

For 1 client (thread) the interesting results are:
* hot-points - relative QPS is ~0.25, this microbenchmark fetches many rows by PK via select with a large inlist
* points-covered-si, points-notcovered-si - the relative QPS is ~0.72 and ~0.60, these also do a select with an inlist with a predicate that can use a secondary index. And "covered" means the query can be index-only while "not-covered" means it cannot.
* range-covered-pk - relative QPS is ~0.72, this does a range scan on the PK index and can be index-only
* scan_range - this does a full table scan, it isn't index-only and is written so that no rows match the WHERE clause. The interesting result here is that patches17_5 does much better than patches17_6
* range-covered-si, range-notcovered-si - relative QPS is ~0.01 and this is by far the worst result for OrioleDB. They both do range scans on the secondary index and "covered" can be index-only while "not-covered" cannot.
```
Relative to: x.pg172_o2nofp.x10a_c32r128.pk1
col-1 : x.pg_oriole_patches17_5_o2nofp.x10a_oriole_c32r128.pk1
col-2 : x.pg_oriole_patches17_6_o2nofp.x10a_oriole_c32r128.pk1

col-1	col-2
0.25	0.26	hot-points_range=100
0.98	1.00	point-query_range=100
0.79	0.78	points-covered-pk_range=100
0.71	0.72	points-covered-si_range=100
0.99	0.99	points-notcovered-pk_range=100
[...]

PR #661, opened 2025-11-27
Reimplement bitmap scan with adaptive sparse/dense strategy to improve performance

Serge-sudo:
Bitmap scans in OrioleDB were significantly slower than in PostgreSQL due to fundamental architectural differences. PostgreSQL can map bitmap entries directly to heap blocks and use block-level prefetching, but OrioleDB indexes store primary keys without any direct block references. The previous implementation stored PKs in an RB-tree and executed a sequential table scan with subtree callbacks. However, OrioleDB’s seqscan does not cache pages in shared buffers — it performs disk reads directly — meaning subsequent bitmap scans do not benefit from previously accessed pages and performance degrades severely.

This patch introduces a new bitmap scan mechanism based on adaptive key-range classification. The PK ranges extracted from the RB-tree are split into sparse and dense regions.

Sparse regions: perform individual key lookups using the fast path (BTREE_PAGE_FIND_FETCH), which is efficient for isolated keys.

Dense regions: use iterator-based scanning (BTREE_PAGE_FIND_IMAGE) to load and walk pages sequentially. Since this path loads pages into memory, subsequent bitmap scans become faster due to warm page cache.

Fixed performance problem in "range-notcovered-si" test from #450 
on main branch - 145 tps
with patch - 3684 tps
on PG17 - 4261 tps

### DIFF
--- a/include/tableam/bitmap_scan.h
+++ b/include/tableam/bitmap_scan.h
@@ -45,6 +45,17 @@ extern TupleTableSlot *o_exec_bitmap_fetch(OBitmapScan *scan,
 										   CustomScanState *node);
 extern void o_free_bitmap_scan(OBitmapScan *scan);
 
+typedef struct
+{
+	RBTNode		rbtnode;
+	uint64		key;
+	uint8	   *bitmap;
+} OKeyBitmapRBTNode;
+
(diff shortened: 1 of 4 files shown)


akorotkov:
#945 contains a more global reimplementation of the bitmap scan.

Issue #548, opened 2025-09-01
Slow index creation


e-ivkov:
OrioleDB shows significantly slower index creation times than vanilla. Here is what I got from my benchmark:

```
create table a (a int);
create unique index a_i on a(a);
```

- empty table - in oriole 2 times slower
- 1 tuple - 13.5 times
- 10 tuples - 31 times
- 100 tuples - 20 times
- 1000 tuples - 16 times
- 10,000 tuples - 3 times

*Until this point, oriole shows constant perf mean_exec_time~11ms, only the vanilla value changes*

- 100,000 tuples - 2 times

PR #668, opened 2025-12-02
Compute parallel workers number for index build

e-ivkov:
Parallel workers number computation adapted from pg source to oriole.

Relates to https://github.com/orioledb/orioledb/issues/548

ambuild time:

| N Tuples | Heap        | Oriole Before | Oriole After |
| ------------- | ------------- | ------------------- | ----------------- |
| 1 | 0.274 ms | 3.778 ms | 1.061 ms |
|10 | 0.244 ms | 4.009 ms | 0.910 ms |
| 100 | 0.284 ms | 4.049 ms | 0.998 ms |
| 1 000 | 0.681 ms | 4.101 ms | 1.627 ms |
| 10 000 | 4.938 ms | 9.074 ms | 8.484 ms |

Measured like that:
```
echo "create table h_10 (a int);" | psql postgres
echo "insert into h_10 select g from generate_series(1,10) as g;" | psql postgres
echo "create unique index h_10_i on h_10(a);" | psql postgres
```

With default min_parallel_table_scan_size (8MB).

So up to 10 000 tuples there is performance gain.

### DIFF
--- a/src/catalog/indices.c
+++ b/src/catalog/indices.c
@@ -38,6 +38,7 @@
 #include "access/genam.h"
 #include "access/relation.h"
 #include "access/table.h"
+#include "access/xact.h"
 #include "catalog/heap.h"
 #include "catalog/index.h"
 #include "catalog/namespace.h"
@@ -56,6 +57,8 @@
 #include "utils/lsyscache.h"
 #include "utils/syscache.h"
 #include "utils/tuplesort.h"
+#include "optimizer/plancat.h"
+#include "optimizer/paths.h"
 
 /* copied from nbtsort.c with modifications*/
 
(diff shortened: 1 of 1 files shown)


src/catalog/indices.c
```diff
@@ -1389,13 +1475,47 @@ build_secondary_index(OTable *o_table, OTableDescr *descr, OIndexNumber ix_num,
...
+		 *
+		 * Note that each tuplesort participant receives an even share of the
+		 * total maintenance_work_mem budget.  Aim to leave participants
+		 * (including the leader as a participant) with no less than 32MB of
+		 * memory.  This leaves cases where maintenance_work_mem is set to
+		 * 64MB immediately past the threshold of being capable of launching a
+		 * single parallel worker to sort.
+		 */
+		while (parallel_workers > 0 &&
+			   maintenance_work_mem / (parallel_workers + 1) < 32768L)
```
za-arthur, line 1496: What is `32768` here?

src/catalog/indices.c
```diff
@@ -1389,13 +1475,47 @@ build_secondary_index(OTable *o_table, OTableDescr *descr, OIndexNumber ix_num,
...
+			parallel_workers--;
```
za-arthur, line 1497: Why do we need a loop here? Cannot it be written like the following?
```
int limit = maintenance_work_mem / 32768L;

if (limit <= 1)
    parallel_workers = 0;
else if (parallel_workers + 1 > limit)
    parallel_workers = limit - 1;
```

src/catalog/indices.c
e-ivkov: It's explained in the comment. In general I agree that it is somewhat ugly, but I took this peace from [postgres](https://github.com/orioledb/postgres/blob/b940ae16d45448cbe15808d23b5ed8b977e99e7f/src/backend/optimizer/plan/planner.c#L6827) and decided to leave it as it is

src/catalog/indices.c
e-ivkov: See previous comment thread

src/catalog/indices.c
```diff
@@ -1370,6 +1373,89 @@ build_secondary_index_worker_heap_scan(OTableDescr *descr, OIndexDescr *idx,
...
+ * -1 if we don't expect to scan any.
+ *
+ * "index_pages" is the number of pages from the index that we expect to scan, or
+ * -1 if we don't expect to scan any.
+ *
+ * "max_workers" is caller's limit on the number of workers.  This typically
+ * comes from a GUC.
+ */
+static int
+o_compute_parallel_worker(double table_pages, double index_pages,
```
pashkinelfe, line 1380: Could we call compute_parallel_worker() which is extern in PG? Appreciate it if we don't duplicate existing PG code.

e-ivkov: Ideally we would, but it [requires](https://github.com/orioledb/postgres/blob/b940ae16d45448cbe15808d23b5ed8b977e99e7f/src/backend/optimizer/path/allpaths.c#L4213) RelOptInfo which is a node in the plan and we are not working with the plan here in build_secondary_index

e-ivkov: I thought about changing the PG source, but decided better to duplicate here than to have a dubious diff in postgres fork

pashkinelfe: Makes sense. It's hard way to have RelOptInfo in the call. I also agree it's not worth patching PG code for that. Let's have duplicate function.

Could you then just mention in the comment on o_compute_parallel_worker() that it's a simplified version of compute_parallel_worker() (without  need to provide RelOptInfo etc.) ?

pashkinelfe (approved):

Issue #673, opened 2025-12-04
Segfault when creating many Orioledb tables.


Serge-sudo:
When creating a large number of Orioledb tables in a single loop, PostgreSQL consistently crashes with a segmentation fault. 

Below is the exact log line from the postmaster:

```
2025-12-04 13:13:16.422 MSK postmaster[2142140] LOG:  server process (PID 2142344) was terminated by signal 11: Segmentation fault
2025-12-04 13:13:16.422 MSK postmaster[2142140] DETAIL:  Failed process was running: DO $$
    DECLARE
        i INT;
    BEGIN
        FOR i IN 1..2000 LOOP
            EXECUTE format('
                CREATE TABLE indices.tbl_idx_%s (
                    id SERIAL PRIMARY KEY,
                    data1 INT,
                    data2 TEXT
                ) USING orioledb;',
                i);
        END LOOP;
    END $$;
```

In addition to crashing during the CREATE loop, **segfaults also occasionally appear during**:

```
DROP EXTENSION orioledb CASCADE;
```

—but this only when the number of created tables is around **600+**.

---

### **Reproduction script**

```sql
CREATE SCHEMA indices;
SET SESSION search_path = 'indices';
CREATE EXTENSION orioledb;
SELECT orioledb_parallel_debug_start();

-- create 2000 tables in a loop
DO $$
DECLARE
    i INT;
BEGIN
    FOR i IN 1..2000 LOOP
        EXECUTE format('
            CREATE TABLE indices.tbl_idx_%s (
                id SERIAL PRIMARY KEY,
                data1 INT,
                data2 TEXT
            ) USING orioledb;',
            i);
    END LOOP;
END $$;

SELECT orioledb_parallel_debug_stop();
DROP EXTENSION orioledb CASCADE;
DROP SCHEMA indices CASCADE;
RESET search_path;
```

---

### **Notes**

The issue likely stems from internal Orioledb catalog, metadata, or memory-tracking structures that do not scale to large numbers of relations created in a single session/transaction. Potential candidates include:

* relation metadata allocation,
* undo/meta buffers,

PR #675, opened 2025-12-04
Extend UndoStackItem.itemSize from u16 to u48 to avoid overflow and segfault

Serge-sudo:
Previously itemSize used only 16 bits and could overflow when many undo items or relation-level operations occurred in a single transaction. The field is now expanded to 48 bits (u32 + u16).

The high 16 bits are placed into the former padding at the end of the struct, allowing backward compatibility when the padding was zeroed.

Also fixed dynamic buffer expansion: previously the buffer only doubled in size, which was insufficient when itemSize became large, leading to possible buffer overflow and memory corruption.

Fix for issue #673 
We had a significant out-of-bounds memory access issue here, which could have corrupted any memory in the system. This fix may also resolve other problems or segmentation faults that we experienced previously because of unknown reasons.

As of now I haven’t worked with OrioleDB’s undo-log system in depth, so my understanding is still limited. In case `UndoStackItem` is written to disk and we need to maintain compatibility with older undo logs, I placed the new field into the structure’s padding area. This way we should remain backward-compatible with existing undo records (assuming the memory was zero-initialized before use).

### DIFF
--- a/src/btree/undo.c
+++ b/src/btree/undo.c
@@ -318,7 +318,7 @@ make_undo_record(BTreeDescr *desc, OTuple tuple, bool is_tuple,
 	item = (BTreeModifyUndoStackItem *) get_undo_record(desc->undoType,
 														&undoLocation,
 														MAXALIGN(size));
-	item->header.itemSize = size;
+	UNDO_SET_ITEM_SIZE(&item->header, size);
 	if (action == BTreeOperationLock)
 		item->header.type = RowLockUndoItemType;
 	else
@@ -391,7 +391,7 @@ make_waiter_undo_record(BTreeDescr *desc, OInMemoryBlkno blkno, int pgprocno,
 	item = (BTreeModifyUndoStackItem *) get_undo_record(desc->undoType,
 														&undoLocation,
 														MAXALIGN(size));
-	item->header.itemSize = size;
+	UNDO_SET_ITEM_SIZE(&item->header, size);
 	item->header.type = ModifyUndoItemType;
 	item->header.indexType = desc->type;
 	item->action = BTreeOperationInsert;
@@ -947,7 +947,7 @@ add_undo_relnode(ORelOids oldOids, ORelOids *oldTreeOids, int oldNumTreeOids,
 				 ORelOids newOids, ORelOids *newTreeOids, int newNumTreeOids,
 				 bool fsync)
 {
-	LocationIndex size;
+	Size size;
 	UndoLocation location;
 	RelnodeUndoStackItem *item;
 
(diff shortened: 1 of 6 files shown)


src/transam/undo.c
```diff
@@ -2533,15 +2533,15 @@ void
 o_add_rewind_relfilenode_undo_item(RelFileNode *onCommit, RelFileNode *onAbort,
 								   int nOnCommit, int nOnAbort)
 {
-	LocationIndex size;
+	Size size;
 	UndoLocation location;
 	RewindRelFileNodeUndoStackItem *item;
 
 	size = offsetof(RewindRelFileNodeUndoStackItem, rels) + sizeof(RelFileNode) * (nOnCommit + nOnAbort);
```
Serge-sudo, line 2540: Pretty much this was the source of overflow.

src/transam/undo.c
```diff
@@ -695,15 +695,15 @@ undo_item_buf_read_item(UndoItemBuf *buf,
...
 
 	ASAN_UNPOISON_MEMORY_REGION(buf->data, buf->length);
 	undo_read(undoType, location, sizeof(UndoStackItem), buf->data);
 
-	itemSize = ((UndoStackItem *) buf->data)->itemSize;
+	itemSize = UNDO_GET_ITEM_SIZE(((UndoStackItem *) buf->data));
 	if (itemSize > buf->length)
 	{
-		buf->length *= 2;
+		buf->length = pg_nextpower2_32(itemSize);
```
Serge-sudo, line 706: The buffer wasn’t large enough to hold itemSize. For example, with itemSize = 2 * 2048 + 1, we would already be accessing one byte outside our allocated memory.

src/btree/undo.c
```diff
@@ -318,7 +318,7 @@ make_undo_record(BTreeDescr *desc, OTuple tuple, bool is_tuple,
 	item = (BTreeModifyUndoStackItem *) get_undo_record(desc->undoType,
 														&undoLocation,
 														MAXALIGN(size));
-	item->header.itemSize = size;
+	UNDO_SET_ITEM_SIZE(&item->header, size);
```
pashkinelfe, line 321: Probably definition above should be changed
from ```LocationIndex size;```
to ```Size size;```

src/btree/undo.c
```diff
@@ -391,7 +391,7 @@ make_waiter_undo_record(BTreeDescr *desc, OInMemoryBlkno blkno, int pgprocno,
...
```
pashkinelfe, line 394: Probably definition above should be changed
from ```LocationIndex size;```
to ```Size size;```

PR #679, opened 2025-12-10
Oriole systrees visibility on logical decoding

GremSnoort:
Improve logical decoding correctness for OrioleDB catalogs and add system undo retention for replication
-----------------------------------------------

Logical decoding in OrioleDB must reconstruct table and index descriptors
that are consistent with the MVCC snapshot at the WAL position being
decoded. Previously, descriptor resolution was not fully snapshot- or
version-aware, which could lead to schema mismatches, missing historical
descriptors, or incorrect decoding results.

This patch introduces snapshot-aware descriptor fetching, extends WAL
metadata required for schema resolution, and adds safeguards to detect
cases where DML cannot be reconstructed due to undo cleanup. It also
ensures that system undo needed for logical decoding is retained while
replication slots depend on it.

Snapshot- and version-aware descriptor fetching
-----------------------------------------------

Introduce a unified OTableFetchContext used when resolving catalog
objects and provide new extended APIs:

    o_fetch_table_descr_extended()
    o_fetch_index_descr_extended()

These APIs allow callers to request descriptors corresponding to a
specific snapshot and schema version, enabling MVCC-correct reconstruction
during decoding and recovery.

OTable now tracks versions of special indexes (primary, toast, bridge)
so that OTableDescr construction can pin the correct index descriptors.

WAL relation records are extended with additional metadata required to
resolve the proper schema version during decoding.

DDL WAL ordering barrier
-----------------------

OrioleDB WAL records are buffered locally and flushed independently from
PostgreSQL WAL, which may cause reordering between PG-native WAL records
and Oriole WAL. Such reordering is unsafe for logical decoding around DDL.

To guarantee correct ordering, a DDL barrier is introduced in
orioledb_utility_command() that forces a local WAL flush before executing
utility commands.

Skipped DML detection
--------------------

Undo chains may be cleaned during transaction abort/rollback, making it
impossible for the logical decoder to reconstruct certain changes.

To detect these situations, maintain a TopMemoryContext cache tracking
transactions for which DML was skipped.

Behavior:

- skipped DML during COMMIT is treated as an error
- skipped DML during ROLLBACK/ABORT is allowed

The cache survives multiple decoder invocations.

System undo retention for replication
------------------------------------

Add a new system tree SYS_TREES_XID_UNDO_LOCATION mapping PostgreSQL
transaction ids (xid) to UndoLocation values representing the earliest
system undo required to reconstruct catalog state.

When computing minimal undo retention, logical replication requirements
are now considered. The system undo log will not be truncated past the
location required by active replication slots.

To reduce lookup overhead, introduce CachedReplicationRetainUndoTuple,
a fast-path cache for repeated xmin queries.

Behavior and compatibility
-------------------------

Existing behavior is preserved by allowing callers to use
O_TABLE_INVALID_VERSION together with a default snapshot when versioned
resolution is not required.

These changes improve logical decoding stability, recovery correctness,
and schema visibility without introducing user-visible incompatibilities.


src/catalog/ddl.c
```diff
@@ -2427,20 +2427,18 @@ drop_bridge_index(Relation tbl, OTable *o_table)
 	int			ix_num = InvalidIndexNumber;
 
 	old_o_table = o_table;
-	o_table = o_tables_get(o_table->oids);
+	o_table = o_tables_get(o_table->oids, NULL, NULL);
 	o_table->index_bridging = false;
-	o_table->bridge_oids.datoid = InvalidOid;
-	o_table->bridge_oids.reloid = InvalidOid;
-	o_table->bridge_oids.relnode = InvalidOid;
+	o_table->bridge_oids = create_oids();
```
pashkinelfe: Could we use existing ORelOidsSetInvalid() ?

src/catalog/o_tables.c
```diff
@@ -1163,9 +1173,9 @@ o_tables_get_by_oids_and_version(ORelOids oids, uint32 *version)
  * Find OTable by its oids
  */
 OTable *
-o_tables_get(ORelOids oids)
+o_tables_get(ORelOids oids, OSnapshot *snapshot, uint32 *version)
```
pashkinelfe: Propose to rename o_tables_get_by_oids_and_version() to o_tables_get() and leave only one function.
Both functions are equivalent after this PR

src/recovery/wal.c
```diff
@@ -158,6 +158,12 @@ wal_parse_rec_relation(Pointer ptr, uint8 *treeType, ORelOids *oids)
 	PARSE(ptr, &oids->reloid);
 	PARSE(ptr, &oids->relnode);
 
+	PARSE(ptr, xmin);
```
pashkinelfe: Additional reads should be under if (wal_version >= 17)
We bumped CURRENT_WAL_VERSION to 17 after beta13 and don't need to bump it again yet. But if we issue beta14 before this patch then we'll need to bump CURRENT_WAL_VERSION to 18 and use if (wal_version >= 18) here.

include/recovery/wal.h
```diff
@@ -103,6 +103,10 @@ typedef struct
 	uint8		datoid[sizeof(Oid)];
 	uint8		reloid[sizeof(Oid)];
 	uint8		relnode[sizeof(Oid)];
+	uint8		xmin[sizeof(OXid)]; /* @TODO @INFO */
```
pashkinelfe: /* Since ORIOLEDB_WAL_VERSION = 17 */

src/recovery/wal.c
```diff
@@ -728,6 +733,19 @@ add_rel_wal_record(ORelOids oids, OIndexType type)
 	memcpy(rec->reloid, &oids.reloid, sizeof(Oid));
 	memcpy(rec->relnode, &oids.relnode, sizeof(Oid));
 
+	runXmin = pg_atomic_read_u64(&xid_meta->runXmin);
```
pashkinelfe, line 778: /* Since ORIOLEDB_WAL_VERSION = 17 */

src/recovery/logical.c
```diff
@@ -782,8 +783,14 @@ orioledb_decode(LogicalDecodingContext *ctx, XLogRecordBuffer *buf)
 		{
 			int			sys_tree_num = -1;
 			uint8		treeType = 0;
+			OXid		xmin;
+
+			OSnapshot	snapshot;
+
+			ptr = wal_parse_rec_relation(ptr, &treeType, &cur_oids, &xmin, &snapshot.csn, &snapshot.cid, &cur_version);
```
pashkinelfe: Could get undifined values of xmin, snapshot.csn, snapshot.cid if reading WAL_REC_RELATION with wal_version < 17

include/orioledb.h
```diff
@@ -201,6 +201,18 @@ typedef struct
 	Oid			relnode;
 } ORelOids;
 
+static inline ORelOids
+create_oids(void)
```
pashkinelfe: to delete: duplicate for ORelOidsSetInvalid()

src/catalog/o_tables.c
```diff
@@ -253,7 +262,8 @@ ToastAPI	oTablesToastAPI = {
 	.getTupleChunknum = oTablesGetTupleChunknum,
 	.getTupleDataSize = oTablesGetTupleDataSize,
 	.deleteLogFullTuple = false,
-	.versionCallback = oTablesVersionCallback
+	.versionCallback = oTablesGetTupleKeyVersion,
```
pashkinelfe: It is possible to combine versionCallback and fetchCallback into one?

src/catalog/o_tables.c
```diff
@@ -253,7 +262,8 @@ ToastAPI	oTablesToastAPI = {
...
+	.fetchCallback = oTablesVersionCallback
```
pashkinelfe: Maybe rename oTablesVersionCallback -> oTablesFetchCallback?

src/tableam/descr.c
```diff
@@ -1581,7 +1583,7 @@ recreate_table_descr_by_oids(ORelOids oids)
 		recreate_table_descr(descr);
 	}
 	else
-		(void) create_table_descr(oids);
+		(void) create_table_descr(oids, NULL, NULL);
```
pashkinelfe: won't table descriptor without version and snapshot here break anything?

include/catalog/o_tables.h
```diff
@@ -153,10 +153,10 @@ extern void o_tables_truncate_all_unlogged(void);
 extern bool o_tables_add(OTable *table, OXid oxid, CommitSeqNo csn);
 
 /* Returns OTable by its oids */
-extern OTable *o_tables_get(ORelOids oids);
+extern OTable *o_tables_get(ORelOids oids, OSnapshot *snapshot, uint32 *version);
```
za-arthur: I think it makes sense to change order of arguments to make it consistent to `o_tables_get_by_oids_and_version`.
And also you can just use value instead of pointer:
```suggestion
extern OTable *o_tables_get(ORelOids oids, uint32 version, OSnapshot *snapshot);
```

include/catalog/o_tables.h
```diff
@@ -153,10 +153,10 @@ extern void o_tables_truncate_all_unlogged(void);
...
 
 /* Returns OTable by its oids and version */
-extern OTable *o_tables_get_by_oids_and_version(ORelOids oids, uint32 *version);
+extern OTable *o_tables_get_by_oids_and_version(ORelOids oids, uint32 *version, OSnapshot *snapshot);
```
za-arthur: Same as above,  it makes sense to use value for the version argument instead of pointer:

```suggestion
extern OTable *o_tables_get_by_oids_and_version(ORelOids oids, uint32 version, OSnapshot *snapshot);
```

akorotkov:
Also, regarding `o_table_get()`.  I think it doesn't make sense to have `o_tables_get()` and `o_tables_get_by_oids_and_version()` with the same set of arguments.  I propose to leave `o_tables_get()` with the same signature as now (just ORelOids as an argument), but change `o_tables_get_by_oids_and_version()` to `o_tables_get_extended()` and add OSnapshot there.

src/catalog/o_tables.c
pashkinelfe: @akorotkov has better proposal on this https://github.com/orioledb/orioledb/pull/679#issuecomment-3650699355

src/catalog/ddl.c
GremSnoort: Yes, it is fixed, create_oids removed and changed to ORelOidsSetInvalid

include/orioledb.h
GremSnoort: Fixed

src/catalog/o_tables.c
GremSnoort: Renamed: oTablesVersionCallback -> oTablesFetchCallback

src/catalog/o_tables.c
GremSnoort: Yes, callbacks merged into one (`fetchCallback`), but it's still necessary to have `.getTupleKeyVersion = oTablesGetTupleKeyVersion`, for WAL records

include/catalog/o_tables.h
GremSnoort: `o_tables_get` reverted to its original version: by oids only

include/catalog/o_tables.h
```diff
@@ -67,6 +67,7 @@ typedef struct
 	ORelOids	oids;
 	OIndexType	type;
 	OCompress	compress;
+	uint32		version;
```
kostja, line 71: Please add a comment explaining the purpose of the new member.

include/catalog/o_tables.h
```diff
@@ -89,6 +90,9 @@ typedef struct
 {
 	ORelOids	oids;
 	ORelOids	toast_oids;
+	uint32		toast_ixversion;
+	uint32		primary_ixversion;
+	uint32		bridge_ixversion;
```
kostja: Ditto.

include/catalog/sys_trees.h
```diff
@@ -144,5 +145,6 @@ extern PrintFunc sys_tree_key_print(BTreeDescr *desc);
 extern PrintFunc sys_tree_tup_print(BTreeDescr *desc);
 extern void sys_tree_set_extra(int tree_num, Pointer extra);
 extern Pointer sys_tree_get_extra(int tree_num);
+extern const text *retrieve_orioledb_sys_tree_structure(int systree, int depth);
```
za-arthur: I propose to have consistent name for the function:

```suggestion
extern const text *get_sys_tree_structure(int systree, int depth);
```

za-arthur: On the other hand it isn't used anywhere and can be removed?

src/catalog/o_indices.c
```diff
@@ -1102,13 +1218,27 @@ o_indices_add(OTable *table, OIndexNumber ixNum, OXid oxid, CommitSeqNo csn)
 	Pointer		data;
 	int			len;
 	BTreeDescr *sys_tree;
+	uint32	   *version_ptr;
 
 	oIndex = make_o_index(table, ixNum);
+	version_ptr = get_version(table, ixNum);
+	if (version_ptr)
+		*version_ptr = 0;
```
za-arthur: This code looks a bit unusual. I propose to rename the function into `get_version` and do not return pointer to the version field.

src/catalog/o_indices.c
```diff
@@ -1144,24 +1282,42 @@ o_indices_del(OTable *table, OIndexNumber ixNum, OXid oxid, CommitSeqNo csn)
...
+											&snapshot,
+											get_sys_tree(SYS_TREES_O_INDICES),
+											(Pointer *) &found_key);
 
 	if (result == NULL)
 		return NULL;
 
 	oIndex = deserialize_o_index(&key, result, dataLength);
 	pfree(result);
+	pfree(found_key);
```
za-arthur, line 1401: How do we know that `found_key` was palloc'ed and it isn't the same pointer to the local variable `key`. If it wasn't palloc'ed we shouldn't free it.

src/catalog/o_indices.c
```diff
@@ -1176,9 +1332,11 @@ o_indices_update(OTable *table, OIndexNumber ixNum, OXid oxid, CommitSeqNo csn)
 	Pointer		data;
 	int			len;
 	BTreeDescr *sys_tree;
+	uint32		version;
 
 	oIndex = make_o_index(table, ixNum);
-	oIndexOld = o_indices_get(oIndex->indexOids, oIndex->indexType);
+	version = oIndex->indexVersion == O_TABLE_INVALID_VERSION ? O_TABLE_INVALID_VERSION : oIndex->indexVersion - 1;
```
za-arthur: What does happen if `oIndex->indexVersion` is `0`?

src/recovery/logical.c
```diff
@@ -505,6 +505,9 @@ o_decode_modify_tuples(ReorderBuffer *reorderbuf, uint8 rec_type, OIndexType ix_
 	}
 }
 
+static ORelOids cur_oids = {0, 0, 0};
+static uint32 cur_version = O_TABLE_INVALID_VERSION;
```
za-arthur: Why aren't these variables defined as local variables within `orioledb_decode`?

src/recovery/logical.c
```diff
@@ -782,13 +786,39 @@ orioledb_decode(LogicalDecodingContext *ctx, XLogRecordBuffer *buf)
...
+
+			snapshot.xmin = xmin;
+			snapshot.xlogptr = changeXLogPtr;
+
+			if (wal_version < 17)
+			{
+				/* Override undefined values from WAL */
+				snapshot = o_non_deleted_snapshot;
+				latest_version = O_TABLE_INVALID_VERSION;
+			}
```
za-arthur: I propose to move this into `wal_parse_rec_relation`.

src/recovery/logical.c
```diff
@@ -782,13 +786,39 @@ orioledb_decode(LogicalDecodingContext *ctx, XLogRecordBuffer *buf)
...
 
 			ix_type = treeType;
 			relreplident = REPLICA_IDENTITY_DEFAULT;
 			descr = NULL;
 
+			if (ix_type == oIndexInvalid)
+			{
+				cur_oids = latest_oids;
+				cur_version = latest_version;
```
za-arthur: Do `cur_oids` and `cur_version` need to be invalidated in other cases since they are global variables? Otherwise they might store values from a pervious relation.

src/tuple/toast.c
```diff
@@ -873,23 +892,37 @@ generic_toast_get_any(ToastAPI *api, void *key, Size *data_size,
...
 	tkey.attnum = attn;
 	tkey.chunknum = 0;
 
 	Assert(toast->desc.type == oIndexToast);
 
-	result = generic_toast_insert(&tableToastAPI, &tkey, data,
+	if (desc->storageType == BTreeStoragePersistence)
+		/* @TODO @ CHECK && wal == true */
+	{
+		add_rel_wal_record(descr->oids, oIndexInvalid, descr->version);
```
za-arthur: I think that this need to be moved into either `add_modify_wal_record` or `generic_toast_insert_optional_wal` to add WAL_REC_RELATION only if the record was inserted.

src/tuple/toast.c
```diff
@@ -914,21 +947,32 @@ o_toast_sort_add(OIndexDescr *primary, OIndexDescr *toast,
...
 	tkey.attnum = attn;
 	tkey.chunknum = 0;
 
 	Assert(toast->desc.type == oIndexToast);
 
-	result = generic_toast_delete(&tableToastAPI, (Pointer) &tkey,
+	if (desc->storageType == BTreeStoragePersistence)
+		/* @TODO @ CHECK && wal == true */
+	{
+		add_rel_wal_record(descr->oids, oIndexInvalid, descr->version);
```
za-arthur: Same as above

src/catalog/o_tables.c
```diff
@@ -678,6 +691,9 @@ o_table_tableam_create(ORelOids oids, TupleDesc tupdesc, char relpersistence,
 	o_table->fillfactor = fillfactor;
 	o_table->persistence = relpersistence;
 	o_table->data_version = ORIOLEDB_DATA_VERSION;
+	o_table->toast_ixversion = UINT32_MAX;
+	o_table->primary_ixversion = UINT32_MAX;
+	o_table->bridge_ixversion = UINT32_MAX;
```
za-arthur: Is it intentional that initial value of `*_ixversion` is `UINT32_MAX`? On the next usage it will be wrapped around:

https://github.com/orioledb/orioledb/blob/oriole_systrees_visibility_on_logical_decoding_with_retention/src/catalog/o_indices.c#L553-L553

```C
result->indexVersion = ++table->primary_ixversion;
```

Wouldn't it be more correct and easier to understand, first to initialize them to zero and the post-increment, if you want them to start strictly from `0`?

```C
result->indexVersion = table->primary_ixversion++;
```

src/recovery/logical.c
GremSnoort: Removed in 40bc02fa29b378ac1b09e0ad1df1299b04aa6c36

src/recovery/logical.c
GremSnoort: Removed in 40bc02fa29b378ac1b09e0ad1df1299b04aa6c36

src/tuple/toast.c
GremSnoort: This delta is outdated, see 40bc02fa29b378ac1b09e0ad1df1299b04aa6c36
This was intended to solve dependency in logical decoder between TOAST relation and its base/main/heap relation.
Info:
Logical decoder receives TOAST relation's `WAL_REC_RELATION` with oids and version of TOAST relation. Then, TOAST relation index descr is retrieved from O_INDEX, then, base relation descr is retrieved from O_TABLE. Oids for the base relation we know from TOAST relation index descr's `tableOids`, but the version of base rel - is unknown.
To solve this problem, we have two ways:
1) add one more `WAL_REC_RELATION` record with base relation before TOAST rel - this may cause problems on local WAL buffer flushing when overflow, if it becomes separated from TOAST rel record - and this is an outdated delta
2) extend `WAL_REC_RELATION` with `base_version` field - I used this way for now

src/tuple/toast.c
GremSnoort: This delta is outdated, see 40bc02fa29b378ac1b09e0ad1df1299b04aa6c36 and extended answer above

src/tableam/descr.c
```diff
@@ -101,6 +101,21 @@ static MemoryContext descrCxt = NULL;
...
+ORelFetchContext
+default_in_progress_fetch_context(void)
+{
+	ORelFetchContext ctx = {.snapshot = o_in_progress_snapshot,.version = O_TABLE_INVALID_VERSION};
+
+	return ctx;
+}
+
+ORelFetchContext
+default_non_deleted_fetch_context(void)
```
pashkinelfe: Maybe it's better to make it extern variable, not a function?

src/tableam/descr.c
```diff
@@ -101,6 +101,21 @@ static MemoryContext descrCxt = NULL;
 
 static void o_find_toastable_attrs(OTableDescr *tableDescr);
 
+ORelFetchContext
+default_in_progress_fetch_context(void)
```
pashkinelfe: Maybe it's better to make it extern variable, not a function?

src/catalog/o_indices.c
```diff
@@ -951,8 +946,53 @@ cache_scan_tupdesc_and_slot(OIndexDescr *index_descr, OIndex *oIndex)
...
+
+OIndexDescrFillSource
+fill_idescr_from_ctx(ORelFetchContext *base_ctx_ptr)
+{
+	Assert(base_ctx_ptr);
+	OIndexDescrFillSource output = {.get_oTable = provider_loaded,.arg = base_ctx_ptr};
+
+	return output;
+}
+
```
pashkinelfe: IMO this concept of providers/callbacks looks quite complicated for implementing just two variants of getting OTable. Is it meant for further extension?

GremSnoort: I suppose no, because there may be only two invariants: ready OTable (provider_external) and loaded OTable (provider_loaded).

pashkinelfe: Then for simplicity and clarity of the code I propose then to use two-variants functions with adding just a bool value to distinguish differences for ready and loaded OTable.

src/tableam/descr.c
GremSnoort: Done in 1016a8724be4b498a4fe85007b0c1ee2ad904c4b

src/tableam/descr.c
GremSnoort: Done in 1016a8724be4b498a4fe85007b0c1ee2ad904c4b

src/catalog/o_indices.c
GremSnoort: I propose a way to simplify the interface in commit c42383c300194c63ecd1f5d61fcf062c69dbd5f0

src/catalog/o_indices.c
GremSnoort: In this case, `pfree(found_key)` is actually permissible due to the ownership contract between `o_indices_get_extended()` and `generic_toast_get_any_with_key()`/`common()`.

Initially, `found_key = &key;` -  points to a stack object.
Next, the double pointer is passed to the method:
`generic_toast_get_any_with_key(..., (Pointer *)&found_key);`
and at the end, if this method returned `result != NULL`, `found_key` is freed:
`pfree(found_key);`

In `generic_toast_get_any_common()`, when the first chunk is found:
```c
if (found_key)
{
    Size key_size = api->getKeySize(arg);
    *found_key = palloc(key_size);            // we overwrite the pointer
    memcpy(*found_key, tuple.data, key_size); // found_key is now a heap object
}
```
`found_key` is an in/out parameter. Caller may pass a pointer to any temporary key object as an initial value. On the first successfully fetched chunk, `generic_toast_get_any_common()` replaces `*found_key` with a freshly `palloc()`'d copy of the tuple key. Therefore, upon success, `*found_key` always points to heap memory owned by the caller and must be freed with `pfree()`. On failure (no tuples found), the function returns NULL and does not allocate `*found_key`.

And therefore `pfree(found_key)` in the calling code is correct in the successful path, because by the time `pfree` is called, `found_key` is **no longer** `&key`, but **a heap pointer**.

This results in an implicit dependency of `found_key` on the value of `result`:
If `result == NULL` => no `palloc()` so no free
If `result != NULL` => needs `palloc()`

This may not be the best design, but in this case it is done by analogy with the mechanics in `o_tables_get_extended()`, which uses exactly the same approach.

include/catalog/sys_trees.h
GremSnoort: This method is very useful for debugging, and I have used it in some places in the code to monitor the contents of system trees. Previously, this code snippet was shared in chat rooms in response to the question: _how can I check the status and contents of system trees in the right place?_ I would like to introduce this method to the code base so that this question does not arise.

pashkinelfe: I propose naming consistent with debugging semantics e.g. inspect_sys_tree_structure, debug_sys_tree_structure, or debug_inspect_sys_tree_structure etc.

If it's not called from the code now, I'd propose to have this function under #ifdef IS_DEV like the other parts of the code that are useful for debugging and CI, but unneeded in prod build.

src/catalog/o_indices.c
pashkinelfe: @GremSnoort Thank you!

 I'd propose a couple of additional commits that further reduce complexity 
https://github.com/orioledb/orioledb/commit/e4af12169c243d3a3ce66162b6035a19847bc752
https://github.com/orioledb/orioledb/commit/5ccc896126b40d287a78b95c474622dfff670fa4

pashkinelfe: And a third one: https://github.com/orioledb/orioledb/commit/253083f8c34d6cc7d000671438e1dd8dd201e120

src/tuple/toast.c
```diff
@@ -445,8 +464,13 @@ generic_toast_insert_optional_wal(ToastAPI *api, void *key, Pointer data,
 		}
 
 		if (desc->storageType == BTreeStoragePersistence && wal)
+		{
+			uint32		version = GET_VERSION(api, getBTreeVersion, arg);
```
pashkinelfe: I'd propose following minor changes:
https://github.com/orioledb/orioledb/commit/5dc30ad9ea851146aa05f0cb3a9d13056ad4c611
https://github.com/orioledb/orioledb/commit/a537bf99dfd1fedcbdb36b3d581d2e04f0fec50d
https://github.com/orioledb/orioledb/commit/53088ff0766b528a462d1404944c568a01659c1b
https://github.com/orioledb/orioledb/commit/1f575548ec1afcfc46f9931e894a75c150c7f68b

pashkinelfe (commented):
I propose following minor change:
https://github.com/orioledb/orioledb/commit/5dc30ad9ea851146aa05f0cb3a9d13056ad4c611

src/catalog/o_indices.c
za-arthur: @GremSnoort thank you for the explanation!

I see, the `generic_toast_get_any_with_key` function logic was already here. I think it is worth to refactor it a bit, otherwise it is hard to follow. But this is definitely not the target of this PR and it needs to be done separately.

src/tuple/toast.c
GremSnoort: Accepted

src/catalog/o_indices.c
GremSnoort: Accepted
May I change for this e4af12169c243d3a3ce66162b6035a19847bc752 (_Simplify getting table and index descrs from ready OTable or from ORelFetchContext_)
`bool source_is_context` to enum?
```c
typedef enum
{
	oTableSourceTable = 0,
	oTableSourceCntext = 1
} OTableSource;
```

src/catalog/o_indices.c
GremSnoort: Updated in acf74a9db5d2aae0feab3cb7f7de1eb3e937457b

src/catalog/o_tables.c
GremSnoort: Yes, wrapping around was intentional, but this is improved in acf74a9db5d2aae0feab3cb7f7de1eb3e937457b

src/catalog/o_indices.c
GremSnoort: Done

include/catalog/sys_trees.h
GremSnoort: Renamed `retrieve_orioledb_sys_tree_structure` -> `inspect_sys_tree_structure` and hidden under `#ifdef IS_DEV`
Commit 6083e8f82177119258f34a1b86b1414d5e011519

src/tuple/toast.c
pashkinelfe: Thank you!

include/catalog/o_tables.h
GremSnoort: Comments added in f20f09f357700632bb800127d55bd3b0d3597e70

src/catalog/o_indices.c
GremSnoort: Outdated, please see the latest commits.

src/recovery/recovery.c
```diff
@@ -2779,9 +2779,11 @@ handle_o_tables_meta_unlock(ORelOids oids, Oid oldRelnode)
 
 		if (!OidIsValid(oldRelnode))
 		{
-			uint32		version = (new_o_table->version == O_TABLE_INVALID_VERSION) ? O_TABLE_INVALID_VERSION : new_o_table->version - 1;
+			uint32		version;
 
-			/* @TODO ! !! */
+			Assert(new_o_table->version > 0);
```
pashkinelfe: As version is unsigned, we'd better do Assert(new_o_table->version != 0);

GremSnoort: Done in 4e3d5cf80202df9c27102d2f42a17fabc76a8359

src/recovery/logical.c
GremSnoort: `wal_parse_rec_relation` now sets all input variables:
- assigns the value parsed if the WAL version meets the requirements
- otherwise assigns default values

`snapshot` is a local variable for logical decoder so its initialization stays in `orioledb_decode`

pashkinelfe (commented):
This CI flake crash is also reproduced on main
https://github.com/orioledb/orioledb/actions/runs/19306024951/job/55213676749?pr=645#step:10:58
https://github.com/orioledb/orioledb/actions/runs/21474847760/job/61856183080

Likely it's not related to this PR and doesn't block it. I propose removing commit https://github.com/orioledb/orioledb/pull/679/changes/e24b505758aeaebb25c58f4574e11ec2c03b0561 and test/fix this condition after merging this PR

Issue #657, opened 2025-11-26
Incorrect Hash Index Behavior on Tuple Update in OrioleDb table


xsion07:
Hi everyone!

While analyzing the execution of the vanilla regress test using a hash index on an orioledb test table (main test: postgres/src/test/regress/sql/hash_index.sql), I encountered several issues related to tuple updates and index consistency.
Below are the problems I’ve identified, along with reproduction steps, analysis, and a proposed fix.

### **Issue 1:** Duplicate Rows Returned After UPDATE (No PK Defined)

```
CREATE TEMP TABLE hash_i4_heap (
    seqno   int4,
    random  int4
) USING orioledb;

\set filename :abs_srcdir '/data/hash.data'
COPY hash_i4_heap FROM :'filename';

CREATE INDEX hash_i4_index ON hash_i4_heap USING hash (random int4_ops);

UPDATE hash_i4_heap
   SET seqno = 20000
 WHERE random = 1492795354;
```
After the UPDATE, querying the row via index scan returns two identical rows:

```
SELECT h.seqno AS i20000
  FROM hash_i4_heap h
 WHERE h.random = 1492795354;
Result:

 i20000 
--------
  20000
  20000
(2 rows)
```

**Root Cause:**
When updating a tuple without a PRIMARY KEY, **o_tbl_update** calls **o_tbl_indices_overwrite**, which rewrites the tuple in-place (same ctid).
The new tuple overwrites the old one, but the ctid remains unchanged.
During the hash index update, **hashinsert** adds a new index entry pointing to the same ctid.
However, **hashinsert** does not check for duplicates, so two index entries now point to the same ctid (one from before the update, one after).
On subsequent Index Scan, both entries are returned → duplicate rows.

This does not happen in heap tables, because UPDATE marks the old row as dead and inserts a new one with a new ctid, avoiding index duplication.

### **Issue 2:** Zero Rows Returned After UPDATE (PK Defined)
```
CREATE TEMP TABLE hash_i4_heap (
    seqno   int4 PRIMARY KEY,
    random  int4
) USING orioledb;

\set filename :abs_srcdir '/data/hash.data'
COPY hash_i4_heap FROM :'filename';

CREATE INDEX hash_i4_index ON hash_i4_heap USING hash (random int4_ops);

UPDATE hash_i4_heap
   SET seqno = 20000
 WHERE random = 1492795354;
```
After the update, the same SELECT returns zero rows:
```
SELECT h.seqno AS i20000
[...]

PR #681, opened 2025-12-15
Fix missing bridge index tuple, while update operation with PK fields

xsion07:
When updating a tuple in an orioledb table, the HASH/GIN index is not updated if indexed fields are unchanged, causing duplicates or missing rows. 
Fix: mark the hash index as touched (touched_indices) during any UPDATE in orioledb, regardless of whether indexed columns changed. This ensures apply_new_bridge_index_ctid is called and maintains index consistency. Additionally, ctid duplicate checks are added during hash index insertion.
Main discussion in Issue #657

### DIFF
--- a/src/tableam/operations.c
+++ b/src/tableam/operations.c
@@ -697,15 +697,8 @@ o_tbl_update(OTableDescr *descr, TupleTableSlot *slot,
 
 	if (descr->bridge)
 	{
-		List	   *indexIds;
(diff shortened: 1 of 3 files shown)


xsion07:
Fixed in ([c8590e](https://github.com/orioledb/orioledb/commit/c8590e8b0ccbce3ac7416e8963c05a8a7f97ffaf))

PR #684, opened 2025-12-18
Add make.flags file with default build parameters and import it in Makefile

pgnickb:
This enables one to adjust `make.flags` file (which is in `.gitignore`) to their needs.
This is backwards compatible. Variables supplied with `make` command take precedence.

### DIFF
--- a/.dockerignore
+++ b/.dockerignore
@@ -85,3 +85,6 @@ doc/
 
 # Exclude the PostGIS Docker build directory
 docker-postgis
+
+# Misc
+make.flags
--- a/.gitignore
+++ b/.gitignore
@@ -34,6 +34,7 @@ win32ver.rc
 *.exe
 lib*dll.def
 lib*.pc
+make.flags
 
 # Local excludes in root directory
 /test/t/__pycache__/
--- a/Makefile
+++ b/Makefile
@@ -1,4 +1,7 @@
 # contrib/orioledb/Makefile
+# Import make flags if make.flags file is present
+-include make.flags
+export
 
 MODULE_big = orioledb
 EXTENSION = orioledb
--- a/make.flags.example
+++ b/make.flags.example
@@ -1,0 +1,8 @@
+# Makefile checks if the variable is defined, the value is ignored
+# Remove/comment out the flag to deactivate it
+
+USE_PGXS=1
+
+IS_DEV=1
+
+#VALGRIND=1


pgnickb:
> Hi @pgnickb! it looks useful for local development workflows! (commenting as an external contributor who uses the orioledb Dockerfiles and is thinking about potential side effects)
> 
> I'd like to suggest adding the `.env` file to `.dockerignore` to prevent it from being copied into the Docker image during build.

Thanks for pointing it out. This is a very valid concern. Do you think we should just add it to `.dockerignore` or maybe also rename it to something else to avoid possible confusion?

pgnickb:
@ImreSamu Sorry for another force push. What do you think about the current state?

ImreSamu:
@pgnickb
Thank you!  The changes look good to me.

Note: While I am a long-time contributor, my experience with PostgreSQL internals 
and low-level C is limited. Please treat this as an informal review from an external contributor.

---

One minor personal note: as a newcomer who often works with AI coding assistants, 
I find it helpful when even small config files include brief comments explaining 
the "why" behind each option. For example, in `make.flags.example`:
- A short note about when you would or wouldn't want `IS_DEV=1`, or a link to a relevant markdown doc.
- A note that Docker builds (`./docker/Dockerfile*`) use hardcoded parameters 
  and ignore this file

Not a blocker, just a thought for future consideration.

I realise it is not easy to find the ideal commenting style that works for everyone: 
* { hardcore PostgreSQL devs ;  Copilot/AI agents for PR review ;  and newcomers alike }

pgnickb:
@ImreSamu 
> Note: While I am a long-time contributor, my experience with PostgreSQL internals and low-level C is limited. Please treat this as an informal review from an external contributor.

Sure. As long as the new file name doesn't suggest to you it is related to Docker infra I'm happy :-)
 
> One minor personal note: as a newcomer who often works with AI coding assistants, I find it helpful when even small config files include brief comments explaining the "why" behind each option. For example, in `make.flags.example`:
> 
>     * A short note about when you would or wouldn't want `IS_DEV=1`, or a link to a relevant markdown doc.
> 
>     * A note that Docker builds (`./docker/Dockerfile*`) use hardcoded parameters
>       and ignore this file

I agree in principle. The file merely provides means to specify the said flag and others in a different way. As in normally these flags are supplied to `make` manually for each invocation. Thus the intention here is to simplify `make` calls during development.

As for the nature of `IS_DEV`, it's a tricky one. It has effect on both SQL scripts as well as C source code. The idea is simple though: enable certain infra that is only needed for tests and is otherwise costly and pointless to run for production systems.

Existing contributing readme mentions the flag but doesn't go in detail as to what exactly it does because we don't want to make any promises here. We want to be able to adjust `IS_DEV` and friends as needed to suit the development needs.

I hope this answers your question. If you have an idea on how and where the above is best reflected, please don't hesitate to reach out. Or even better, file a PR ;-)

Issue #685, opened 2025-12-19
Temporary Tables Improvement


e-ivkov:
# Temporary Tables Improvement

Temporary tables performance issues is a well known problem in postgres community. It affects especially user's of 1C systems who are actively migrating from Oracle to Postgres. This ADR looks at OrioleDB implementation of temporary tables, checks if it has some overlapping problems with the heap storage and which perfrormance issues are unique to Oriole. Finally we suggest a plan on how to improve handling tmp tables in Oriole.

# Current State

### Dedicated Storage Type for Temporary Tables

OrioleDB has a specific storage type `BTreeStorageTemporary` for temporary tables. This is part of their storage hierarchy:

```c 
// orioledb/include/btree/btree.h#L71-80

{
	/* just in memory BTree, no eviction and no checkpoint support */
	BTreeStorageInMemory,
	/* no checkpoint support, but pages can be evicted into a disk */
	BTreeStorageTemporary,
	/* like BTreeStoragePersistence, but no wal for data modifications */
	BTreeStorageUnlogged,
	/* checkpoint and eviction for pages support */
	BTreeStoragePersistence
} BTreeStorageType;
```

The storage type influences which B-Tree is build: evictable/non-evictable, with/without checkpointer and wal support.

### Special Handling for Temporary Tables

When a table has `RELPERSISTENCE_TEMP` persistence level, OrioleDB:

- **Assigns the `BTreeStorageTemporary` storage type**: The table data can be kept in memory but can also be evicted to disk if needed (unlike `BTreeStorageInMemory` which is purely in-memory)
- **Skips WAL logging**: Temporary tables don't generate WAL records for data modifications
- **Skips checkpoint operations**: Temporary tables are not included in checkpoints

#### 1. **Shared Buffers Pool**

All pages, including temp table pages (also for TOAST relations), are stored in the global `o_shared_buffers`:

```c 
// orioledb/include/orioledb.h#L397-404

#define O_GET_IN_MEMORY_PAGE(blkno) \
	(AssertMacro(OInMemoryBlknoIsValid(blkno)), \
	 (Page)(o_shared_buffers + (((uint64) (blkno)) * ((uint64) ORIOLEDB_BLCKSZ))))
#define O_GET_IN_MEMORY_PAGEDESC(blkno) \
	(AssertMacro(OInMemoryBlknoIsValid(blkno)), page_descs + (blkno))
```

The `o_shared_buffers` is a **global shared memory area**:

```c 
[...]

PR #686, opened 2025-12-24
Introduce PagePoolOps interface

e-ivkov:
First stage of https://github.com/orioledb/orioledb/issues/685

This PR only concerns itself with introducing PagePoolOps and using this interface throughout the code. Locking is skipped for local pages. The actual implementation of PagePoolOps with local mempool will be introduced in the next PR.


include/orioledb.h
e-ivkov, line 423: blkno can't be negative yet and O_GET_IN_MEMORY_PAGE knows nothing about negative blkno at this stage. This will be changed in my next PR

src/btree/insert.c
```diff
@@ -1249,7 +1249,7 @@ o_btree_insert_item(BTreeInsertStackItem *insert_item, int reserve_kind)
 		int			tupleWaiterProcnums[BTREE_PAGE_MAX_SPLIT_ITEMS];
 		int			tupleWaitersCount;
 
-		Assert(desc->ppool->numPagesReserved[reserve_kind] >= 2);
+		/* Assert(desc->ppool->numPagesReserved[reserve_kind] >= 2); */
```
e-ivkov, line 1252: I commented out this assert as I am not sure that we should introduce a num_pages_reserved function to PagePoolOps just for the sake of one assert. I am open to suggestions

include/utils/page_pool.h
```diff
@@ -56,20 +100,12 @@ struct OPagePool
 	Size		ucmShmemSize;
 	/* seed for random values */
 	pg_prng_state prngSeed;
-};
+} OPagePool;
```
e-ivkov, line 103: Sadly I can't make OPagePool private as some macros use it's internal structure

e-ivkov:
Merged - https://github.com/orioledb/orioledb/commit/1f78088cb88ba57fb0e573a24d879d91a2c06c0e

PR #687, opened 2025-12-26
 fix wrong fit strategy when replaced tuple contains vacated space

ereutt:


akorotkov:
I spend some time analyzing page_locator_fits_item().  I decided that the whole function is crap.  I've re-written it in 55abca8a44f7cfbcf8c0d294fed9ced5d842bbd8.  Could you, please, check?

Issue #533, opened 2025-08-21
Add/enable support for the remaining ALTER TABLE subcommands


homper:
List of subcommands that are currently not supported
```
AT_AddColumnToView
AT_AddIndexConstraint
AT_AddOf
AT_AlterColumnGenericOptions
AT_AlterConstraint
AT_AttachPartition
AT_CheckNotNull
AT_ClusterOn
AT_DetachPartition
AT_DetachPartitionFinalize
AT_DisableTrig
AT_DisableTrigAll
AT_DisableTrigUser
AT_DropCluster
AT_DropOf
AT_DropOids
AT_EnableAlwaysTrig
AT_EnableReplicaTrig
AT_EnableTrig
AT_EnableTrigAll
AT_EnableTrigUser
AT_ForceRowSecurity
AT_NoForceRowSecurity
AT_ReAddComment
AT_ReAddConstraint
AT_ReAddDomainConstraint
AT_ReAddIndex
AT_ReAddStatistics
AT_ReplaceRelOptions
AT_ReplicaIdentity
AT_ResetOptions
AT_SetAccessMethod
AT_SetCompression
AT_SetExpression
AT_SetLogged
AT_SetOptions
AT_SetStatistics
AT_SetUnLogged
AT_ValidateConstraint
```

PR #696, opened 2026-01-23
Tests for remaining ALTER TABLE subcommands

vvtrunov:
AT_ReplicaIdentity is already implemented in master branch, so don't implement any tests.
AT_CheckNotNull is used for partitioned tables, but there is no difference between a partitioned heap table and orioledb, since postgres uses the heap access method anyway. Therefore, do not run any tests.
The AT_ReAdd* subcommands require further development, because it is unclear to me whether any more support needs to be added or whether everything should remain as it is.

In general, it is not entirely clear who is responsible for the correct execution of some commands: PostgreSQL or OrioleDB, so it is not entirely clear which part of the database tests they are written for. But some tests are already failing due to lack of support from OrioleDB.

Broken tests:
AT_SetCompression, AT_SetExpression, AT_AddIndexConstraint.

Related to #533.

### DIFF
--- a/test/expected/row_security.out
+++ b/test/expected/row_security.out
@@ -173,6 +173,135 @@ TABLE document;
  101 |  44 |      1 | regress_rls_dave  | testing sorting of policies
 (12 rows)
 
+-- Test AT_ForceRowSecurity and AT_NoForceRowSecurity
+-- These commands control whether table owners are subject to RLS policies
+-- Create a new table owned by carol with mixed security levels
+SET SESSION AUTHORIZATION regress_rls_carol;
+CREATE TABLE classified_docs (
+    did         int primary key,
+    dlevel      int not null,
+    dauthor     name,
+    dtitle      text
+) USING orioledb;
+GRANT ALL ON classified_docs TO public;
+-- Insert documents with varying security levels
+-- carol has seclv=2, so she can normally see dlevel <= 2
+INSERT INTO classified_docs VALUES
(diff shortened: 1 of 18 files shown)


test/expected/row_security.out
homper, line 1: should be covered by https://github.com/orioledb/postgres/blob/patches17/src/test/regress/sql/rowsecurity.sql
we should probably enable it in https://github.com/orioledb/postgres/blob/patches17/src/test/regress/parallel_schedule_oriole instead 
and row_security.sql might also be removed

test/expected/trigger.out
homper, line 1: Not sure, but looks like most of it should be covered by
https://github.com/orioledb/postgres/blob/patches17/src/test/regress/sql/triggers.sql
And even possibly remove trigger.sql file or most of it

test/expected/row_security.out
vvtrunov: currently rowsecurity postgres test does not work, so need to remove orioledb's rowsecurity later.

test/expected/trigger.out
vvtrunov: currently triggers.sql postgres test does not work, so need to remove orioledb's trigger.sql later.

vvtrunov:
Some commands that are internally generated by PostgreSQL (like AT_CheckNotNull) are fully handled by PostgreSQL part.
I am not sure, if it is necessary to add tests for such commands.

PR #699, opened 2026-01-28
Support AT_SetExpression

vvtrunov:
Related to #533.

Introduce a fix for recalculation of generated columns and fixed tests for DDL.
Also added test for matview over table with generated column

### DIFF
--- a/src/catalog/ddl.c
+++ b/src/catalog/ddl.c
@@ -108,7 +108,7 @@ static object_access_hook_type old_objectaccess_hook = NULL;
 List	   *drop_index_list = NIL;
 List	   *partition_drop_index_list = NIL;
 static List *alter_type_exprs = NIL;
-static List	*o_alter_generated_column_id = NIL;
+static List *o_alter_generated_column_id = NIL;
 Oid			o_saved_relrewrite = InvalidOid;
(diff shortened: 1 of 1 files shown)


src/catalog/ddl.c
```diff
@@ -3324,6 +3179,17 @@ orioledb_object_access_hook(ObjectAccessType access, Oid classId, Oid objectId,
...
 
+					/*
+					 * Alter table on generated column triggers table rewrite
+					 * due to need of recalculating column value for existing
+					 * rows
+					 */
+					if (old_field.generated)
+					{
+						in_rewrite = true;
+						o_alter_generated_column_id = subId;
```
homper: Will it work for multiple SET EXPRESSION in the same ALTER? It probably should
Like this:
```
ALTER TABLE o_test_set_expression
  ALTER COLUMN total SET EXPRESSION AS (price * quantity * 1.1),
  ALTER COLUMN total2 SET EXPRESSION AS (price * quantity * 0.5);
```
Just an example but it could also be runned for multiple different generated fields

vvtrunov: Thanks.

Fixed with collecting columns into List.

PR #700, opened 2026-01-28
Refactor OrioleDB WAL container parsing into dedicated wal_reader module

GremSnoort:
# Summary

This patch extracts the OrioleDB WAL container parsing logic into a dedicated, reusable module (`wal_reader.[ch]`) and migrates existing consumers (recovery, logical decoding, etc.) to a unified descriptor-driven parsing API.

The primary goal is to establish a **single source of truth for the WAL container binary protocol**, covering:
- container header (version + flags),
- flag prefix parsing,
- record tag/payload iteration,
- descriptor-based payload decoding,
- safe cursor advancement with strict bounds checking.

This eliminates duplicated parsing logic and reduces the risk of protocol desynchronization when extending WAL with new record types or flags.

# Motivation

Before this change, WAL container parsing logic was partially distributed across different parts of the codebase:
- container framing and cursor movement were handled in multiple places,
- record type dispatch was duplicated,
- flag handling was not structurally enforced,
- extension of the protocol required coordinated manual updates.

This creates typical binary protocol risks:
- adding a new record type or flag requires modifying multiple code paths,
- forgetting to parse a flag payload causes cursor misalignment,
- desynchronization errors manifest far from their root cause,
- it is difficult to reason about invariants related to cursor movement.

This patch consolidates parsing into a minimal, forward-only reader with strict responsibilities and explicit contracts.

# High-Level Design

## 1. New WAL Reader Module

Introduced:
- `include/recovery/wal_reader.h`
- `src/recovery/wal_reader.c`

This module implements a generic OrioleDB WAL container parser responsible strictly for:
- reading container header (version + flags),
- processing container-level flags,
- iterating over record stream [tag][payload]...,
- invoking descriptor-driven parse functions,
- delivering parsed records to consumer callbacks.

The reader:
- performs no memory allocation,
- performs no buffering,
- makes no semantic decisions about applying or decoding changes,
- operates in a strictly forward-only, single-pass manner.

All higher-level logic (recovery, logical decoding, policy decisions) is delegated to the consumer.

## 2. Descriptor-Driven Parsing (X-macro Tables)

Record types and container flags are now registered via X-macro lists:
- `ORIOLE_WAL_RECORDS(X)`
- `ORIOLE_WAL_FLAGS(X)`

These generate:
- `g_wal_descs[]`
- `g_wal_flag_descs[]`

This makes the descriptor tables the **single authoritative** mapping:
- type -> name (debugging)
- type -> payload parser

As a result:
- adding a new record type or flag is a **single-point change**,
- the parser cannot forget to handle a type,
- unknown types become explicit hard protocol errors,
- payload length is always derived from the descriptor contract.

If a descriptor is missing, parsing fails with `WALPARSE_BAD_TYPE`, preventing unsafe cursor advancement.

## 3. Unified Record Structure: WalRecord

`WalRecord` represents a parsed WAL entity (record or flag). This provides a uniform record abstraction independent of consumer logic.

## 4. Explicit Consumer Contract

The parsing API is consumer-driven via callbacks stored in `WalReaderState`:
- `check_version()` — container version compatibility policy,
- `on_flag()` — invoked for container-level flags,
- `on_record()` — invoked for each WAL record.

The parser itself is version-agnostic and policy-free. It only guarantees protocol correctness and safe cursor movement.

## 5. Deterministic Container Flag Handling

Container flags are stored on-wire as a bitmask but may have payloads.

This patch enforces:
- all flag payloads must be parsed before scanning record tags,
- flag parsing occurs in deterministic order defined by `ORIOLE_WAL_FLAGS`.

This guarantees:
- header-prefix bytes are fully consumed before record scanning,
- cursor alignment invariants are preserved,
- extending flags does not require manual updates in parsing logic.

## 6. Explicit Ownership Boundary for Tuple Payloads

Modify-type records expose tuple data as pointers into the WAL container buffer.

The parser does not allocate or copy tuple bytes.

This clarifies the boundary:
- parser = zero-allocation, zero-copy,
- consumer = responsible for stabilizing data if needed.

# Behavioral Impact

Protocol format is unchanged. However:
- header and flag handling is now centralized and mandatory,
- record scanning is descriptor-driven,
- adding new record types or flags requires only updating X-macro lists and defining a parse function,
- consumers no longer perform manual cursor arithmetic.

This improves safety and maintainability without altering WAL format semantics.

# Design Rationale

This design prioritizes:
- protocol safety,
- centralized framing logic,
- descriptor-driven extensibility,
- compile-time enforcement of type registration,
- explicit ownership semantics.

It intentionally avoids:
- implicit behavior,
- hidden parsing entry points,
- scattered cursor manipulation,
- duplication of binary protocol handling.

The result is a minimal yet sufficient API surface for WAL container parsing.

### DIFF
--- a/include/recovery/wal.h
+++ b/include/recovery/wal.h
@@ -14,41 +14,63 @@
 #ifndef __WAL_H__
 #define __WAL_H__
 
-/* WAL record types */
-#define WAL_REC_NONE		(0)
-#define WAL_REC_XID			(1)
-#define WAL_REC_COMMIT		(2)
-#define WAL_REC_ROLLBACK	(3)
-#define WAL_REC_RELATION	(4)
-#define WAL_REC_INSERT		(5)
-#define WAL_REC_UPDATE		(6)
-#define WAL_REC_DELETE		(7)
-#define WAL_REC_O_TABLES_META_LOCK (8)
-#define WAL_REC_O_TABLES_META_UNLOCK (9)
-#define WAL_REC_SAVEPOINT	(10)
-#define WAL_REC_ROLLBACK_TO_SAVEPOINT (11)
-#define WAL_REC_JOINT_COMMIT (12)
-#define WAL_REC_TRUNCATE	(13)
-#define WAL_REC_BRIDGE_ERASE (14)
-#define WAL_REC_REINSERT (15)	/* UPDATE with changed pkey represented as
-								 * DELETE + INSERT in OrioleDB but externally
-								 * exported as an UPDATE in logical decoding */
-#define WAL_REC_REPLAY_FEEDBACK	(16)
+#include "recovery/wal_reader.h"
 
 /*
- * Record type for a case when both heap and Oriole apply changes within a single transaction,
- * so one logical xid is assigned by heap, and the other is assigned by Oriole.
+ * WAL record types
  *
- * This record defines a connection between Oriole's sub-transaction xid and a xid of the top heap transaction
- * which is needed for logical decoder.
+ * WAL_REC_REINSERT :	UPDATE with changed pkey represented as DELETE + INSERT in OrioleDB but externally exported as an UPDATE in logical decoding
  *
- * Otherwise, without this connection, main transaction suddenly becomes splitted into two independent parts.
- * From logical decoder's point of view this looks like two independent transactions but in fact internally related to each other.
- * This situation outcomes in troubles for logical decoder with visibility of heap modifications
- * in Oriole's sub-part due to incorrect state of the MVCC-historical snapshot.
+ * WAL_REC_SWITCH_LOGICAL_XID :	Record type for a case when both heap and Oriole apply changes within a single transaction, so one logical xid is assigned by heap, and the other is assigned by Oriole.
+ * 								This record defines a connection between Oriole's sub-transaction xid and a xid of the top heap transaction which is needed for logical decoder.
+ *								Otherwise, without this connection, main transaction suddenly becomes splitted into two independent parts.
+ * 								From logical decoder's point of view this looks like two independent transactions but in fact internally related to each other.
+ * 								This situation outcomes in troubles for logical decoder with visibility of heap modifications
+ * 								in Oriole's sub-part due to incorrect state of the MVCC-historical snapshot.
  */
-#define WAL_REC_SWITCH_LOGICAL_XID	(17)
-#define WAL_REC_RELREPLIDENT		(18)
+
+#define WAL_REC_NONE           (0)
+
+#define ORIOLE_WAL_RECORDS(X) \
+	X(WAL_REC_XID,                   1, "XID",                wal_parse_rec_xid) \
+	X(WAL_REC_COMMIT,                2, "COMMIT",             wal_parse_rec_finish) \
+	X(WAL_REC_ROLLBACK,              3, "ROLLBACK",           wal_parse_rec_finish) \
+	X(WAL_REC_RELATION,              4, "RELATION",           wal_parse_rec_relation) \
+	X(WAL_REC_INSERT,                5, "INSERT",             wal_parse_rec_modify) \
+	X(WAL_REC_UPDATE,                6, "UPDATE",             wal_parse_rec_modify) \
+	X(WAL_REC_DELETE,                7, "DELETE",             wal_parse_rec_modify) \
+	X(WAL_REC_O_TABLES_META_LOCK,    8, "META_LOCK",          NULL) \
+	X(WAL_REC_O_TABLES_META_UNLOCK,  9, "META_UNLOCK",        wal_parse_rec_o_tables_meta_unlock) \
+	X(WAL_REC_SAVEPOINT,            10, "SAVEPOINT",          wal_parse_rec_savepoint) \
+	X(WAL_REC_ROLLBACK_TO_SAVEPOINT,11, "RB_TO_SP",           wal_parse_rec_rollback_to_savepoint) \
+	X(WAL_REC_JOINT_COMMIT,         12, "JOINT_COMMIT",       wal_parse_rec_joint_commit) \
+	X(WAL_REC_TRUNCATE,             13, "TRUNCATE",           wal_parse_rec_truncate) \
+	X(WAL_REC_BRIDGE_ERASE,         14, "BRIDGE_ERASE",       wal_parse_rec_bridge_erase) \
+	X(WAL_REC_REINSERT,             15, "REINSERT",           wal_parse_rec_modify) \
+	X(WAL_REC_REPLAY_FEEDBACK,      16, "REPLAY_FEEDBACK",    NULL) \
+	X(WAL_REC_SWITCH_LOGICAL_XID,   17, "SWITCH_LOGICAL_XID", wal_parse_rec_switch_logical_xid) \
+	X(WAL_REC_RELREPLIDENT,         18, "RELREPLIDENT",       wal_parse_rec_relreplident)
+
+/* Parsers */
+extern WalParseResult wal_parse_rec_xid(WalReaderState *r, WalRecord *rec);
+extern WalParseResult wal_parse_rec_finish(WalReaderState *r, WalRecord *rec);
+extern WalParseResult wal_parse_rec_relation(WalReaderState *r, WalRecord *rec);
+extern WalParseResult wal_parse_rec_o_tables_meta_unlock(WalReaderState *r, WalRecord *rec);
+extern WalParseResult wal_parse_rec_savepoint(WalReaderState *r, WalRecord *rec);
+extern WalParseResult wal_parse_rec_rollback_to_savepoint(WalReaderState *r, WalRecord *rec);
+extern WalParseResult wal_parse_rec_joint_commit(WalReaderState *r, WalRecord *rec);
+extern WalParseResult wal_parse_rec_truncate(WalReaderState *r, WalRecord *rec);
+extern WalParseResult wal_parse_rec_bridge_erase(WalReaderState *r, WalRecord *rec);
+extern WalParseResult wal_parse_rec_switch_logical_xid(WalReaderState *r, WalRecord *rec);
+extern WalParseResult wal_parse_rec_relreplident(WalReaderState *r, WalRecord *rec);
+extern WalParseResult wal_parse_rec_modify(WalReaderState *r, WalRecord *rec);
+
+enum OrioleWalRecType
+{
+#define X(sym, val, name, fn) sym = (val),
+	ORIOLE_WAL_RECORDS(X)
+#undef X
+};
 
 /*
  * Value has been fixed at the moment of introducing WAL versioning.
@@ -67,7 +89,7 @@
  */
 #define ORIOLEDB_WAL_VERSION (17)
 
-#define ORIOLEDB_XACT_INFO_WAL_VERSION (17)
+#define ORIOLEDB_CONTAINER_FLAGS_WAL_VERSION (17)
 
 /* Constants for commitInProgressXlogLocation */
 #define OWalTmpCommitPos			(0)
@@ -200,8 +222,20 @@ typedef struct
  * WAL_CONTAINER_HAS_* flags.
  */
 
-#define WAL_CONTAINER_HAS_XACT_INFO		(1U << 0)
-#define WAL_CONTAINER_HAS_ORIGIN_INFO	(1U << 1)
+#define ORIOLE_WAL_FLAGS(X) \
+	X(WAL_CONTAINER_HAS_XACT_INFO,		(1U << 0), "XACT_INFO",		wal_parse_container_xact_info) \
+	X(WAL_CONTAINER_HAS_ORIGIN_INFO,	(1U << 1), "HAS_ORIGIN",	wal_parse_container_origin_info)
+
+/* Flag Parsers */
+extern WalParseResult wal_parse_container_xact_info(WalReaderState *r, WalRecord *rec);
+extern WalParseResult wal_parse_container_origin_info(WalReaderState *r, WalRecord *rec);
+
+enum OrioleWalFlagType
+{
+#define X(sym, val, name, fn)	sym = (val),
+	ORIOLE_WAL_FLAGS(X)
+#undef  X
+};
 
 typedef struct
 {
@@ -222,48 +256,7 @@ typedef struct
 
 extern const char *wal_record_type_to_string(int wal_record);
 
-/* API for parsing WAL-records */
-
-/* Parser for WAL_REC_XID */
-extern Pointer wal_parse_rec_xid(Pointer ptr, OXid *oxid, TransactionId *logicalXid, TransactionId *heapXid, uint16 wal_version);
-
-/* Parser for WAL_REC_COMMIT and WAL_REC_ROLLBACK */
-extern Pointer wal_parse_rec_finish(Pointer ptr, OXid *xmin, CommitSeqNo *csn);
-
-/* Parser for WAL_REC_JOINT_COMMIT */
-extern Pointer wal_parse_rec_joint_commit(Pointer ptr, TransactionId *xid, OXid *xmin, CommitSeqNo *csn);
-
-/* Parser for WAL_REC_RELATION */
-extern Pointer wal_parse_rec_relation(Pointer ptr, uint8 *treeType, ORelOids *oids, OXid *xmin, CommitSeqNo *csn, CommandId *cid, uint32 *version, uint32 *base_version, uint16 wal_version);
-
-/* Parser for WAL_REC_RELREPLIDENT */
-extern Pointer wal_parse_rec_relreplident(Pointer ptr, char *relreplident, Oid *relreplident_ix_oid);
-
-/* Parser for WAL_REC_O_TABLES_META_UNLOCK */
-extern Pointer wal_parse_rec_o_tables_meta_unlock(Pointer ptr, ORelOids *oids, Oid *old_relnode);
-
-/* Parser for WAL_REC_SAVEPOINT */
-extern Pointer wal_parse_rec_savepoint(Pointer ptr, SubTransactionId *parentSubid, TransactionId *logicalXid, TransactionId *parentLogicalXid);
-
-/* Parser for WAL_REC_ROLLBACK_TO_SAVEPOINT */
-extern Pointer wal_parse_rec_rollback_to_savepoint(Pointer ptr, SubTransactionId *parentSubid, OXid *xmin, CommitSeqNo *csn, uint16 wal_version);
-
-/* Parser for WAL_REC_TRUNCATE */
-extern Pointer wal_parse_rec_truncate(Pointer ptr, ORelOids *oids);
-
-/* Parser for WAL_REC_BRIDGE_ERASE */
-extern Pointer wal_parse_rec_bridge_erase(Pointer ptr, ItemPointerData *iptr);
-
-/* Parser for WAL_REC_SWITCH_LOGICAL_XID */
-extern Pointer wal_parse_rec_switch_logical_xid(Pointer ptr, TransactionId *topXid, TransactionId *subXid);
-
-/* Parser for WAL_REC_INSERT, WAL_REC_UPDATE, WAL_REC_DELETE or WAL_REC_REINSERT */
-extern Pointer wal_parse_rec_modify(Pointer ptr, OFixedTuple *tuple1, OFixedTuple *tuple2, OffsetNumber *length1, bool read_two_tuples);
-
 extern void add_rel_wal_record(ORelOids oids, OIndexType type, uint32 version, uint32 base_version);
-
-/* Parser for WAL_CONTAINER_XACT_INFO */
-extern Pointer wal_parse_container_xact_info(Pointer ptr, TimestampTz *xactTime, TransactionId *xid);
 
 extern void add_modify_wal_record(uint8 rec_type, BTreeDescr *desc,
 								  OTuple tuple, OffsetNumber length, char relreplident, uint32 version, uint32 base_version);
@@ -293,10 +286,5 @@ extern void o_wal_reinsert(BTreeDescr *desc, OTuple oldtuple, OTuple newtuple, c
 extern void add_truncate_wal_record(ORelOids oids);
 extern bool get_local_wal_has_material_changes(void);
 extern void set_local_wal_has_material_changes(bool value);
-
-extern Pointer wal_container_read_header(Pointer ptr, uint16 *version,
-										 uint8 *flags);
-extern Pointer wal_parse_container_origin_info(Pointer ptr, RepOriginId *origin_id,
-											   XLogRecPtr *origin_lsn);
 
 #endif							/* __WAL_H__ */
--- a/include/recovery/wal_reader.h
+++ b/include/recovery/wal_reader.h
@@ -1,0 +1,275 @@
+/*-------------------------------------------------------------------------
+ *
+ * wal_reader.h
+ * 		WAL parser declarations for OrioleDB.
+ *
+ * Copyright (c) 2026, Oriole DB Inc.
+ * Copyright (c) 2026, Supabase Inc.
+ *
+ * IDENTIFICATION
+ *	  contrib/orioledb/include/recovery/wal_reader.h
+ *
+ *-------------------------------------------------------------------------
+ */
+#ifndef __WAL_READER_H__
+#define __WAL_READER_H__
+
+typedef unsigned int wal_type_t;
+
+/*
+ * WalRecord instances are transient and reused across iterations.
+ *
+ * Callers must not retain pointers to the record itself.
+ * Any data that must outlive the callback must be copied.
+ */
+typedef struct WalRecord
+{
+	wal_type_t	type;
+
+	uint32		delta;
+	Pointer		value_ptr;
+	uint16		wal_version;
+
+	ORelOids	oids;
+	OXid		oxid;
+	TransactionId logicalXid;
+	TransactionId heapXid;
+	char		relreplident;
+
+	RepOriginId origin_id;
+	XLogRecPtr	origin_lsn;
+
+	union
+	{
+		struct
+		{
+			OXid		xmin;
+			CommitSeqNo csn;
+		}			finish;
+		struct
+		{
+			TransactionId topXid;
+			TransactionId subXid;
+		}			swxid;
+		struct
+		{
+			TransactionId xid;
+			OXid		xmin;
+			CommitSeqNo csn;
+		}			joint_commit;
+		struct
+		{
+			uint8		treeType;
+			OSnapshot	snapshot;
+			uint32		version;
+			uint32		base_version;
+		}			relation;
+		struct
+		{
+			Oid			relreplident_ix_oid;
+		}			relreplident;
+		struct
+		{
+			ORelOids	oids;
+			Oid			oldRelnode;
+		}			unlock;
+		struct
+		{
+			ORelOids	oids;
+		}			truncate;
+		struct
+		{
+			SubTransactionId parentSubid;
+			TransactionId parentLogicalXid;
+		}			savepoint;
+		struct
+		{
+			SubTransactionId parentSubid;
+			OXid		xmin;
+			CommitSeqNo csn;
+		}			rb_to_sp;
+		struct
+		{
+			ItemPointerData iptr;
+		}			bridge_erase;
+
+		struct
+		{
+			OTuple		t1;
+			OffsetNumber len1;
+
+			OTuple		t2;
+			OffsetNumber len2;
+
+			bool		read_two_tuples;
+		}			modify;
+
+		/* Flags */
+		struct
+		{
+			TimestampTz xactTime;
+			TransactionId xid;
+		}			xact_info;
+
+	}			u;
+
+} WalRecord;
+
+/*
+ * WalParseResult
+ *
+ * Status codes returned by the WAL container parser and callbacks.
+ *
+ * WALPARSE_OK
+ *     Success.
+ *
+ * WALPARSE_STOP
+ *     Terminate parser.
+ *
+ * WALPARSE_EOF
+ *     Not enough bytes in the input buffer to parse the requested element.
+ *     This is a "need more data" / framing error depending on the caller.
+ *
+ * WALPARSE_BAD_TYPE
+ *     Unknown record/flag type, missing descriptor, or otherwise
+ *     unparseable tag. Treated as a hard protocol error.
+ *
+ * WALPARSE_BAD_VERSION
+ *     Container version policy rejected by the consumer (typically WAL from
+ *     a newer or unsupported OrioleDB version).
+ *
+ * WALPARSE_INTERNAL
+ *     Internal invariant violation or unexpected condition.
+ */
+typedef enum WalParseResult
+{
+	WALPARSE_OK = 0,
+	WALPARSE_STOP,
+	WALPARSE_EOF,				/* not enough bytes */
+	WALPARSE_BAD_TYPE,
+	WALPARSE_BAD_VERSION,
+	WALPARSE_INTERNAL
+
+} WalParseResult;
+
+struct WalReaderState;
+
+/*
+ * WalCheckVersionFn
+ *
+ * Optional consumer-level version policy check.
+ *
+ * Called after container framing has been validated, but before any
+ * records are delivered. Allows recovery/decoder code to reject WAL
+ * based on semantic compatibility rules.
+ */
+typedef WalParseResult (*WalCheckVersionFn) (const struct WalReaderState *r);
+typedef WalParseResult (*WalOnFlagFn) (void *ctx, const WalRecord *rec);
+typedef WalParseResult (*WalOnRecordFn) (void *ctx, WalRecord *rec);
+
+/*
+ * Cursor advancement invariant:
+ *
+ * r->ptr must only be advanced by:
+ *
+ *   - WR_PARSE / WR_SKIP,
+ *   - record parse routines,
+ *   - container flag parsers.
+ *
+ * Consumers must never modify ptr.
+ */
+typedef struct WalReaderState
+{
+	Pointer		start;
+	Pointer		end;
+	Pointer		ptr;
+	uint16		wal_version;
+	uint8		wal_flags;
+
+	/* Consumer */
+	void	   *ctx;
+	WalCheckVersionFn check_version;
+	WalOnFlagFn on_flag;
+	WalOnRecordFn on_record;
+
+} WalReaderState;
+
+/*
+ * WalParseFn
+ *
+ * Parser routine for a single record type.
+ *
+ * The parser must:
+ *   - read exactly this record's payload from r->ptr,
+ *   - populate rec->u.* fields as needed,
+ *   - leave r->ptr positioned at the next record tag.
+ *
+ * It must not read beyond r->end; use WR_REQUIRE_SIZE / WR_PARSE / WR_SKIP.
+ *
+ * For payload-less records, descriptor->parse is NULL (record is tag-only).
+ */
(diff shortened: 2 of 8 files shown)


src/recovery/wal_reader.c
```diff
@@ -0,0 +1,135 @@
...
+		/*
+		 * Container starts with rec_type of first WAL record (its maximum
+		 * value was under FIRST_ORIOLEDB_WAL_VERSION at the time of
+		 * introducing WAL versioning. Consider this as version 0 and don't
+		 * increase pointer
+		 */
+		wal_version = 0;
+	}
+
+	if (wal_version > ORIOLEDB_WAL_VERSION)
```
akorotkov, line 223: This means we have an unsupported WAL version.  Should we throw a FATAL error unconditionally?

akorotkov:
Looks good for me. Please, go ahead and use it in all the places.

src/orioledb.c
```diff
@@ -234,10 +236,95 @@ PG_FUNCTION_INFO_V1(orioledb_ucm_check);
 PG_FUNCTION_INFO_V1(orioledb_parallel_debug_start);
 PG_FUNCTION_INFO_V1(orioledb_parallel_debug_stop);
 
+#define ORIOLEDB_DEV
```
pashkinelfe: There need a comment about semantic difference of IS_DEV and ORIOLEDB_DEV. Maybe they should be tied to each other

src/recovery/wal_container_iter.c
```diff
@@ -0,0 +1,136 @@
...
+		 * value was under FIRST_ORIOLEDB_WAL_VERSION at the time of
+		 * introducing WAL versioning. Consider this as version 0 and don't
+		 * increase pointer
+		 */
+		wal_version = 0;
+	}
+
+	if (wal_version > ORIOLEDB_WAL_VERSION)
+	{
+		if (allow_logging)
```
pashkinelfe: allow_logging = false prevents elog(FATAL) that is used for detecting wal_version mismatch in tests

src/recovery/wal_reader.c
pashkinelfe: Previously this had different meaning for recovery and logical decoding. Recovery ended with PANIC, logical decoding just skipped the container.

Current PR adds parsing that's duplicate to what is already implemented in orioledb_decode and replay_container, but doesn't replace it. I'm afraid that possible differences between recovery/logical decoding could be missed in a new parsing. I can't verify equivalence of a new routines to old ones.

akorotkov: Hmm.. I'm not sure skipping unsupported WAL containers in orioledb_decode() is a good idea. Should we just do PANIC in all the cases?

src/orioledb.c
akorotkov: Yes, probably just use IS_DEV instead of ORIOLEDB_DEV?

src/recovery/wal_container_iter.c
```diff
@@ -0,0 +1,136 @@
...
+
+	memset(&ev, 0, sizeof(ev));
+
+	ev.relreplident = REPLICA_IDENTITY_DEFAULT;
+
+	st = wr_wal_container_read_header(r, allow_logging);
+	if (st)
+		return st;
+
+	if (r->wal_flags & WAL_CONTAINER_HAS_XACT_INFO)
```
pashkinelfe: WALRecXactInfo should be used in replay_container. https://github.com/orioledb/orioledb/pull/679/changes#diff-764a798f825ba685442692cb1e2b6b41bb5cf3b2befd48aa43644c2ab1aadf16R2983 (don't forget this when replace parsing in replay_container with this)

src/recovery/wal_container_iter.c
```diff
@@ -0,0 +1,135 @@
...
+		/*
+		 * Container starts with rec_type of first WAL record (its maximum
+		 * value was under FIRST_ORIOLEDB_WAL_VERSION at the time of
+		 * introducing WAL versioning. Consider this as version 0 and don't
+		 * increase pointer
+		 */
+		wal_version = 0;
+	}
+
+	if (wal_version > ORIOLEDB_WAL_VERSION)
```
pashkinelfe: skipping unsupported WAL containers breaks logical decoding only. PANIC can make the whole cluster down. I think logical decoding should not do this thing to a whole cluser

src/recovery/wal_reader.c
pashkinelfe: PANIC at parsing when called from logical decoding can shut down the whole db cluster. I think logical decoding should not be allowed to do anything with cluster besides throwing errors

pashkinelfe: Existing processing in orioledb_decode looks correct to me 
image

akorotkov: I thought about possible cases when logical decoding could stop future versions of OrioleDB WAL.  We could downgrade OrioleDB on a cluster that has been shut down cleanly.  Then our recovery wouldn't spot future WAL versions (as it wouldn't process any).  But logical slots, which are probably in the past, could spot them.  So, yes, throwing a ERROR should be correct in this case.

src/recovery/wal_container_iter.c
GremSnoort: Introduced `WalOnFlagFn on_flag` callback in `struct WalConsumer`, now consumer may customize behavior:
- `replay_wal_on_flag` - set `recoveryHeapTransactionId`
- `decode_wal_on_flag` - do nothing

Define flags via `ORIOLE_WAL_FLAGS`

src/recovery/wal_container_iter.c
GremSnoort: `allow_logging` arg in `wal_container_iterate` is necessary to prevent logging from `orioledb_rm_desc` because logging there causes PANIC on ERRORDATA_STACK_SIZE from elog's `get_error_stack_entry`.

Call `wal_container_iterate` from logical decoder `orioledb_decode` -> `true /* allow_logging */`
Call `wal_container_iterate` from recovery `replay_container` -> `true /* allow_logging */`

pashkinelfe: Is there way to know that we had WAL version mismatch or unsupported WAL version during call from orioledb_rm_desc?

include/recovery/wal_event.h
```diff
@@ -0,0 +1,124 @@
...
+ *	  contrib/orioledb/include/recovery/wal_event.h
+ *
+ *-------------------------------------------------------------------------
+ */
+#ifndef __WAL_EVENT_H__
+#define __WAL_EVENT_H__
+
+typedef unsigned int wal_type_t;
+
+typedef struct WalTupleView
```
pashkinelfe: *TupleView naming looks misleading to me. It's OTuple got from WAL, not anything view-like

src/recovery/wal_reader.c
```diff
@@ -0,0 +1,362 @@
...
+	 * Flags describe metadata that applies to the whole container and may
+	 * carry their own payload. These bytes logically belong to the header
+	 * area and must be consumed before we start scanning record tags.
+	 * Otherwise, r->ptr would be misaligned and record parsing would
+	 * desynchronize.
+	 *
+	 * The flag handling also gives the consumer a chance to capture
+	 * header-wide context (e.g. xact-info) before any records are delivered.
+	 */
+	st = wal_container_flags_iterate(r, consumer);
```
pashkinelfe: I see no actual iteration in wal_container_flags_iterate.
wal_container_flag_check() has only one call from the code

I propose to combine wal_container_flags_iterate() and wal_container_flag_check() in one function named as  wal_read_container_flags() or similar

src/recovery/wal.c
```diff
@@ -110,7 +110,50 @@ do { \
 	if (ptr) *ptr = value; \
 } while(0)
 
+WalParseStatus
+wal_flag_parse_container_xact_info(WalReader *r, WalEvent *ev)
+{
+	Assert(r);
+	Assert(ev);
+
+	WR_READ(r, &ev->u.xact_info.xactTime);
```
pashkinelfe: Do we need both WR_READ and PARSE macro?

src/recovery/wal.c
```diff
@@ -110,7 +110,50 @@ do { \
...
+
+	WR_READ(r, &ev->origin_id);
+	WR_READ(r, &ev->origin_lsn);
+
+	return WALPARSE_OK;
+}
+
 /* Parser for WAL_REC_XID */
+WalParseStatus
+wal_parse_xid(WalReader *r, WalEvent *ev)
```
pashkinelfe: Do we need both wal_parse_xid() and wal_parse_rec_xid()?

include/recovery/wal_reader.h
```diff
@@ -0,0 +1,65 @@
...
+#define WR_REQUIRE(r, nbytes) \
+	do { if (wr_remaining((r)) < (size_t)(nbytes)) return WALPARSE_EOF; } while (0)
+
+#define WR_READ(r, out) \
+{ \
+	WR_REQUIRE(r, sizeof(*out)); \
+	memcpy(out, r->ptr, sizeof(*out)); \
+    r->ptr += sizeof(*out); \
+}
+#define WR_PEEK(r, out) \
```
pashkinelfe: One-caller macro. Propose to remove.

include/recovery/wal_reader.h
```diff
@@ -0,0 +1,65 @@
...
+	WALPARSE_OK = 0,
+	WALPARSE_EOF,				/* not enough bytes */
+	WALPARSE_BAD_TYPE,
+	WALPARSE_BAD_VERSION,
+	WALPARSE_INTERNAL
+
+} WalParseStatus;
+
+static inline size_t
+wr_remaining(const WalReader *r)
```
pashkinelfe: One-caller fn. Propose to remove.

include/recovery/wal_reader.h
```diff
@@ -0,0 +1,65 @@
...
+{
+	Pointer		start;
+	Pointer		end;
+
+	Pointer		ptr;
+
+	uint16		wal_version;
+	uint8		wal_flags;
+
+} WalReader;
```
pashkinelfe: Semantics not clear (WalReaderState?)

include/recovery/wal_reader.h
```diff
@@ -0,0 +1,65 @@
...
+
+typedef enum WalParseStatus
```
pashkinelfe: WalParseResult?

include/recovery/wal_reader.h
```diff
@@ -0,0 +1,65 @@
...
+
+} WalParseStatus;
+
+static inline size_t
+wr_remaining(const WalReader *r)
+{
+	return (size_t) (r->end - r->ptr);
+}
+
+#define WR_REQUIRE(r, nbytes) \
```
pashkinelfe: Semantics not clear. WR_CHECK_EOF, WR_CHECK_SIZE ?

include/recovery/wal_dispatch.h
```diff
@@ -0,0 +1,36 @@
+/*-------------------------------------------------------------------------
+ *
+ * wal_dispatch.h
```
pashkinelfe: Semantics not clear. I'd propose to join wal_container_iter.h + wal_dispatch.h + wal_event.h + wal_reader.h into one (wal_reader.h)

src/recovery/wal_container_iter.c
```diff
@@ -0,0 +1,362 @@
...
+		}
+		else
+			st = WALPARSE_BAD_TYPE;
+	}
+
+	return st;
+}
+
+static WalParseStatus
+wal_container_flags_iterate(WalReader *r, WalConsumer *consumer)
```
pashkinelfe: I'd propose to compare container flags to a fixed list of defines, without using iterator. Now we have only a couple of flags

src/recovery/wal_container_iter.c
```diff
@@ -0,0 +1,362 @@
...
+#undef X
+
+		return st;
+}
+
+WalParseStatus
+wal_container_iterate(WalReader *r, WalConsumer *consumer, bool allow_logging)
+{
+	WalParseStatus st;
+	WalEvent	ev;
```
pashkinelfe: Event semantics is unclear. I'd propose to stick to existing container-record semantics in naming

include/recovery/wal_reader.h
```diff
@@ -0,0 +1,65 @@
...
+	do { if (wr_remaining((r)) < (size_t)(nbytes)) return WALPARSE_EOF; } while (0)
+
+#define WR_READ(r, out) \
```
pashkinelfe: WR_READ-WR_PEEK-WR_SKIP semantically are little bit unclear 
READ and PEEK are different by pointer increase, and that's not clear from the names
READ semantics includes pointer increase

Semantics PARSE = READ (memcpy) + PROCEED (add ptr) or similar looks better to me

src/recovery/wal_container_iter.c
```diff
@@ -0,0 +1,362 @@
...
+
+	ORIOLE_WAL_FLAGS(X)
+
+#undef X
+
+		return st;
+}
+
+WalParseStatus
+wal_container_iterate(WalReader *r, WalConsumer *consumer, bool allow_logging)
```
pashkinelfe: Iteration semantics is not clear. I'd propose rename to parse_wal_container()

include/recovery/wal_event.h
GremSnoort: The naming is intentional, but I agree it deserves better documentation.

`WalTupleView` is not a materialized tuple but a lightweight, non-owning reference into the WAL reader buffer. The parser does zero-copy extraction and simply points to the tuple payload inside the current WAL chunk. This means the memory is transient and only valid while the reader buffer remains unchanged.

The term `View` is used here to emphasize that this structure **does not own the data** and should not be treated as a stable tuple. Consumers that need to retain the tuple **must copy** it.

I’ve added a clarifying comment describing the lifetime and ownership semantics to make this explicit, commit a52784a1d79ffb2940f0733c8c78ed9a1ab1f3a8.

src/recovery/wal.c
GremSnoort: Remove unused macro `PARSE` in c927970a72b1fe3d8821c18487dc4a37f18cdad1.

src/recovery/wal.c
GremSnoort: Remove unused methods in a52784a1d79ffb2940f0733c8c78ed9a1ab1f3a8 and c927970a72b1fe3d8821c18487dc4a37f18cdad1.

src/recovery/wal_reader.c
GremSnoort: Yes, `wal_container_flags_iterate()` doesn't look like an iteration: it's an X-macro expansion, so the "loop" is unrolled at compile time rather than expressed as a runtime `for`.

I propose to keep the two-level structure (`*_flags_iterate()` + `*_flag_check()`) for separation of concerns:

(1) **Single-flag semantics live in one place**
`wal_container_flag_check()` encapsulates the contract for one flag:
- check `r->wal_flags` bit,
- resolve descriptor (`wal_flag_get_desc()`),
- parse optional flag payload via descriptor parser,
- call `consumer->on_flag()` if present,
- return the correct `WalParseStatus`.

This is a tight unit with clear invariants and is easy to reason about/test independently (including error paths like missing desc/parse).

(2) **The set of supported flags is a different concern**
`wal_container_flags_iterate()` is responsible for protocol-level ordering: consume all recognized container flags (header prefix) before record tags. It is intentionally generated from `ORIOLE_WAL_FLAGS(X)` so that when we add a new container flag, there's only one authoritative list to update.

(3) **Keeps future extensibility open**
The macro `ORIOLE_WAL_FLAGS(X)` currently expands into a sequence of calls to `wal_container_flag_check()`, one per supported flag. Keeping the single-flag logic separate lets us change or extend flag-handling behavior without coupling it to the macro-generated enumeration.
So even if now `wal_container_flag_check()` is only called from `wal_container_flags_iterate()`, the split is intentional: one function means "process one flag", the other is - "process all known flags".

I think the current split makes the protocol handling cleaner and reduces the chance of mistakes when we add more flags.

I agree the naming could be improved to match what the code actually does. I propose to rename for clarity, e.g.:
- `wal_container_flags_iterate()`   ->   `wal_container_read_flags()` / `wal_read_container_flags()`
- `wal_container_flag_check()`   ->   `wal_container_read_flag()` / `wal_read_container_flag()`

include/recovery/wal_event.h
pashkinelfe: IMO it's the same concept as OTuple that references actual data but may not "own" it. I'd propose to stick to existing semantics of OTuple and not creating extra terms around that to highlight the "referencing" nature of it.

src/recovery/wal_reader.c
pashkinelfe: I'd disagree in that future flags compatibility justifies this big complication. To me doing easy things in the light and easily understandable way it huge benefit at the moment. Any complications for the future come with a burden for other contributors to sort them out, when they want to make some changes in the present time.

include/recovery/wal_event.h
GremSnoort: Ok, is it good to rename `WalTupleView`  -> `OWalTuple`?

pashkinelfe: Current WalTupleView is actually OTuple+len. Maybe we can go out without additional abstraction that combines these two things. 

More generally, I think that grouping some things into separate abstraction/context is justified if the number of objects is 5-10. For just 2-4 objects it looks easier to work with them explicitly as they are without grouping into additional abstraction/context. The main idea that grouping variables into context may look convenient and beautiful in some view, but it complicates understanding of a code.

src/recovery/wal_reader.c
GremSnoort: The split here is also about protocol safety and avoiding a class of bugs that are otherwise very easy to introduce.

Container flags are part of the binary WAL protocol. If a new flag is added but not parsed, the reader will become desynchronized because it will not skip the flag payload before starting to read record tags. This is a silent failure mode that can corrupt the entire parse stream.

The X-macro gives us a single authoritative list of container flags, ensuring that:
- every declared flag is **automatically processed**,
- adding a flag **cannot be forgotten** in the parser,
- protocol changes remain **mechanically enforced** rather than relying on copy-paste.

Without this pattern, adding a new flag requires updating multiple locations manually, which increases the probability of protocol-breaking mistakes.

So while the current structure may look slightly more abstract, it actually reduces cognitive load for future changes by making the correct behavior automatic.

The solution without X-macro involves manually writing the flag parsing for each flag, which looks like duplicate code and copy-paste and can also be easily forgotten.

src/recovery/wal_container_iter.c
GremSnoort: Updated in b1c1d27b44609dee177c28e8dab4ed3b2629acc2.

For no-logging case in `orioledb_rm_desc`, return `WALPARSE_BAD_VERSION` in fatal case of version mismatch.
This leads to termination of parsing after `wr_wal_container_read_header`, and subsequent `appendStringInfo(buf, " [PARSE ERROR %d]", (int) st);`

src/recovery/wal_container_iter.c
GremSnoort: Renamed in 907e222abe38af3997e5248222c5a52e9fad688c.

include/recovery/wal_reader.h
GremSnoort: Renamed in cf04aafe3d4c2dd33c92cc0f99f72f4088605f90.

include/recovery/wal_reader.h
GremSnoort: Renamed in 623c5758ed880e58e6a8fa489d3210e784f8739b.

include/recovery/wal_reader.h
GremSnoort: Renamed `WR_READ` -> `WR_PARSE` in ede4eab4747439e20039628162fcade905d794b1.

include/recovery/wal_reader.h
GremSnoort: Renamed `WR_REQUIRE` -> `WR_REQUIRE_SIZE` in ba702ffd527016260a8e4ea17f4b9e21838da743.

include/recovery/wal_reader.h
GremSnoort: Removed in ba702ffd527016260a8e4ea17f4b9e21838da743.

include/recovery/wal_reader.h
GremSnoort: Removed in f8ea1c2b46c3810900597da9e86116cc5d73a70b.

include/recovery/wal_event.h
GremSnoort: @pashkinelfe , new update in 32e34eaf2ed75dc906c2982655b921d56db5d3d4.
Struct `WalTupleView` removed permanently.

src/recovery/wal_container_iter.c
GremSnoort: Renamed `WalEvent` -> `WalRecord` in f141d612dc7c4f28bbcaa44c9797e5f6715beb84.

include/recovery/wal_dispatch.h
GremSnoort: Files merged in 7bd4f600d2c6ab2210906909b3d667c6c1896dd2.

src/orioledb.c
GremSnoort: Done in 7c0f23f790de93b471f02b34d1c6fba722421d7e.

include/recovery/wal.h
```diff
@@ -200,8 +225,20 @@ typedef struct
...
 
-#define WAL_CONTAINER_HAS_XACT_INFO		(1U << 0)
-#define WAL_CONTAINER_HAS_ORIGIN_INFO	(1U << 1)
+#define ORIOLE_WAL_FLAGS(X) \
+	X(WAL_CONTAINER_HAS_XACT_INFO,		(1U << 0), "XACT_INFO",		wal_flag_parse_container_xact_info) \
+	X(WAL_CONTAINER_HAS_ORIGIN_INFO,	(1U << 1), "HAS_ORIGIN",	wal_flag_parse_container_has_origin)
+
+/* Flag Parsers */
+extern WalParseResult wal_flag_parse_container_xact_info(WalReaderState *r, WalRecord *rec);
+extern WalParseResult wal_flag_parse_container_has_origin(WalReaderState *r, WalRecord *rec);
```
pashkinelfe: I'd propose to stick to previous naming wal_parse_container_origin_info and wal_parse_container_xact_info

GremSnoort: Done in 89a2cebc9f291cc4a009394379080f2d47d2dcbc.

include/recovery/wal.h
```diff
@@ -14,41 +14,66 @@
...
+	X(WAL_REC_REINSERT,             15, "REINSERT",           wal_parse_modify) \
+	X(WAL_REC_REPLAY_FEEDBACK,      16, "REPLAY_FEEDBACK",    NULL) \
+	X(WAL_REC_SWITCH_LOGICAL_XID,   17, "SWITCH_LOGICAL_XID", wal_parse_switch_logical_xid) \
+	X(WAL_REC_RELREPLIDENT,         18, "RELREPLIDENT",       wal_parse_relreplident)
+
+#define ORIOLE_WAL_RECORD_IS_MODIFY(rec) \
+	(rec == WAL_REC_INSERT || rec == WAL_REC_UPDATE || rec == WAL_REC_DELETE || rec == WAL_REC_REINSERT)
+
+/* Parsers */
+extern WalParseResult wal_parse_xid(WalReaderState *r, WalRecord *rec);
```
pashkinelfe: I'd propose to stick to previous naming wal_parse_rec_* / wal_parse_container_*

include/recovery/wal.h
```diff
@@ -14,41 +14,66 @@
...
+	X(WAL_REC_ROLLBACK_TO_SAVEPOINT,11, "RB_TO_SP",           wal_parse_rollback_to_savepoint) \
+	X(WAL_REC_JOINT_COMMIT,         12, "JOINT_COMMIT",       wal_parse_joint_commit) \
+	X(WAL_REC_TRUNCATE,             13, "TRUNCATE",           wal_parse_truncate) \
+	X(WAL_REC_BRIDGE_ERASE,         14, "BRIDGE_ERASE",       wal_parse_bridge_erase) \
+	X(WAL_REC_REINSERT,             15, "REINSERT",           wal_parse_modify) \
+	X(WAL_REC_REPLAY_FEEDBACK,      16, "REPLAY_FEEDBACK",    NULL) \
+	X(WAL_REC_SWITCH_LOGICAL_XID,   17, "SWITCH_LOGICAL_XID", wal_parse_switch_logical_xid) \
+	X(WAL_REC_RELREPLIDENT,         18, "RELREPLIDENT",       wal_parse_relreplident)
+
+#define ORIOLE_WAL_RECORD_IS_MODIFY(rec) \
```
pashkinelfe: One-caller macro, I propose to join it with caller

include/recovery/wal.h
GremSnoort: Done in 445f85014dbf9d5c1ab556bf4de2d45593261b7e. The old names have been retained.

include/recovery/wal.h
GremSnoort: Done in 67c3f975ab46d0bd5a7dfc81e4287a103503545c.

src/recovery/logical.c
```diff
@@ -580,702 +581,744 @@ remove_skipped_dml(HTAB **cache, uint64 oxid)
...
+	OTableDescr *descr;
+	OIndexDescr *indexDescr;
+	OIndexType	ix_type;
+	TupleDescData *o_toast_tupDesc;
+	TupleDescData *heap_toast_tupDesc;
+	bool		has_origin;
+} DecodeWalDescCtx;
+
+static WalParseResult
+decode_wal_on_flag(void *vctx, const WalRecord *rec)
```
pashkinelfe: Propose to rename to decode_wal_containter_info This is to fit the semantic that on some flag in WAL container we parse additional per-container info field

src/recovery/logical.c
```diff
@@ -580,702 +581,744 @@ remove_skipped_dml(HTAB **cache, uint64 oxid)
...
+	}
 
-		rec_type = *ptr;
-		ptr++;
+	return WALPARSE_OK;
+}
 
-		rec_type_str = wal_record_type_to_string(rec_type);
+static WalParseResult
+decode_wal_on_record(void *vctx, WalRecord *rec)
```
pashkinelfe: Propose to rename to decode_wal_record

src/recovery/logical.c
GremSnoort: Renamed in 2559622b7fa4d81570ca132bdec45096d9398a65.
:question: Perhaps it would be worth renaming `replay_wal_on_flag` to `replay_wal_containter_info` as well?

src/recovery/wal_reader.c
```diff
@@ -0,0 +1,597 @@
...
+			elog(FATAL, "WAL container version %u is older than current %u. Intentionally fail tests", wal_version, ORIOLEDB_WAL_VERSION);
+
+		return WALPARSE_BAD_VERSION;
+#else
+		if (allow_logging)
+			elog(LOG, "WAL container version %u is older than current %u. Applying with conversion.", wal_version, ORIOLEDB_WAL_VERSION);
+#endif
+	}
+
+	if (wal_version >= ORIOLEDB_XACT_INFO_WAL_VERSION)
```
pashkinelfe: Maybe it's worth to rename ORIOLEDB_XACT_INFO_WAL_VERSION -> ORIOLEDB_CONTAINER_FLAGS_WAL_VERSION to fit semantics that at this version flags byte was added to WAL container? (It was done to accommodate XACT_INFO container record and was named because of it. Now we can move name to its semantical meaning)

src/recovery/logical.c
GremSnoort: Renamed in 2559622b7fa4d81570ca132bdec45096d9398a65.
❓ Perhaps it would be worth renaming `replay_wal_on_record` to `replay_wal_record` ? And `wal_desc_on_record` -> `wal_desc_record` ?

src/recovery/wal_reader.c
```diff
@@ -0,0 +1,597 @@
...
+ *     consumer via r->on_flag(). The callback may capture header-wide context
+ *     (e.g. PG xid/origin) needed for interpreting subsequent records, but must
+ *     not attempt to move r->ptr.
+ *
+ * Note: flags with payload are protocol-critical. Forgetting to consume a flag
+ * payload would desynchronize the stream and cause the next record tag to be
+ * misread.
+ */
+static inline WalParseResult
+wal_container_parse_flag(WalReaderState *r, wal_type_t type)
```
pashkinelfe, line 300: One-caller function. Propose to combine with caller

src/recovery/wal_reader.c
```diff
@@ -0,0 +1,597 @@
...
+	 * Read and validate container header framing.
+	 *
+	 * The header establishes wal_version, wal_flags and positions r->ptr at
+	 * the first byte after the header area.
+	 *
+	 * From this point on, parse_wal_container() is the sole authority that
+	 * drives the reader forward: each iteration consumes a record tag byte
+	 * and (optionally) a record payload according to the descriptor.
+	 */
+	st = wr_wal_container_read_header(r, allow_logging);
```
pashkinelfe: Propose to rename to wal_container_read_header

src/recovery/wal_reader.c
GremSnoort: Renamed in 0c161c675a6d8317f7cf24ce43d341b39d690e40.

src/recovery/wal_reader.c
GremSnoort: `wal_container_parse_flag` is used within an X-macro (in `wal_container_parse_flags`), so its contents are moved to a separate function to simplify the macro code.

src/recovery/logical.c
pashkinelfe: Yes, please proceed with this

src/recovery/logical.c
pashkinelfe: Yes, please proceed with this

GremSnoort: Renamed in 4fdf1659014f68ee1cc00a3ecf71261cdaeebf44.

src/recovery/logical.c
GremSnoort: Renamed in 4fdf1659014f68ee1cc00a3ecf71261cdaeebf44.

src/recovery/wal_reader.c
GremSnoort: Review please 00a5dffc51c325c24f1216b4883187563ceab95a

Old version of `wal_container_read_header` has been used in `orioledb_recovery_stops_before_hook`, so I replace this occurrence with WAL parser with early termination after flags / on record.

src/recovery/wal_reader.c
```diff
@@ -0,0 +1,596 @@
...
+	 *
+	 * The header establishes wal_version, wal_flags and positions r->ptr at
+	 * the first byte after the header area.
+	 *
+	 * From this point on, parse_wal_container() is the sole authority that
+	 * drives the reader forward: each iteration consumes a record tag byte
+	 * and (optionally) a record payload according to the descriptor.
+	 */
+	st = wal_container_read_header(r, allow_logging);
+	if (st)
```
pashkinelfe: if(st != WALPARSE_OK)

src/recovery/wal_reader.c
```diff
@@ -0,0 +1,596 @@
...
+
+	/*
+	 * Version compatibility check.
+	 *
+	 * We keep parse_wal_container() largely version-agnostic and delegate
+	 * version policy to the consumer (recovery / logical decoding / etc.).
+	 * Parsing WAL from a newer OrioleDB version can be unsafe even if the
+	 * container framing is understood, because record encodings may differ.
+	 */
+	if (r->check_version)
```
pashkinelfe, line 441: We check version in wal_container_read_header, and then check it in check_version(). Duplicate check complicates understanding. At least need short and precise comment what  wal_container_read_header checks and what  check_version (see proposed comments above)

src/recovery/wal_reader.c
```diff
@@ -0,0 +1,596 @@
...
+	 * From this point on, parse_wal_container() is the sole authority that
+	 * drives the reader forward: each iteration consumes a record tag byte
+	 * and (optionally) a record payload according to the descriptor.
+	 */
+	st = wal_container_read_header(r, allow_logging);
+	if (st)
+		return st;
+
+	/*
+	 * Version compatibility check.
```
pashkinelfe: Second?

src/recovery/wal_reader.c
```diff
@@ -0,0 +1,596 @@
...
+	Assert(r->on_record);		/* consumer must handle every record */
+
+	memset(&rec, 0, sizeof(rec));
+
+	rec.relreplident = REPLICA_IDENTITY_DEFAULT;
+	rec.origin_id = InvalidRepOriginId;
+	rec.origin_lsn = InvalidXLogRecPtr;
+
+	/*
+	 * Read and validate container header framing.
```
pashkinelfe: And first WAL version check?

src/recovery/wal_reader.c
```diff
@@ -0,0 +1,596 @@
...
+	 *
+	 * From this point on, parse_wal_container() is the sole authority that
+	 * drives the reader forward: each iteration consumes a record tag byte
+	 * and (optionally) a record payload according to the descriptor.
+	 */
+	st = wal_container_read_header(r, allow_logging);
+	if (st)
+		return st;
+
+	/*
```
pashkinelfe, line 427: "Consumer's (recovery / logical decoding) version compatibility check. Relies on first (common) check that doesn't pass versions from the future that can't be parsed."

src/recovery/wal_reader.c
```diff
@@ -0,0 +1,596 @@
...
+	 * Read and validate container header framing.
+	 *
+	 * The header establishes wal_version, wal_flags and positions r->ptr at
+	 * the first byte after the header area.
+	 *
+	 * From this point on, parse_wal_container() is the sole authority that
+	 * drives the reader forward: each iteration consumes a record tag byte
+	 * and (optionally) a record payload according to the descriptor.
+	 */
```
pashkinelfe:
```suggestion
	 * First (common) check of wal_version ensures that WAL container from 
	 * the future that can't be parsed is rejected 
	 */
```

src/recovery/wal_reader.c
```diff
@@ -0,0 +1,596 @@
...
+		return st;
+
+	/*
+	 * Version compatibility check.
+	 *
+	 * We keep parse_wal_container() largely version-agnostic and delegate
+	 * version policy to the consumer (recovery / logical decoding / etc.).
+	 * Parsing WAL from a newer OrioleDB version can be unsafe even if the
+	 * container framing is understood, because record encodings may differ.
+	 */
```
pashkinelfe:
```suggestion
	 * Consumer's (recovery / logical decoding) version compatibility check. Relies on first (common) check that doesn't pass versions from the future that can't be parsed.
	 */
```

src/recovery/wal_reader.c
GremSnoort: Done in d5bd2fed13850ce0462a5cd95315d4930096ffae.

src/recovery/wal_reader.c
GremSnoort: Yes, the comments have been updated in c664c0634b5a375ec5b774911993aa1ef4d73a65, review please

src/recovery/wal_reader.c
GremSnoort: Updated in c664c0634b5a375ec5b774911993aa1ef4d73a65.

src/recovery/wal_reader.c
GremSnoort: I propose update c664c0634b5a375ec5b774911993aa1ef4d73a65

src/recovery/wal_reader.c
GremSnoort: Updated in 77899851cd7d37fabc0d7047a982d480582fb427.

PR #701, opened 2026-01-29
Remove useless rootInfo assignment

e-ivkov:
It is already assigned a couple of lines earlier


pashkinelfe:
need some investigation on two identical crashes with this commit (mentioned in https://github.com/orioledb/orioledb/pull/703)

PR #711, opened 2026-02-06
Introduce Local page pool

### LINKED ISSUE #685: Temporary Tables Improvement

e-ivkov:
Introduces local page pool for temporary tables. Temporary tables are now always created on local page pool instead of shared. Eviction is supported and works similar to vanilla postgres.

Closes https://github.com/orioledb/orioledb/issues/685

Previous PR: https://github.com/orioledb/orioledb/pull/686

Generally the implementation follows the plan in the issue, but instead of small allocator it uses postgres' slab.

Performance improvements are significant: better latency and TPS after 8 connections and the difference only increases with the number of connections.
image

### DIFF
--- a/include/btree/btree.h
+++ b/include/btree/btree.h
@@ -176,7 +176,7 @@ struct BTreeDescr
 	OSmgr		smgr;
 	ORelOids	oids;
 	OIndexType	type;
-	OPagePool  *ppool;
+	PagePool   *ppool;
 	OCompress	compress;
 	uint8		fillfactor;
 	UndoLogType undoType;
--- a/include/btree/merge.h
+++ b/include/btree/merge.h
@@ -22,7 +22,7 @@ extern bool btree_try_merge_pages(BTreeDescr *desc,
 								  OFixedKey *parent_hikey,
 								  bool *merge_parent,
 								  OInMemoryBlkno left_blkno,
-								  BTreePageItemLocator right_loc,
+								  BTreePageItemLocator *right_loc,
 								  OInMemoryBlkno right_blkno,
 								  bool checkpoint);
 extern bool btree_try_merge_and_unlock(BTreeDescr *desc, OInMemoryBlkno blkno,
--- a/include/orioledb.h
+++ b/include/orioledb.h
@@ -449,6 +449,7 @@ extern XLogRecPtr replay_until_lsn;
 	(AssertMacro(MYPROCNUMBER >= 0 && \
 				 MYPROCNUMBER < max_procs), \
 	 &oProcData[MYPROCNUMBER])
+#define O_PAGE_IS_LOCAL(blkno) ((blkno) < 0)
 #define O_GET_IN_MEMORY_PAGE(blkno) \
 	(AssertMacro(OInMemoryBlknoIsValid(blkno)), \
 	 (Page)(o_shared_buffers + (((uint64) (blkno)) * ((uint64) ORIOLEDB_BLCKSZ))))
@@ -494,13 +495,13 @@ typedef enum OPagePoolType
 } OPagePoolType;
 #define OPagePoolTypesCount 3
 
-typedef struct OPagePool OPagePool;
+typedef struct PagePool PagePool;
 struct BTreeDescr;
 
 extern void o_verify_dir_exists_or_create(char *dirname, bool *created, bool *found);
 extern uint64 orioledb_device_alloc(struct BTreeDescr *descr, uint32 size);
-extern OPagePool *get_ppool(OPagePoolType type);
-extern OPagePool *get_ppool_by_blkno(OInMemoryBlkno blkno);
+extern PagePool *get_ppool(OPagePoolType type);
+extern PagePool *get_ppool_by_blkno(OInMemoryBlkno blkno);
 extern OInMemoryBlkno get_dirty_pages_count_sum(void);
 extern void jsonb_push_key(JsonbParseState **state, char *key);
 extern void jsonb_push_null_key(JsonbParseState **state, char *key);
--- a/include/orioledb.h
+++ b/include/orioledb.h
@@ -129,7 +129,10 @@
 
 #define GetMaxBackends() MaxBackends
 
-/* Number of orioledb page */
+/*
+ * Number of orioledb page.
+ * If high bit is set, it means the page is in the local memory.
+ */
 typedef uint32 OInMemoryBlkno;
 #define OInvalidInMemoryBlkno		((OInMemoryBlkno) 0xFFFFFFFF)
 #define OInMemoryBlknoIsValid(blockNumber) \
@@ -396,6 +399,7 @@ typedef struct
 /* orioledb.c */
 extern Size orioledb_buffers_size;
 extern Size orioledb_buffers_count;
+extern Size orioledb_temp_buffers_count;
 extern Size undo_circular_buffer_size;
 extern uint32 undo_buffers_count;
 extern Size xid_circular_buffer_size;
@@ -407,7 +411,9 @@ extern uint32 rewind_buffers_count;
 extern Pointer o_shared_buffers;
 extern ODBProcData *oProcData;
 extern int	max_procs;
+extern Page *local_ppool_pages;
 extern OrioleDBPageDesc *page_descs;
+extern OrioleDBPageDesc *local_ppool_page_descs;
 extern bool remove_old_checkpoint_files;
 extern bool skip_unmodified_trees;
 extern bool debug_disable_bgwriter;
@@ -449,12 +455,17 @@ extern XLogRecPtr replay_until_lsn;
 	(AssertMacro(MYPROCNUMBER >= 0 && \
 				 MYPROCNUMBER < max_procs), \
 	 &oProcData[MYPROCNUMBER])
-#define O_PAGE_IS_LOCAL(blkno) ((blkno) < 0)
+/* Needed to get blkno without high bit that defines if page is local or not */
+#define O_BLKNO_MASK ((OInMemoryBlkno) 0x7FFFFFFF)
+#define O_PAGE_IS_LOCAL(blkno) ((blkno) >> 31 != 0)
 #define O_GET_IN_MEMORY_PAGE(blkno) \
 	(AssertMacro(OInMemoryBlknoIsValid(blkno)), \
-	 (Page)(o_shared_buffers + (((uint64) (blkno)) * ((uint64) ORIOLEDB_BLCKSZ))))
+	 (O_PAGE_IS_LOCAL(blkno) ? local_ppool_pages[(blkno) & O_BLKNO_MASK] : \
+	  (Page)(o_shared_buffers + (((uint64) ((blkno) & O_BLKNO_MASK)) * ((uint64) ORIOLEDB_BLCKSZ)))))
 #define O_GET_IN_MEMORY_PAGEDESC(blkno) \
-	(AssertMacro(OInMemoryBlknoIsValid(blkno)), page_descs + (blkno))
+	(AssertMacro(OInMemoryBlknoIsValid(blkno)), \
+     (O_PAGE_IS_LOCAL(blkno) ? local_ppool_page_descs + ((blkno) & O_BLKNO_MASK) : \
+      page_descs + ((blkno) & O_BLKNO_MASK)))
 #define O_GET_IN_MEMORY_PAGE_CHANGE_COUNT(blkno) \
 	(O_PAGE_GET_CHANGE_COUNT(O_GET_IN_MEMORY_PAGE(blkno)))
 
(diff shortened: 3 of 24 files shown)


e-ivkov:
Merged here - https://github.com/orioledb/orioledb/commit/06c95b5b1ca785e387e6b60bf8cfe08c96c0af10

PR #713, opened 2026-02-09
Add origin tracking

pgnickb:
- Added origin info to the header of WAL container if the change was done within the Apply Worker
- Modified the orioledb_decode to parse and filter based on the origin filtering of the logical decoding context

TODO: In the vanilla postgres the origin info is stored on a per-row basis which makes sense in cascading replication contexts and more complicated logical replication setups. Our current implementation packs the origin info into the container header which doesn't support that level of granularity but is consistent with the existing features in native postgres logical replication. We should remember to add per-row origin info when more advanced filtering is implemented in postgres.

### DIFF
--- a/include/recovery/wal.h
+++ b/include/recovery/wal.h
@@ -200,13 +200,21 @@ typedef struct
  * WAL_CONTAINER_HAS_* flags.
  */
 
-#define WAL_CONTAINER_HAS_XACT_INFO	(1U << 0)
+#define WAL_CONTAINER_HAS_XACT_INFO		(1U << 0)
+#define WAL_CONTAINER_HAS_ORIGIN_INFO	(1U << 1)
 
 typedef struct
 {
 	uint8		xactTime[sizeof(TimestampTz)];
 	uint8		xid[sizeof(TransactionId)];
 } WALRecXactInfo;
+
+typedef struct
+{
+	uint8		origin_id[sizeof(RepOriginId)];
+	uint8		origin_lsn[sizeof(XLogRecPtr)];
+} WALRecOriginInfo;
+
 
 #define LOCAL_WAL_BUFFER_SIZE	(8192)
 #define ORIOLEDB_WAL_PREFIX	"o_wal"
@@ -288,5 +296,7 @@ extern void set_local_wal_has_material_changes(bool value);
 
 extern Pointer wal_container_read_header(Pointer ptr, uint16 *version,
 										 uint8 *flags);
+extern Pointer wal_parse_container_origin_info(Pointer ptr, RepOriginId *origin_id,
+											   XLogRecPtr *origin_lsn);
 
 #endif							/* __WAL_H__ */
--- a/src/recovery/logical.c
+++ b/src/recovery/logical.c
@@ -32,6 +32,8 @@
 #include "access/detoast.h"
 #include "tuple/toast.h"
 
+static inline bool FilterByOrigin(LogicalDecodingContext *ctx, RepOriginId origin_id);
+
 /*
  * Copy identity attributes from srcSlot to dstSlot.
  */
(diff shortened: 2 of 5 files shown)


src/recovery/logical.c
```diff
@@ -1179,6 +1193,10 @@ orioledb_decode(LogicalDecodingContext *ctx, XLogRecordBuffer *buf)
 			ReorderBufferTXN *txn;
 			bool		read_two_tuples;
 
+			/* If the origin is defined and filtering is enabled, Skip */
```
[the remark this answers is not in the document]
pgnickb: This is done for consistency reasons. Grepping for `Skip` would net all cases where we skip the record

include/recovery/wal.h
```diff
@@ -200,14 +200,22 @@ typedef struct
...
 {
 	uint8		xactTime[sizeof(TimestampTz)];
 	uint8		xid[sizeof(TransactionId)];
 } WALRecXactInfo;
 
+typedef struct
+{
+	uint8		origin_id[sizeof(RepOriginId)];
+	uint8		origin_lsn[sizeof(XLogRecPtr)];
+} WALRecOriginInfo;
```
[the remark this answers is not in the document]
pgnickb: This is not how the wal.h is maintained.

src/recovery/wal.c
```diff
@@ -1252,3 +1267,13 @@ wal_parse_rec_modify(Pointer ptr, OFixedTuple *tuple1, OFixedTuple *tuple2, Offs
 
 	return ptr;
 }
+
+Pointer
+wal_parse_origin_info(Pointer ptr, RepOriginId *origin_id, XLogRecPtr *origin_lsn)
```
pashkinelfe: I'd propose renaming to wal_parse_container_origin_info to stick to existing naming i.e.
we have wal_parse_rec_* for parsing wal records and wal_parse_container_* for parsing per-container header info.

pgnickb: Good point. I missed this somehow. Should be fixed now.

src/recovery/wal.c
```diff
@@ -982,6 +983,9 @@ log_logical_wal_container(Pointer ptr, int length, bool withXactTime)
 	uint16		wal_version = ORIOLEDB_WAL_VERSION;
 	uint8		flags = 0;
 	WALRecXactInfo rec;
+	WALRecOriginInfo origin;
+	bool		hasOrigin = replorigin_session_origin != InvalidRepOriginId
+		&& replorigin_session_origin_lsn != InvalidXLogRecPtr;
```
za-arthur: I think we should insert `WALRecOriginInfo` independently of `replorigin_session_origin_lsn`'s value, similar to Postgres' behavior:
https://github.com/orioledb/postgres/blob/b8c86807a8cf85eada60f7e488367532042860c3/src/backend/access/transam/xact.c#L5917-L5923
```suggestion
	bool		hasOrigin = replorigin_session_origin != InvalidRepOriginId;
```

src/recovery/logical.c
```diff
@@ -656,6 +660,13 @@ orioledb_decode(LogicalDecodingContext *ctx, XLogRecordBuffer *buf)
 		ptr = wal_parse_container_xact_info(ptr, NULL, NULL);
 	}
 
+	if (wal_flags & WAL_CONTAINER_HAS_ORIGIN_INFO)
+	{
+		ptr = wal_parse_container_origin_info(ptr, &origin_id, &origin_lsn);
+		has_origin = origin_id != InvalidRepOriginId
+					 && origin_lsn != InvalidXLogRecPtr;
```
za-arthur: Similar to [the comment](https://github.com/orioledb/orioledb/pull/713/changes#r2783291550), I think we shouldn't depend on `origin_lsn`'s value. It also isn't used by `FilterByOrigin` anyway.

PR #730, opened 2026-02-24
Implementation of concurrent creation algorithm for secondary indexes.

Serge-sudo:
The procedure follows a two-pass approach similar to PostgreSQL's concurrent index build, adapted for OrioleDB architecture:

1. Initial build using a consistent snapshot
   - Scan the table and build the index from tuples visible in the snapshot
   - Resulting index is not yet valid for queries, and write operations.

2. Validation phase
   - Collect index entries (secondary key, primary key, xid)
   - Perform a second table scan to reconcile differences between the table and the partially built index
   - Increase the validation boundary.

3. Boundary-controlled insertion
   - The index does not become fully writable all at once, but rather in a monotonically increasing fashion.
   - Ensure tuple stability while transferring missing entries
   - Handle concurrent modifications without blocking OLTP workload

More details about the algorithm and design can be found in: [src/tableam/concurrent_index_build.md](https://github.com/orioledb/orioledb/pull/730/changes#diff-a21d846b805cf55d9d04463b2f8cdf38c9fb294bc7ee0a39d0929d4c60465d83)

uplink to postgres -> https://github.com/orioledb/postgres/pull/45

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -7,7 +7,7 @@ MODULE_big = orioledb
 EXTENSION = orioledb
 PGFILEDESC = "orioledb - orioledb transactional storage engine via TableAm"
 SHLIB_LINK += -lzstd -lcurl -lssl -lcrypto
-
+IS_DEV=1
 DATA_built = $(patsubst %_prod.sql,%.sql,$(wildcard sql/*_prod.sql))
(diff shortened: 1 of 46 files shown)


homper:
Closing in favor of #879

PR #733, opened 2026-02-24
Fix fetching partition's tuple when it has different attrs order than partitioned table itself

vvtrunov:


src/tuple/slot.c
```diff
@@ -233,6 +298,8 @@ tts_orioledb_getsomeattrs(TupleTableSlot *slot, int __natts)
 	bool		index_order;
 	int			cur_tbl_attnum = 0;
 
+	bool *isfilled = palloc0_array(bool, __natts);
```
pashkinelfe: I propose using MemoryContextAllocZero(slot->tts_mcxt...)

src/tuple/slot.c
```diff
@@ -456,9 +536,22 @@ tts_orioledb_getsomeattrs(TupleTableSlot *slot, int __natts)
...
+			}
+		}
+
+		/* Update the slot's valid attribute count. */
+		slot->tts_nvalid = first_unfilled;
+	}
 
-	/* Update the slot's valid attribute count. */
-	slot->tts_nvalid = natts;
+	pfree(isfilled);
```
pashkinelfe: I propose:
if (!is_bump_memory_context(slot->tts_mcxt))
pfree(isfilled);

src/tuple/slot.c
pashkinelfe, line 551: This could be reason of ci crash https://github.com/orioledb/orioledb/actions/runs/22443090784/job/64990340661

vvtrunov: The problem was that isfilled was allocated for _natts, but in case (oslot->ixnum == PrimaryIndexNumber && oslot->leafTuple && index_order) natts = descr->tupdesc->natt, where descr->tupdesc->natt can be greater than _natts and that can cause out-of-bound memory access

test/t/logical_test.py
```diff
@@ -2551,3 +2551,54 @@ def test_recvlogical_and_drop_database(self):
...
+			publisher.start()
+
+			subscriber = self.getSubsriber()
+			with subscriber.start() as subscriber:
+				setup_sql = """
+					CREATE EXTENSION IF NOT EXISTS orioledb;
+					CREATE TABLE tab1 (
+						a int PRIMARY KEY,
+						b text
+					) PARTITION BY LIST (a) USING orioledb;
```
pashkinelfe: Propose to remove `USING orioledb` from parent table. This is not supported in PG16, and parent table would not be orioledb anyway

Image

pashkinelfe:
@vvtrunov Could you squash into a single commit and add a commit message?

pashkinelfe:
Pushed to main with minor edit of a commit message

PR #737, opened 2026-02-26
tableam: assign heap XID in logical apply worker DML path

GremSnoort:
# Problem
In logical apply context, a transaction may reach commit without an assigned top-level heap XID.
For apply workers, this is critical because replication origin progress is advanced at commit via the regular commit machinery that expects a valid top XID. If no XID is assigned, origin progress may not be advanced correctly.

**Result:** potential replication-origin progress inconsistencies during logical replay through Oriole table AM paths.

This problem occurs on PG **subscription** test `025_rep_changes_for_schema.pl`.

# Solution
Add apply-worker context detection and lazily allocate top-level heap XID before Oriole table AM DML callbacks. This ensures logical apply commits have a valid top XID so replication-origin progress is advanced correctly, while keeping non-apply paths unchanged.

This PR adds a safety mechanism in the OrioleDB table AM for logical replication apply context: before DML operations, it now guarantees that a top-level heap XID is assigned when running inside an apply worker.


src/tableam/handler.c
```diff
@@ -72,6 +76,72 @@
...
+ * - replication origin is set to a valid origin id.
+ *
+ * DML operations should assign a heap XID during apply, as apply worker's replorigin_session_advance
+ * only occurs on commit with a valid top XID in RecordTransactionCommit.
+ *
+ * This check is intentionally strict to avoid assigning a heap XID for
+ * unrelated write paths, preserving Oriole's regular oxid-based behavior.
+ */
+static inline bool
+context_is_apply_worker(void)
```
pashkinelfe: I propose remanimg in_apply_worker_with_replorigin() or similar

src/tableam/handler.c
```diff
@@ -72,6 +76,72 @@
...
+ * To prevent that, in apply replay context we lazily allocate a heap XID if
+ * none has been assigned yet.  Outside apply replay context, this function is
+ * a no-op, so regular Oriole paths remain unaffected.
+ *
+ * This function is expected to be called at the entry of Oriole table AM DML
+ * callbacks (tuple_insert/tuple_update/tuple_delete/tuple_insert_with_arbiter,
+ * and once per multi_insert batch).
+ */
+static inline void
+assign_heap_xid_optionally(void)
```
pashkinelfe: I propose renaming assign_heap_xid_in_apply_worker_with_replorigin() or similar.
Or if this looks long: add /* Only for apply worker with valid replogin */ for each call

pashkinelfe (commented):
From the test in this PR (test_logical_subscription), I suppose that setting heap xid will affect the general case of logical replication. How would this affect the performance of logical apply?

Also I proposed a couple of renamings

src/tableam/handler.c
GremSnoort: Renamed in ccf9637fd4b31a84f2d55c788b3ac31a14bc4105

src/tableam/handler.c
GremSnoort: Renamed in ccf9637fd4b31a84f2d55c788b3ac31a14bc4105

PR #738, opened 2026-02-26
Add an argument to rewind functions to make restart optional 

pgnickb:
- Added a new argument attempt_restart (false by default) to all user facing rewind functions
- Bumped extension version to 1.7
- Updated the documentation

### DIFF
--- a/doc/usage/rewind.mdx
+++ b/doc/usage/rewind.mdx
@@ -47,17 +47,38 @@ While the Rewind Worker manages its "to-do list" in a fixed-size circular buffer
 
 ## Rewind Functions
 
-### `orioledb_rewind_by_time(seconds integer)`
+### `orioledb_rewind_by_time(seconds int4[, attempt_restart bool])`
 Rewinds the cluster state by the specified number of seconds from the current time.
-*   **Example:** `SELECT orioledb_rewind_by_time(600);`
 
-### `orioledb_rewind_to_timestamp(target_time timestamptz)`
+* **Parameters:**
+  * `seconds`: The number of seconds to rewind the cluster state.
+  * `attempt_restart`: If `true`, the database system will automatically restart after the rewind. If `false`, the database will only be shut down. Default is `false`.
+* **Examples:**
+  * `SELECT orioledb_rewind_by_time(600);` — Rewinds by 10 minutes and shuts down the database without restarting.
+  * `SELECT orioledb_rewind_by_time(600, true);` — Rewinds by 10 minutes and restarts the database.
+
+### `orioledb_rewind_to_timestamp(target_time timestamptz[, attempt_restart bool])`
 Rewinds the cluster to a specific point in time.
-*   **Example:** `SELECT orioledb_rewind_to_timestamp('2025-01-01 12:00:00 UTC');`
 
-### `orioledb_rewind_to_transaction(xid xid8, oxid int8)`
-Rewinds the cluster to a state **before** specific transaction pair identified by the PostgreSQL Transaction ID (`xid`) and the OrioleDB Transaction ID (`oxid`). `oxid` can be acquired by calling `orioledb_get_current_oxid()`.
-*   **Example:** `SELECT orioledb_rewind_to_transaction(1750, 555);`
+* **Parameters:**
+  * `target_time`: The exact timestamp to which the cluster state will be rewound.
+  * `attempt_restart`: If `true`, the database system will automatically restart after the rewind. If `false`, the database will only be shut down.  Default is `false`.
+* **Examples:**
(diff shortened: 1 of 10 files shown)


doc/usage/rewind.mdx
```diff
@@ -47,17 +47,38 @@ While the Rewind Worker manages its "to-do list" in a fixed-size circular buffer
...
+  * `SELECT orioledb_rewind_to_timestamp('2025-01-01 12:00:00 UTC');` — Rewinds to the specified timestamp and shuts down the database without restarting.
+
+---
+
+### `orioledb_rewind_to_transaction(xid int4, oxid int8, attempt_restart bool DEFAULT false)`
+Rewinds the cluster to a state **before** a specific transaction pair identified by the PostgreSQL Transaction ID (`xid`) and the OrioleDB Transaction ID (`oxid`).
+
+* **Parameters:**
+  * `xid`: The PostgreSQL Transaction ID.
+  * `oxid`: The OrioleDB Transaction ID (can be acquired by calling `orioledb_get_current_oxid()`).
```
pashkinelfe: (can be acquired by calling `orioledb_get_current_oxid()` this reads like you need to get oxid just before caliing rewind function which is not true

sql/orioledb--1.6--1.7_prod.sql
```diff
@@ -0,0 +1,27 @@
+/* contrib/orioledb/sql/orioledb--1.6--1.7_prod.sql */
+
+-- complain if script is sourced in psql, rather than via CREATE EXTENSION
+\echo Use "ALTER EXTENSION orioledb UPDATE TO '1.7'" to load this file. \quit
+
+DROP FUNCTION orioledb_rewind_by_time(int);
```
pashkinelfe: We'd better not drop old functions that are already in the release. Just add new functions with  the same name and extra arg

src/rewind/rewind.c
```diff
@@ -98,8 +98,9 @@ Datum
 orioledb_rewind_by_time(PG_FUNCTION_ARGS)
 {
 	int			rewind_time = PG_GETARG_INT32(0);
+	bool		attempt_restart = PG_GETARG_BOOL(1);
```
pashkinelfe: You may use (PG_NARGS() to detenmine if function called with extra arg or without.

doc/usage/rewind.mdx
pgnickb: I'm not sure I understand you correctly. Can you suggest a better wording?

sql/orioledb--1.6--1.7_prod.sql
pgnickb: We can't do that if we want to have `DEFAULT` argument. If we're okay with dropping the default and instead handling it in `C`, I can do that.

Why do you think it's a bad idea to drop the function here?

pashkinelfe: Let's ask @akorotkov on second opinion

doc/usage/rewind.mdx
pashkinelfe: I propose to drop what's in parenteses

sql/orioledb--1.6--1.7_prod.sql
pashkinelfe: I suppose we can get defaults in a new function and have an old function as they are not tied to each other in Sql

pgnickb: > get defaults in a new function

If by a new function you mean a function with a different name, then yes, we can do that.

pashkinelfe: I think of this way:
- Leave old function as it is
- Add new function with an extra argument

If someone calls without extra argument, he calls old function
If someone calls with extra agrument, that means he has already provided a value he wants (restart or not)

pgnickb: Yes, that's an option. 

But I'm still curious as to why we shouldn't drop old functions?

pashkinelfe: I don't remember details precisely, but this caused some problems when migrating between AMI's in Supabase and we agreed not to drop old sql functions even if they are not needed.

pgnickb: I understand not dropping the C function so that the upgrade path doesn't fail when looking up the symbol in the .so. This one is news to me. Thanks.

pashkinelfe: As an example `orioledb_tbl_indices()` functions in `orioledb--1.0` and `orioledb--1.3--1.4`

pgnickb: Yeah the idea is clear. That's exactly what I meant when saying:

>  If we're okay with dropping the default and instead handling it in C, I can do that.

PR #753, opened 2026-03-06
Include OrioleDB tables size in counting database size

pashkinelfe:
Orioledb relations reside in a separate orioledb_data dir, for which calculate_database_size was unaware for.

This patch implements hook function to count orioledb_data files both
 for datadir and pg_tblspc/ dir.

Limitation for counting: orioledb_undo is not per-database so it's not
 counted now. This could underestimate reported database size results
for workload heavy with DML modifications.

### DIFF
--- a/src/orioledb.c
+++ b/src/orioledb.c
@@ -158,6 +158,8 @@ static void (*prev_shmem_request_hook) (void) = NULL;
 static base_init_startup_hook_type prev_base_init_startup_hook = NULL;
 static get_relation_info_hook_type prev_get_relation_info_hook = NULL;
 static skip_tree_height_hook_type prev_skip_tree_height_hook = NULL;
+database_size_hook_type prev_database_size_hook = NULL;
+
 CheckPoint_hook_type next_CheckPoint_hook = NULL;
 static bool o_newlocale_from_collation(void);
 
@@ -1173,6 +1175,8 @@ _PG_init(void)
 	waitSnapshotHook = orioledb_wait_snapshot;
 	GetReplayXlogPtrHook = recovery_get_effective_replay_ptr;
 
+	prev_database_size_hook = database_size_hook;
+	database_size_hook = orioledb_calculate_database_size;
 	RecoveryStopsBeforeHook = orioledb_recovery_stops_before_hook;
 
 	if (enable_rewind)
--- a/test/sql/database.sql
+++ b/test/sql/database.sql
@@ -23,3 +23,88 @@ CREATE DATABASE orioledb TEMPLATE orioledb_template;
 DROP DATABASE orioledb_template;
 DROP DATABASE heapdb;
 DROP DATABASE heapdb_template;
+
+-- Check pg_database_size()
+CREATE EXTENSION orioledb;
+CREATE DATABASE oriole_database;
+\c oriole_database
+
+-- generate pseudo-random string function in deterministic way
+CREATE FUNCTION generate_string(seed integer, length integer) RETURNS text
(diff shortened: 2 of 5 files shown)


src/orioledb.c
```diff
@@ -151,13 +151,15 @@ int			rewind_max_time = 0;
...
 bool		orioledb_strict_mode = false;
+database_size_hook_type prev_database_size_hook = NULL;
 
 /* Previous values of hooks to chain call them */
 static shmem_startup_hook_type prev_shmem_startup_hook = NULL;
 static void (*prev_shmem_request_hook) (void) = NULL;
 static base_init_startup_hook_type prev_base_init_startup_hook = NULL;
 static get_relation_info_hook_type prev_get_relation_info_hook = NULL;
 static skip_tree_height_hook_type prev_skip_tree_height_hook = NULL;
+
```
za-arthur: I propose to move the hook variable to hooks section:
```suggestion
/* Previous values of hooks to chain call them */
static shmem_startup_hook_type prev_shmem_startup_hook = NULL;
static void (*prev_shmem_request_hook) (void) = NULL;
static base_init_startup_hook_type prev_base_init_startup_hook = NULL;
static get_relation_info_hook_type prev_get_relation_info_hook = NULL;
static skip_tree_height_hook_type prev_skip_tree_height_hook = NULL;
database_size_hook_type prev_database_size_hook = NULL;

```

test/sql/database.sql
```diff
@@ -23,3 +23,60 @@ CREATE DATABASE orioledb TEMPLATE orioledb_template;
...
+LANGUAGE SQL;
+
+CREATE EXTENSION orioledb;
+CHECKPOINT;
+select round(pg_database_size('oriole_database'), -6);
+
+CREATE TABLE oriole_table (i SERIAL PRIMARY KEY, t text STORAGE PLAIN) USING orioledb;
+INSERT INTO oriole_table(t) select generate_string(i, 270) FROM  generate_series(1, 10000) as i;
+CHECKPOINT;
+select round(pg_database_size('oriole_database'), -6);
```
za-arthur, line 48: I suggest to add tablespace test:

```
SET allow_in_place_tablespaces = true;
CREATE TABLESPACE regress_tblspace LOCATION '';

CREATE TABLE oriole_table_tblspc (i SERIAL PRIMARY KEY USING INDEX TABLESPACE regress_tblspace, t text STORAGE PLAIN) USING orioledb TABLESPACE regress_tblspace;
INSERT INTO oriole_table_tblspc(t) select generate_string(i, 270) FROM  generate_series(1, 10000) as i;
CHECKPOINT;
select round(pg_database_size('oriole_database'), -6);

...
DROP TABLESPACE regress_tblspace;
```

test/sql/database.sql
```diff
@@ -23,3 +23,78 @@ CREATE DATABASE orioledb TEMPLATE orioledb_template;
...
+
+CREATE TABLE oriole_table_tblspc (i SERIAL PRIMARY KEY, t text STORAGE PLAIN) USING orioledb TABLESPACE regress_tblspace;
```
za-arthur: BTW the primary key won't be created on `regress_tblspace` tablespace unless you specify it with:
```
i SERIAL PRIMARY KEY USING INDEX TABLESPACE regress_tblspace
```
Not sure if you intended to create it also on `regress_tblspace` or on the default tablespace.

pashkinelfe: I did this for simplicity; it doesn't affect the test. Can change this.

Issue #721, opened 2026-02-13
`ON CONFLICT` with large `STORAGE MAIN` unique key fails with index row size overflow


ImreSamu:
version:  [beta14](https://github.com/orioledb/orioledb/releases/tag/beta14)   (  `docker.io/orioledb/orioledb:beta14-pg17` ) 

SQL:
```sql
DROP TABLE IF EXISTS t CASCADE;
CREATE TABLE t (
	id bigint PRIMARY KEY,
	bt text STORAGE MAIN
) 
USING orioledb;
-- USING heap;
CREATE UNIQUE INDEX t_bt_uidx ON t (bt);
INSERT INTO t VALUES (1, repeat('t', 9000));
INSERT INTO t VALUES (2, (SELECT bt FROM t WHERE id = 1))
  ON CONFLICT (bt) DO UPDATE
     SET bt = EXCLUDED.bt;
SELECT id, length(bt) FROM t ORDER BY id;
```

# Expected
- `INSERT ... ON CONFLICT (bt) DO UPDATE ...` should resolve conflict without failing.
- Existing large key value should be reusable in conflict path.

# Actual
- Fails at conflicting insert with:
  *   `ERROR:  index row size 9024 exceeds orioledb maximum 2688 for index "t_bt_uidx"`
- First insert succeeds, second insert with same large key in `ON CONFLICT` path fails.

LOG:
```sql
psql (17.6 OrioleDB public beta 14 PGTAG=patches17_16 alpine:3.21+clang build:2026-02-10T17:38:39+00:00 17.6)
Type "help" for help.

postgres=# DROP TABLE IF EXISTS t CASCADE;
NOTICE:  table "t" does not exist, skipping
DROP TABLE
postgres=# CREATE TABLE t (
postgres(# id bigint PRIMARY KEY,
postgres(# bt text STORAGE MAIN
postgres(# ) 
postgres-# USING orioledb;
CREATE TABLE
postgres=# -- USING heap;
postgres=# CREATE UNIQUE INDEX t_bt_uidx ON t (bt);
CREATE INDEX
postgres=# INSERT INTO t VALUES (1, repeat('t', 9000));
INSERT 0 1
postgres=# INSERT INTO t VALUES (2, (SELECT bt FROM t WHERE id = 1))
postgres-#   ON CONFLICT (bt) DO UPDATE
postgres-#      SET bt = EXCLUDED.bt;
[...]

PR #757, opened 2026-03-12
Fix "index row size exceeds maximum" when building index on compressed toasted data

### LINKED ISSUE #756: Index creation fails if large tuple inserted before index build (compression not reapplied)
There is inconsistent behavior when working with large values in a column with `STORAGE EXTENDED/MAIN` in **OrioleDB** depending on whether the index is created **before** or **after** the large tuple insertion.

If the index is created **first** and the large tuple is inserted afterward, everything works correctly.

However, if the large tuple is inserted **before** the index is created, then **index creation fails** with:

```
ERROR:  index row size 10018 exceeds orioledb maximum 2688 for index "idx"
```

This suggests that during index build the value is **decompressed but not recompressed**, and the raw large value is used as the index key, causing the index tuple size to exceed the allowed limit.

#### Case 1 — insert before index creation (fails)

```sql
CREATE TABLE cmdata(a int, f1 text STORAGE EXTENDED) using orioledb;

INSERT INTO cmdata VALUES(1, repeat('1234567890', 1000));

CREATE INDEX idx ON cmdata(f1);
```

Result:

```
ERROR:  index row size 10018 exceeds orioledb maximum 2688 for index "idx"
```

Verification:

```sql
SELECT length(f1) FROM cmdata WHERE a = 1;
```

Cleanup:

```sql
DROP TABLE cmdata;
```

---

#### Case 2 — insert after index creation (works)

```sql
CREATE TABLE cmdata(a int, f1 text STORAGE EXTENDED) using orioledb;

CREATE INDEX idx ON cmdata(f1);

INSERT INTO cmdata VALUES(1, repeat('1234567890', 1000));

SELECT length(f1) FROM cmdata WHERE a = 1;

DROP TABLE cmdata;
```

This case works without errors.

### Expected behavior

Index creation should succeed regardless of whether large tuples are inserted **before** or **after** index creation.

### Possible cause

During index build, large values may be **decompressed but not recompressed**, and the uncompressed value is used to form the index key, exceeding the index tuple size limit.

Serge-sudo:
CREATE INDEX on a column containing existing PGLZ/LZ4-compressed data fails with index row size N exceeds orioledb maximum 2688, while inserting data into a pre-existing index on the same column works fine.

Root Cause
o_get_tbl_att() calls PG_DETOAST_DATUM on all VARATT_IS_EXTENDED values, which includes inline compressed varlenas. During build_secondary_index_worker_heap_scan, this decompresses stored PGLZ values back to their full size (e.g., 10000 bytes), blowing past the index tuple size limit.

fixes #756 and #721

### DIFF
--- a/include/tuple/slot.h
+++ b/include/tuple/slot.h
@@ -90,7 +90,7 @@ extern bool tts_orioledb_modified(TupleTableSlot *oldSlot,
 								  Bitmapset *attrs);
 extern void tts_orioledb_set_ctid(TupleTableSlot *slot, ItemPointer iptr);
 extern Datum o_get_tbl_att(TupleTableSlot *slot, int attnum, bool primaryIsCtid,
-						   bool *isnull, Oid *typid);
+						   bool *isnull, Oid *typid, bool decompress);
 Datum		o_get_idx_expr_att(TupleTableSlot *slot, OIndexDescr *idx,
 							   ExprState *exp_state, bool *isnull);
 
--- a/src/tableam/operations.c
+++ b/src/tableam/operations.c
@@ -538,7 +538,7 @@ exclusion_fill_bound(TupleTableSlot *slot, OIndexDescr *idx, OBTreeKeyBound *bou
 
 		if (attnum != EXPR_ATTNUM)
 			value = o_get_tbl_att(slot, attnum, idx->primaryIsCtid,
-								  &isnull, &typid);
+								  &isnull, &typid, true);
 		else
 		{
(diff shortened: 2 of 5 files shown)


test/expected/compression_storage.out
```diff
@@ -0,0 +1,113 @@
+/*
+ * Test for compression of extended storage.
```
pashkinelfe: It's better to add tests to the existing file toast_column_compress.sql, not create a new one for this purpose

PR #758, opened 2026-03-13
Add SQL function to report undo log sizes

pashkinelfe:
Results are output separately for row-level, page-level ans system Undo logs This function could be useful for monitoring undo logs


sql/orioledb--1.6--1.7_prod.sql
```diff
@@ -18,3 +18,7 @@ RETURNS void
 AS 'MODULE_PATHNAME'
 VOLATILE LANGUAGE C;
 
+CREATE FUNCTION orioledb_undo_size(OUT row_level int8, OUT page_level int8, OUT system int8)
```
za-arthur: What about if we return the following table instead:
| undo_type | undo_size |
| ----------- | ----------- |
| row | size |
| page | size |
| system | size |

This way it will be easier to extend the undo log with new undo log types in the future without changing this function.

src/tableam/handler.c
```diff
@@ -1368,6 +1378,98 @@ orioledb_relation_toast_am(Relation rel)
...
+	/* Build a tuple descriptor for our result type */
+	if (get_call_result_type(fcinfo, NULL, &tupdesc) != TYPEFUNC_COMPOSITE)
+		elog(ERROR, "return type must be a row type");
+
+	tupstore = tuplestore_begin_heap(true, false, work_mem);
+	rsinfo->returnMode = SFRM_Materialize;
+	rsinfo->setResult = tupstore;
+	rsinfo->setDesc = tupdesc;
+
+	MemoryContextSwitchTo(oldcontext);
```
za-arthur: I think instead of doing this manually you can use the `InitMaterializedSRF` function:
```C
InitMaterializedSRF(rsinfo, 0);
```

pashkinelfe: Using this will throw error `set-valued function called in context that cannot accept a set`
As I see InitMaterializedSRF is never used in Orioledb in similar scenarios.
Probably, it would be a separate task to sort out how to make all Orioledb functions use this simplified workflow

sql/orioledb--1.6--1.7_prod.sql
pashkinelfe: @za-arthur  Modified in this way

src/tableam/handler.c
za-arthur: I don't have this error. If you have the error `set-valued function called in context that cannot accept a set`, then something is wrong because the function `orioledb_undo_size` should be set returning function.

I tried this changes and it works:

```diff
diff --git a/src/tableam/handler.c b/src/tableam/handler.c
index 1ba60ad2..0992a505 100644
--- a/src/tableam/handler.c
+++ b/src/tableam/handler.c
@@ -1386,29 +1386,13 @@ Datum
 orioledb_undo_size(PG_FUNCTION_ARGS)
 {
        ReturnSetInfo *rsinfo = (ReturnSetInfo *) fcinfo->resultinfo;
-       TupleDesc       tupdesc;
-       Tuplestorestate *tupstore;
-       MemoryContext per_query_ctx;
-       MemoryContext oldcontext;
        struct dirent *direntry;
        DIR                *dirdesc;
        char            filename[MAXPGPATH * 2];
        const char      path[15] = "orioledb_undo";
        int                     totalsize[UndoLogsCount];

-       per_query_ctx = rsinfo->econtext->ecxt_per_query_memory;
-       oldcontext = MemoryContextSwitchTo(per_query_ctx);
-
-       /* Build a tuple descriptor for our result type */
-       if (get_call_result_type(fcinfo, NULL, &tupdesc) != TYPEFUNC_COMPOSITE)
-               elog(ERROR, "return type must be a row type");
-
-       tupstore = tuplestore_begin_heap(true, false, work_mem);
-       rsinfo->returnMode = SFRM_Materialize;
-       rsinfo->setResult = tupstore;
-       rsinfo->setDesc = tupdesc;
-
-       MemoryContextSwitchTo(oldcontext);
+       InitMaterializedSRF(fcinfo, 0);

        memset(totalsize, 0, sizeof(totalsize));

@@ -1478,7 +1462,7 @@ orioledb_undo_size(PG_FUNCTION_ARGS)
                nulls[1] = false;
                values[1] = totalsize[i];

-               tuplestore_putvalues(tupstore, tupdesc, values, nulls);
+               tuplestore_putvalues(rsinfo->setResult, rsinfo->setDesc, values, nulls);
        }

        return (Datum) 0;
```

test/t/functions_test.py
```diff
@@ -0,0 +1,53 @@
...
+
+import subprocess
+
+from .base_test import BaseTest
+from testgres.exceptions import QueryException
+
+
+class FunctionTest(BaseTest):
+
+	def test_undo_log_size(self):
```
za-arthur, line 13: Cannot we have just a regression test instead of a python test? I think it is simple enough for that.
A regression test would be simpler and maybe faster.

pashkinelfe: Regression test will gather undo log modifications by all tests in parallel schedule, It won't be similar in different running conditions (parallel, serial)

src/tableam/handler.c
pashkinelfe: @za-arthur  It looks to work indeed, thank you!
Maybe I forgot to clean the repository after changing sql function types
Added to PR

za-arthur: Thanks for the fix!

Issue #785, opened 2026-03-17
Parallel regression tests hang due to eviction stall with small OrioleDB main buffer


Serge-sudo:
When running PostgreSQL native regression tests with OrioleDB as the default table access method and a small main buffer size (e.g. 8MB), parallel test suites may get stuck in eviction logic.

Observed behavior:
- The main shared buffer pool becomes fully occupied.
- Backends attempt to allocate new pages but are unable to find free slots.
- Eviction is not making progress.
- As a result, all participating backends stall and the regression test suite hangs indefinitely.

Environment:
- OrioleDB as default table access method
- orioledb_main_buffers_size = 8MB
- PostgreSQL native regression tests
- Parallel test execution enabled

Expected behavior:
- Eviction should make forward progress even under high contention and small buffer sizes.
- Backends should not enter a global stall while waiting for free pages.

Actual behavior:
- System enters a state where no backend is able to evict pages or proceed with allocation.

PR #786, opened 2026-03-17
Fix root page eviction for CTID PK, smgr init loop, reserved pages race, and add shared pool inspection function

Serge-sudo:
Fix multiple issues related to page management, eviction, and initialization.

1. Fix eviction of root pages for CTID-based primary keys.
   Previously, such root pages were not evicted, which led to accumulation
   of these pages together with seq buffer pages that are evicted alongside
   their corresponding root pages. This could result in an infinite search
   for free slots in the page pool.

2. Adjust pool reservation logic to return pages before eviction.
   When a backend takes multiple pages and then needs to perform eviction,
   it now puts back already taken pages before proceeding. This prevents
   other backends from observing an artificially depleted pool and
   unnecessarily entering eviction.

3. Fix checkpoint logic for unlogged pages.
   Previously, checkpoint did not recreate seq buffers for unlogged pages,
   leading to inconsistent state. The logic is now corrected to properly
   handle these cases.

4. Add auxiliary SQL function to inspect pool contents.
   The function takes a pool type identifier and returns the list of pages
   currently present in the corresponding pool, allowing inspection of
   catalog, main, and free pools.

fix for #785

### DIFF
--- a/doc/architecture/checkpoints.mdx
+++ b/doc/architecture/checkpoints.mdx
@@ -82,3 +82,56 @@ This message indicates that the child page has been processed, and the parent ne
 ### WalkContinue
 
 This message has no parameters. It just indicates that checkpointer must continue processing the same page with the same `next downlink`. That happens when the checkpointer has to wait for the concurrent operation. Such as meeting IO in-process downlink and having to release the log and wait till the IO is finished.
+
+## Sequential buffers
+
+### What are sequential buffers?
+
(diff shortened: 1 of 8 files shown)


sql/orioledb--1.0_prod.sql
```diff
@@ -162,6 +162,12 @@ RETURNS bool
 AS 'MODULE_PATHNAME'
 VOLATILE LANGUAGE C;
 
+CREATE FUNCTION orioledb_print_main_pool_pages(OUT block_num int8, OUT name TEXT, OUT datoid int8, OUT reloid int8, OUT relnode int8, OUT type TEXT, OUT state int8)
```
pashkinelfe: New functions should be added to last version sql (now orioledb--1.6--1.7_prod.sql)

PR #791, opened 2026-03-19
Provide a durable commit LSN for xid-less Oriole commits

GremSnoort:
# Summary

This patch adds Oriole-side support for PostgreSQL xid-less commit integration and fixes origin propagation in Oriole logical decoding.

It contains two related but independent changes:
- support for the new PostgreSQL xid-less commit hook by supplying a durable local commit LSN for top-level independent Oriole transactions
- preservation of WAL container origin metadata during logical decoding so origin-based filtering works correctly

The immediate motivation is fixing logical replication failures on subscriber apply for transactions that commit without a top-level heap XID but still have a real durable commit effect.

# Problem

With the PostgreSQL-side xid-less commit hook in place, a storage engine must be able to supply a durable local commit LSN before `RecordTransactionCommit()` finishes.

For OrioleDB, the durable commit point for an independent Oriole transaction is produced by Oriole commit machinery, not by the regular PostgreSQL heap-XID commit path.

Before this patch:
- the final Oriole commit WAL record was produced later in `undo_xact_callback(XACT_EVENT_COMMIT)`
- PostgreSQL had no earlier way to obtain the correct durable local commit LSN

Oriole WAL containers can carry `origin_id` / `origin_lsn`, but before this patch container-level origin metadata was parsed separately from the main WAL record flow.

As a result, origin metadata could be lost before DML and commit records reached logical decoding, which broke origin-based filtering.

# Root Cause
The PostgreSQL patch introduces a transaction-level hook, but Oriole still needed to provide the missing commit anchor at the right point in its own lifecycle.

Conceptually, the problem was:
- PostgreSQL needs the durable local commit LSN during `RecordTransactionCommit()`
- Oriole originally materialized that LSN only later, in `undo_xact_callback(XACT_EVENT_COMMIT)`
- autonomous Oriole transactions use a separate internal commit path and must not be exposed through the new xid-less hook
- container-level origin metadata was parsed separately from main WAL records, which could lose `origin_id` / `origin_lsn` during logical decoding

Without addressing these points, xid-less subscriber apply remained incomplete.

# Solution
This patch teaches Oriole to participate in the new PostgreSQL xid-less commit path.

It does so by:
- adding `orioledb_get_independent_commit_lsn()`, which supplies a durable local commit LSN for a top-level independent Oriole transaction
- materializing and caching that LSN early enough for PostgreSQL to use it in `RecordTransactionCommit()`
- reusing the cached LSN later in `undo_xact_callback(XACT_EVENT_COMMIT)` instead of producing a second commit WAL record
- preserving the cached LSN across autonomous transaction nesting
- keeping callback-side flush and synchronous replication waits aligned with the final commit LSN selected by PostgreSQL
- preserving WAL container `origin_id` / `origin_lsn` in logical decoding so origin-based filtering continues to work correctly

This keeps replication-origin policy in PostgreSQL core while letting Oriole provide the durable commit anchor required by the new xid-less commit hook.

This patch changes WAL container flag parsing to populate the same `WalRecord` instance that is later passed through normal record processing.


src/recovery/wal_reader.c
```diff
@@ -297,11 +297,12 @@ wal_container_read_header(WalReaderState *r, bool allow_logging)
  * misread.
  */
 static inline WalParseResult
-wal_container_parse_flag(WalReaderState *r, wal_type_t type)
+wal_container_parse_flag(WalReaderState *r, WalRecord *rec, wal_type_t type)
```
pashkinelfe: Looks like flags parsing is per container and we'd better not mix it with things related to WalRecord's

include/transam/undo.h
```diff
@@ -279,6 +279,7 @@ typedef struct
 	bool		needs_wal_flush;
 	bool		has_retained_undo_location[(int) UndoLogsCount];
 	bool		local_wal_has_material_changes;
+	XLogRecPtr	independent_commit_lsn;
```
pashkinelfe: I think "independent" is to be replaced with something more semantically clear

src/recovery/wal_reader.c
```diff
@@ -459,7 +457,7 @@ parse_wal_container(WalReaderState *r, bool allow_logging)
 	 * The flag handling also gives the consumer a chance to capture
 	 * header-wide context (e.g. xact-info) before any records are delivered.
 	 */
-	st = wal_container_parse_flags(r);
+	st = wal_container_parse_flags(r, &rec);
```
pashkinelfe: I think common per-container variables is better to be transferred via WalReaderState (in similar way as wal_version, see R446)

include/transam/undo.h
GremSnoort: Renamed in 5803f3d03518f6e62680e69d99493157bc37635a
`independent_commit_lsn` -> `xidless_commit_lsn`

src/recovery/wal_reader.c
GremSnoort: Yes, the flags are indeed container-level.

Passing `WalRecord *rec` into flag parsing is not accidental; it matches the current parser-state model. It may make sense **to change the naming of this structure in a separate PR**.

Details:
`parse_wal_container` creates a single `WalRecord rec` for the whole container parse in `wal_reader.c`.

This `rec` is:
- first populated with container-level data from the header/flags,
- then reused for each record inside the container.

This is necessary because record parsing depends on sticky context propagated across records.

We need shared context across records, including:
- wal_version,
- origin info,
- transaction info from container flags,
- sticky state parsed from one record and used by subsequent ones, such as `oids`, `logicalXid`, `heapXid`, and `relreplident`.

So in the current design, `WalRecord` is effectively a reused parser-state object for the whole container, not just a per-record payload holder.

za-arthur: I agree with @pashkinelfe , this is a per container function. If you want to keep container level variables it is better to introduce a separate struct (like `WalContainer`, maybe better in a separate PR).

include/transam/undo.h
pashkinelfe: orioledb_get_independent_commit_lsn is remaining

src/transam/undo.c
```diff
@@ -197,6 +198,70 @@ static OBuffersDesc undoBuffersDesc =
...
+ * Materialize a durable WAL anchor for an independent Oriole commit.
+ */
+static XLogRecPtr
+finalize_independent_commit_wal(OXid oxid, TransactionId logicalXid)
+{
+	XLogRecPtr flushPos = InvalidXLogRecPtr;
+
+	Assert(OXidIsValid(oxid));
+
+	if (OXidIsValid(oxid))
```
pashkinelfe: redundant after Assert

src/transam/undo.c
```diff
@@ -197,6 +198,70 @@ static OBuffersDesc undoBuffersDesc =
...
+		GET_CUR_PROCDATA()->autonomousNestingLevel == 0 &&
+		!TransactionIdIsValid(GetTopTransactionIdIfAny()) &&
+		TransactionIdIsValid(logicalXid);
+}
+
+/*
+ * Materialize a durable WAL anchor for an independent Oriole commit.
+ */
+static XLogRecPtr
+finalize_independent_commit_wal(OXid oxid, TransactionId logicalXid)
```
pashkinelfe: This function is called only with invalid xidless_commit_lsn
It's better to push this check into the function to decrease external checks

src/transam/undo.c
```diff
@@ -197,6 +198,70 @@ static OBuffersDesc undoBuffersDesc =
...
+ * path. Return InvalidXLogRecPtr if there is no independent Oriole commit
+ * anchor for the current transaction.
+ */
+XLogRecPtr
+orioledb_get_independent_commit_lsn(void)
+{
+	OXid			oxid = get_current_oxid_if_any();
+	LogicalXidCtx	logicalXidContext;
+
+	get_current_logical_xid_ctx(&logicalXidContext);
```
pashkinelfe: The same as xid = get_current_logical_xid() Why the context is needed here?

src/transam/undo.c
```diff
@@ -197,6 +198,70 @@ static OBuffersDesc undoBuffersDesc =
...
+
+	if (!is_independent_oriole_transaction(oxid, logicalXidContext.xid))
```
pashkinelfe: One-caller fn. Better inline it with comments

src/transam/undo.c
```diff
@@ -197,6 +198,70 @@ static OBuffersDesc undoBuffersDesc =
...
+		return InvalidXLogRecPtr;
+
+	if (XLogRecPtrIsInvalid(xidless_commit_lsn))
+	{
+		if (TransactionIdIsValid(logicalXidContext.xid))
```
pashkinelfe: This check could also be inlined into finalize_independent_commit_wal

src/transam/undo.c
GremSnoort: Done in 5c47a3254092438d0d05fe9d9d781069744fb391

src/transam/undo.c
GremSnoort: Done in 5c47a3254092438d0d05fe9d9d781069744fb391

src/transam/undo.c
GremSnoort: Done in 5c47a3254092438d0d05fe9d9d781069744fb391

src/transam/undo.c
GremSnoort: Done in efb91ad0104eb6bc4736f5b36ee89041d6f4de59

src/transam/undo.c
GremSnoort: Done in efb91ad0104eb6bc4736f5b36ee89041d6f4de59

GremSnoort: Done in efb91ad0104eb6bc4736f5b36ee89041d6f4de59

include/transam/undo.h
pashkinelfe: also finalize_independent_commit_wal remains

src/transam/undo.c
```diff
@@ -197,6 +198,52 @@ static OBuffersDesc undoBuffersDesc =
...
+	OXid			oxid = get_current_oxid_if_any();
+	TransactionId	logicalXid = get_current_logical_xid();
+
+	/*
+	 * Filter out autonomous transactions.
+	 * PG xid-less commit hook is only for top-level Oriole transactions that
+	 * participate in logical apply/origin tracking. Autonomous transactions use
+	 * their own commit path and must stay invisible to this hook.
+	 */
+	if (!(OXidIsValid(oxid) &&
```
pashkinelfe: I propose rewrite condition without  if(!(clause1 && !clause2))

include/transam/undo.h
GremSnoort: Renamed in 134a2968836c9f9fdb4d84d21d42ac00189b36da

src/transam/undo.c
GremSnoort: Done in eb7e74cd13d7085c38879e1505bf9b2d8159e4dc

src/transam/undo.c
```diff
@@ -197,6 +198,52 @@ static OBuffersDesc undoBuffersDesc =
 static void wait_for_reserved_location(UndoLogType undoType,
 									   UndoLocation undoLocationToWait);
 
+/*
+ * Materialize a durable WAL anchor for an independent Oriole commit.
+ */
+static XLogRecPtr
+finalize_xidless_commit_wal(OXid oxid, TransactionId logicalXid)
```
pashkinelfe: Function is called with get_current_logical_xid() as second arg. Call of get_current_logical_xid() could be inlined into the fn, and we can avoid second arg.

src/transam/undo.c
```diff
@@ -2144,14 +2192,15 @@ undo_xact_callback(XactEvent event, void *arg)
...
+					flushPos = finalize_xidless_commit_wal(oxid, get_current_logical_xid());
 
-					current_oxid_xlog_precommit();
+					elog(DEBUG4, "XACT_EVENT_COMMIT [independent Oriole transaction] oxid %lu logicalXid %u top heapXid %u current heapXid %u useHeap %d flushPos %X/%X",
+						 oxid, logicalXidContext.xid, heapXid, GetCurrentTransactionIdIfAny(), logicalXidContext.useHeap, LSN_FORMAT_ARGS(flushPos));
 
-					flushPos = wal_commit(oxid, get_current_logical_xid(),
-										  false);
-					set_oxid_xlog_ptr(oxid, flushPos);
+					if (XLogRecPtrIsInvalid(flushPos) ||
```
pashkinelfe: Could this be changed to flushPos = Max(flushPos, XactLastCommitEnd) ?

src/transam/undo.c
```diff
@@ -2142,16 +2190,15 @@ undo_xact_callback(XactEvent event, void *arg)
...
 
-					elog(DEBUG4, "XACT_EVENT_COMMIT [independent Oriole transaction] oxid %lu logicalXid %u top heapXid %u current heapXid %u useHeap %d",
-						 oxid, logicalXidContext.xid, heapXid, GetCurrentTransactionIdIfAny(), logicalXidContext.useHeap);
+					flushPos = get_xidless_commit_lsn(oxid, &wrote_xlog);
 
-					current_oxid_xlog_precommit();
+					elog(DEBUG4, "XACT_EVENT_COMMIT [independent Oriole transaction] oxid %lu logicalXid %u top heapXid %u current heapXid %u useHeap %d flushPos %X/%X",
+						 oxid, logicalXidContext.xid, heapXid, GetCurrentTransactionIdIfAny(), logicalXidContext.useHeap, LSN_FORMAT_ARGS(flushPos));
 
-					flushPos = wal_commit(oxid, get_current_logical_xid(),
```
pashkinelfe, line 2205: Previous to this patch we called wal_commit unconditionally, now we call it here only if XLogRecPtrIsInvalid(xidless_commit_lsn)

GremSnoort: Yes, previously `wal_commit()` was called unconditionally here.

With this patch it is no longer guaranteed to happen specifically in `XACT_EVENT_COMMIT`, because the xid-less commit WAL may already have been materialized earlier through the new PG hook path.

So the behavior change is intentional: we still want `wal_commit()` to happen for xid-less Oriole commits, but only once. If it has not been done earlier, `xidless_commit_lsn` is still invalid and this path calls `wal_commit()` as before. If it has already been done earlier, we skip the duplicate call here.

Issue #789, opened 2026-03-19
Assertion failure in btree_ctid_get_and_inc due to self-invalidation resetting metapage after o_btree_load_shmem


Serge-sudo:
#### **Problem**

Running the attached script intermittently leads to the following assertion failure:

```
TRAP: failed Assert("OInMemoryBlknoIsValid((desc)->rootInfo.metaPageBlkno)")
File: "src/btree/btree.c", Line: 268
```

Backtrace:

```
#0  0x00007f7f725bf52f in raise () from /lib64/libc.so.6
#1  0x00007f7f72592e65 in abort () from /lib64/libc.so.6
#2  0x0000000000b01128 in ExceptionalCondition (
    conditionName=conditionName@entry=0x7f7f6b9f0c50 "OInMemoryBlknoIsValid((desc)->rootInfo.metaPageBlkno)", 
    fileName=fileName@entry=0x7f7f6b9f0a40 "src/btree/btree.c", lineNumber=lineNumber@entry=268) at assert.c:66
#3  0x00007f7f6b9423be in btree_ctid_get_and_inc (desc=desc@entry=0x17a9900) at ../../src/include/port/atomics.h:520
#4  0x00007f7f6b9cc8d3 in o_tbl_insert (descr=descr@entry=0x17a6420, relation=relation@entry=0x7f7f56fc8288, slot=slot@entry=0x18f2bc0, oxid=78, 
    csn=902) at src/tableam/operations.c:317
#5  0x00007f7f6b96eda8 in rewrite_table (rel=0x7f7f56fc8288, old_o_table=0x18eddc0, new_o_table=0x18f2368) at src/catalog/ddl.c:1959
#6  0x00007f7f6b972869 in orioledb_object_access_hook (access=OAT_POST_ALTER, classId=1259, objectId=16610, subId=0, arg=0x7fff1f019488)
    at src/catalog/ddl.c:3347
#7  0x00000000006b5d03 in RunObjectPostAlterHook (classId=classId@entry=1259, objectId=objectId@entry=16610, subId=subId@entry=0, 
    auxiliaryId=auxiliaryId@entry=0, is_internal=is_internal@entry=true) at objectaccess.c:104
#8  0x0000000000776a87 in RenameRelationInternal (myrelid=16610, newrelname=newrelname@entry=0x7fff1f019640 "pg_toast_16601", 
    is_internal=is_internal@entry=true, is_index=is_index@entry=false) at tablecmds.c:4549
[...]

PR #792, opened 2026-03-19
Fix loss of root/meta blkno after self-invalidation; defer invalidation processing under page locks

Serge-sudo:
Fix an issue where B-tree descriptor state (`metaPageBlkno` and `rootPageBlkno`) could be lost after processing an invalidation message within the same transaction.

Previously, after `o_btree_load_shmem()` initialized the descriptor, a invalidation message could be processed before the descriptor was used. This caused the descriptor to be recreated, resetting `metaPageBlkno` and `rootPageBlkno` to `InvalidBlockNumber`, which led to assertion failures (e.g. in `btree_ctid_get_and_inc`).

While investigating, another problem was identified: invalidation message processing may occur while holding page locks. Invalidation handlers access system trees, and if required pages are evicted, they trigger page loads. This leads to `load_page()` being called while locks are held, hitting assertions that forbid holding locks during page load.

To address both issues:

* Ensure descriptor state is not lost due to mid-operation self-invalidation
* Introduce deferred invalidation handling:

  * Skip invalidation message processing when it is unsafe (e.g. while holding page locks)
  * Process pending invalidation messages at the next safe point

This prevents descriptor corruption and avoids unsafe page loads under locks.

uplink-> https://github.com/orioledb/postgres/pull/55

fix for #789

### DIFF
--- a/include/orioledb.h
+++ b/include/orioledb.h
@@ -32,6 +32,7 @@
 #include "utils/typcache.h"
 #include "utils/rel.h"
 #include "utils/relcache.h"
+#include "utils/inval.h"
 
 #if defined __has_include
 #if __has_include ("sanitizer/asan_interface.h")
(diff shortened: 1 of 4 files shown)


akorotkov:
@Serge-sudo, thank you!
The corresponding bug is now fixed using some ideas from this PR.

PR #812, opened 2026-04-03
Fix dangling symlink in pg_tblspc

bwyltcaxe:
After DROP TABLESPACE the dangling symlink is left at standby.  OrioleDB creates its own data directories and directory cannot be removed by PG core because it's not empty.  This patch removes no more used directories for dropped tablespaces

### DIFF
--- a/src/recovery/recovery.c
+++ b/src/recovery/recovery.c
@@ -3231,6 +3231,9 @@ typedef struct
 	OTableDescr *descr;
 	OIndexDescr *indexDescr;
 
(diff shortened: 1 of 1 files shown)


homper:
This pr will be no longer suitable after merging [this branch](https://github.com/orioledb/orioledb/commit/963ec0a32eda943fff2e2e0fd42f4f3ab8f58b77)
It is probably needed to check for the existence of directories when deleting tables, indexes, and databases when replaying wal now

bwyltcaxe:
Got it. Close

PR #815, opened 2026-04-08
Refactor wal_reader

za-arthur:
Changes:
- decouple WAL containers and records by adding new `WalContainer` struct - before the refactoring `WalRecord` struct contains WAL records and containers level information. Decoupling allows to simplify future modifications and provides better readibility
- move `ORIOLE_WAL_RECORDS` into the new `wal_record.h` file to avoid cycling dependency between `wal.h` and `wal_reader.h` files
- move WAL reader related declarations from `wal.h` to `wal_reader.c`
- remove `ORIOLE_WAL_FLAGS` - it looks unnecessary since there is only two flags and it makes the code rather more complex and harder to read


include/recovery/wal_reader.h
```diff
@@ -164,8 +165,8 @@ struct WalReaderState;
  * based on semantic compatibility rules.
  */
 typedef WalParseResult (*WalCheckVersionFn) (const struct WalReaderState *r);
-typedef WalParseResult (*WalOnFlagFn) (void *ctx, const WalRecord *rec);
-typedef WalParseResult (*WalOnRecordFn) (void *ctx, WalRecord *rec);
+typedef WalParseResult (*WalOnContainerFn) (struct WalReaderState *r);
+typedef WalParseResult (*WalOnRecordFn) (struct WalReaderState *r, WalRecord *rec);
```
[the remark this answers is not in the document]
za-arthur: Fixed

src/recovery/recovery.c
```diff
@@ -980,48 +980,23 @@ recovery_wal_check_version(const WalReaderState *r)
...
-typedef struct
-{
-	TimestampTz xactTime;
-	TransactionId xid;
-
-} RecoveryWalDescCtx;
-
 static WalParseResult
-recovery_wal_containter_info(void *vctx, const WalRecord *rec)
+recovery_on_containter(WalReaderState *r)
```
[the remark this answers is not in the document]
za-arthur: Fixed

src/recovery/logical.c
```diff
@@ -648,37 +655,24 @@ typedef struct
...
+	if (r->container.flags & WAL_CONTAINER_HAS_ORIGIN_INFO)
 	{
-		case WAL_CONTAINER_HAS_XACT_INFO:
-			/* Skip WAL_REC_XACT_INFO */
-			break;
-
-		case WAL_CONTAINER_HAS_ORIGIN_INFO:
-			ctx->has_origin = rec->origin_id != InvalidRepOriginId;
-			break;
+		DecodeWalDescCtx *ctx = (DecodeWalDescCtx *) r->ctx;
```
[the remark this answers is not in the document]
za-arthur: Fixed

src/recovery/recovery.c
```diff
@@ -980,48 +1103,23 @@ recovery_wal_check_version(const WalReaderState *r)
...
 
 	return WALPARSE_EOF;		/* Stop parser */
 }
 
 static WalParseResult
-recovery_wal_record(void *vctx, WalRecord *rec)
+recovery_on_record(WalReaderState *r, WalRecord *rec)
 {
 	return WALPARSE_EOF;		/* Stop parser */
 }
```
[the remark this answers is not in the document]
za-arthur: The current behavior is consistent with the old one

src/recovery/wal_reader.c
```diff
@@ -355,26 +463,23 @@ wal_container_parse_flag(WalReaderState *r, wal_type_t type)
...
 
-#undef X
+	if (r->container.flags & WAL_CONTAINER_HAS_ORIGIN_INFO)
+	{
+		WR_PARSE(r, &r->container.origin_info.id);
+		WR_PARSE(r, &r->container.origin_info.lsn);
+	}
 
-		return st;
+	return WALPARSE_OK;
```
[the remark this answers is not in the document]
za-arthur: The corruption is protected with checksums and we have version check already.

include/recovery/wal_record.h
```diff
@@ -0,0 +1,69 @@
...
+	X(WAL_REC_TRUNCATE,             13, "TRUNCATE",           wal_parse_rec_truncate) \
+	X(WAL_REC_BRIDGE_ERASE,         14, "BRIDGE_ERASE",       wal_parse_rec_bridge_erase) \
+	X(WAL_REC_REINSERT,             15, "REINSERT",           wal_parse_rec_modify) \
+	X(WAL_REC_REPLAY_FEEDBACK,      16, "REPLAY_FEEDBACK",    wal_parse_empty) \
+	X(WAL_REC_SWITCH_LOGICAL_XID,   17, "SWITCH_LOGICAL_XID", wal_parse_rec_switch_logical_xid) \
+	X(WAL_REC_RELREPLIDENT,         18, "RELREPLIDENT",       wal_parse_rec_relreplident)
+
+typedef enum
+{
+#define X(sym, val, name, fn) sym = (val),
```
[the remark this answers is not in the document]
za-arthur: This is by design

src/recovery/logical.c
```diff
@@ -1312,13 +1336,11 @@ orioledb_decode(LogicalDecodingContext *ctx, XLogRecordBuffer *buf)
...
-		.wal_version = 0,
-		.wal_flags = 0,
-		/* Consumer */
+		.container = {0},
 		.ctx = &dctx,
-		.check_version = decode_wal_check_version,
-		.on_flag = decode_wal_containter_info,
-		.on_record = decode_wal_record
+		.check_version = decode_check_version,
+		.on_container = decode_on_container,
```
pashkinelfe, line 1342: I'd propose decode_wal_check_version/decode_wal_containter/decode_wal_record naming for .check_version/.on_container/.on_record events

src/recovery/wal_reader.c
```diff
@@ -222,31 +419,15 @@ wal_container_read_header(WalReaderState *r, bool allow_logging)
 
 	if (wal_version > ORIOLEDB_WAL_VERSION)
 	{
-#ifdef IS_DEV
```
GremSnoort, line 225: We are now removing the old `#ifdef IS_DEV` behavior from `wal_container_read_header`, even though this existed before my first PR with the new WAL parser.

Previously, the logic was explicit:
- in dev/test builds (`IS_DEV`), any mismatch between `wal_version` and `ORIOLEDB_WAL_VERSION` resulted in `elog(FATAL, ...)`
- in regular builds, the behavior was softer: `WARNING`/`LOG`, and the further handling remained caller-specific

So this was not just logging, but an intentional fail-fast mode for tests:
- always fail tests on version mismatch
- otherwise, leave the final error handling to the caller

With the current change, we are effectively removing that behavior.
❓  Is this an intentional semantic change? Do we really want to remove the fail-fast semantics on version mismatch in `wal_container_read_header` entirely?

If fail-fast behavior for test/dev builds is still desirable, then maybe we should not drop it silently, but either:
- restore equivalent handling in `wal_container_read_header`
- properly wire `-DIS_DEV` into the build system for test builds and preserve the previous behavior explicitly

Otherwise, this patch changes not only the parser structure, but also the version-mismatch policy in tests.

include/catalog/sys_trees.h
```diff
@@ -163,8 +164,5 @@ extern PrintFunc sys_tree_key_print(BTreeDescr *desc);
 extern PrintFunc sys_tree_tup_print(BTreeDescr *desc);
 extern void sys_tree_set_extra(int tree_num, Pointer extra);
 extern Pointer sys_tree_get_extra(int tree_num);
-#ifdef IS_DEV
-extern const text *inspect_sys_tree_structure(int systree, int depth);
-#endif
```
GremSnoort, line 168: I would prefer to keep `inspect_sys_tree_structure`. It was added intentionally as a debug helper, because otherwise this call had to be reconstructed manually during debugging. Even though it is small, it makes sys tree inspection much more convenient during debugging, so I do not think it should be removed as dead code cleanup.

src/recovery/wal_reader.c
za-arthur: I think we don't pass `IS_DEV` to C code. It is used only by Makefile to generate .sql script file. Therefore it didn't have any effect before.

include/catalog/sys_trees.h
za-arthur: Is it used anywhere in the code? Anyway it is wrapped inside `IS_DEV` which we don't pass to C code as define.
I can revert this change. In that case would it make sense to remove the `IS_DEV` check and export it as an SQL function similar to other `IS_DEV` functions, in a separate PR?

src/recovery/wal_reader.c
za-arthur: But in general, indeed, this change (or other related to `IS_DEV`) need to be a separate PR and this isn't related to the WAL reader refactoring.

include/catalog/sys_trees.h
GremSnoort: This helper is not used in production code paths; it was meant as a small C-side debugging helper for local investigation. It is typically needed for extended logging and monitoring of system trees, and that logging is usually removed again before the PR is finalized, so no in-tree usage of the helper remains.

However, when system tree monitoring is needed again later, the same wrapper has to be reconstructed manually. In practice, this code fragment has already been reused informally multiple times for that purpose.

I agree that the current `#ifdef IS_DEV` form is not very useful if `IS_DEV` is not actually passed to C compilation. In that case, I think either of these would be reasonable:
- keep this change reverted for now, so we do not lose the helper entirely
- then, in a separate PR, either remove the `IS_DEV` guard or expose it in a more consistent way, depending on how we want to treat debug helpers in general

I would prefer not to drop the helper in this refactoring, since its removal is orthogonal to the WAL parser changes.

src/recovery/wal_reader.c
GremSnoort: Yes, `IS_DEV` does not actually work in this context today, because it is not passed into the C compilation flags. However, the surrounding logic is already present in the code and it may be useful.

I think in this PR we can simply avoid changing the current `IS_DEV`-related logic.
Then, in a separate follow-up PR, we can decide whether this is still behavior we want to preserve:
- fail-fast for tests on WAL version mismatch
- otherwise, leave the final error handling to the caller

And from there decide whether `IS_DEV` should be properly wired into C compilation, or removed entirely.

GremSnoort:
An important functional fix has already landed as part of this refactoring: container-level flag metadata is now visible during record parsing.

Previously, `origin_id` / `origin_lsn` came from the container flag payload, but that metadata was not propagated properly across record parsing, so individual records did not reliably see it. With the new container-level state, that information is preserved and remains available while parsing all records within the same container.

This also fixes PostgreSQL subscription test `030_origin.pl`, so I think it should now be possible to add it to the PG repo's `src/test/subscription/oriole_tests.txt`.

src/recovery/wal_reader.c
za-arthur: I reverted the change related to `IS_DEV`

include/catalog/sys_trees.h
za-arthur: I reverted the change related to  `IS_DEV`.

za-arthur: @copilot create a github issue based on this conversation and link this https://github.com/orioledb/orioledb/pull/815#discussion_r3111095637
Labels: enhancement

src/recovery/recovery.c
```diff
@@ -1066,17 +1038,19 @@ orioledb_recovery_stops_before_hook(XLogReaderState *record,
 	if (XLogRecGetDataLen(record) == 0)
 		return false;
 
-	st = parse_wal_container(&r, true /* allow_logging */ );
```
pashkinelfe, line 1069: I think parse_wal_container is completely ok, no need to rename it. Or rename it to wal_parse_container (wal_**something**_parse is not present in naming of similar functions)

src/recovery/wal_reader.c
```diff
@@ -15,108 +15,305 @@
...
+
+		rec->u.modify.t2.data = r->ptr;
+		WR_SKIP(r, rec->u.modify.len2);
+	}
+
+	return WALPARSE_OK;
+}
+
+/*
+ * This file implements the generic OrioleDB WAL container parser.
```
pashkinelfe, line 296: I think this comment is out-of-place. It could be tied to exact parse function, not a file overall. I'm also for maning it more concise.

src/recovery/logical.c
za-arthur: I renamed the functions.

src/recovery/recovery.c
za-arthur: I renamed it to `wal_parse_container` to make it consistent with parse functions:
https://github.com/orioledb/orioledb/blob/wal_record_refactor/include/recovery/wal_record.h#L43-L60

src/recovery/wal_reader.c
za-arthur: Agree, I moved the comment to the top of the file and shortened it.

src/recovery/recovery.c
pashkinelfe: Thank you! Actually parse_wal_container was in agreement with naming of decode_wal_containter. If we choose wal_parse_container, then decode_wal_check_version/decode_wal_containter/decode_wal_record should become wal_decode_check_version/wal_decode_containter/wal_decode_record

pashkinelfe: The similar names for replay_check_version/replay_on_container/replay_on_record  would be wal_replay_check_version/wal_replay_container/wal_replay_record

pashkinelfe: And for recovery_check_version/recovery_on_container/recovery_on_record similar names would be wal_recovery_check_version/wal_recovery_container/wal_recovery_record

za-arthur: I think that those are the local callbacks and we shouldn't consider them when naming exported API. We may ignore them.

pashkinelfe: I agree local callbacks don't need to follow API functions naming. Still I think it's good to use the same naming among all local callbacks. Now `decode` and `wal_desc` callbacks are named differently from `recovery` and `replay` ones:
```
.check_version = decode_wal_check_version,
		.on_flag = decode_wal_containter_info,
		.on_container = decode_wal_container,
		.on_record = decode_wal_record
```

```
.check_version = recovery_check_version,
		.on_container = recovery_on_container,
		.on_record = recovery_on_record
```
```
.check_version = replay_check_version,
		.on_container = replay_on_container,
		.on_record = replay_on_record
```
```
.check_version = wal_desc_check_version,
		.on_flag = NULL,
		.on_container = NULL,
		.on_record = wal_desc_on_record
```

za-arthur: I renamed logical decoder callbacks back to their original names, which were in same naming style as other callbacks. I don't think that they should be same as the main API functions.

pashkinelfe (approved):

PR #819, opened 2026-04-15
Wake startup when worker is caught up but finishedPtr lags

bwyltcaxe:
When a worker's queue is empty and its commitPtr has reached recovery_ptr but recovery_finished_list_ptr has not yet been updated, wake the startup process to immediately run update_proc_retain_undo_location(), assigning CSNs to committed transactions and updating finishedPtr. This eliminates up to 10-second replay lag visible in pg_stat_replication without spurious wakeups in idle state.


bwyltcaxe:
Took in https://github.com/orioledb/orioledb/commit/62902bbc4d8f93d65e88b8694b270810235c0ba6

PR #833, opened 2026-04-23
Synchronize Oriole recovery state at recovery target stop boundary

### LINKED ISSUE #820: Fix recovery/003_recovery_targets PostgreSQL test
## Test failure

**Test:** `src/test/recovery/t/003_recovery_targets.pl`

## Error

```
Failed test 'check standby content for XID'
at t/003_recovery_targets.pl line 116.
  got: '1000'
  expected: '2000'
```

## Analysis

PostgreSQL considers the standby ready for read-only queries before
OrioleDB has completed bringing replayed data to an SQL-visible state.
Read-only readiness in PostgreSQL and replay visibility in OrioleDB
are not synchronized:

- The LSN-based recovery target case is synchronized with OrioleDB's
  effective replay pointer.
- The XID-based recovery target case is **not** synchronized.

`recovery_target_xid` stops recovery upon the `RM_XACT COMMIT` record
before OrioleDB makes the commit replay-visible on the standby.

Furthermore, in the OrioleDB code path,
`INSERT INTO tab_int VALUES (generate_series(1001,2000))` is committed
as an independent OrioleDB transaction without a PostgreSQL heap XID.
As a result, the subsequent `SELECT pg_current_xact_id()` obtains a
different heap XID, and the test erroneously uses it as
`recovery_target_xid` for the preceding insert.

For independent OrioleDB transactions, there are two distinct stages
on the standby:
1. WAL record replayed — container read, COMMIT inside OrioleDB WAL
   processed.
2. Changes become SQL-visible — state passes through `finished_list`,
   `recovery_finished_list_ptr` is advanced.

OrioleDB considers a replay point visible only after stage 2. For
independent OrioleDB commits, the real visibility barrier occurs only
after worker-side completion at `ctx->xlogRecEndPtr`, whereas the
main/startup process completes its commit path earlier at an earlier
internal pointer.

## Possible Fix

For independent OrioleDB commits, visible completion must be
synchronized on the worker-side final container end pointer rather
than on the earlier main-process finish marker.

Add a new function in `src/recovery/recovery.c`:

```c
static void workers_synchronize_used(XLogRecPtr ptr)
{
    int i;
    for (i = 0; i < recovery_pool_size_guc && !unexpected_worker_detach; i++)
    {
        int j = 0;
        if (!cur_recovery_xid_state->used_by[i] &&
            !cur_recovery_xid_state->checkpoint_xid)
            continue;
        while (pg_atomic_read_u64(&worker_ptrs[i].commitPtr) < ptr &&
               workers_pool[i].queue)
        {
            BgwHandleStatus status;
            pid_t pid;
            pg_usleep(10);
            if (j % 100 == 0)
            {
                status = GetBackgroundWorkerPid(workers_pool[i].handle, &pid);
                if (status != BGWH_STARTED && status != BGWH_NOT_YET_STARTED)
                {
                    unexpected_worker_detach = true;
                    break;
                }
            }
            j++;
        }
    }
}
```

Inside `replay_wal_record`, before `recovery_finish_current_oxid`,
add the following for independent OrioleDB commits (`heapXid = 0`):

```c
if (commit && !ctx->single &&
    !TransactionIdIsValid(rec->heapXid))
{
    workers_synchronize_used(ctx->xlogRecEndPtr);
}
```

This ensures that for independent OrioleDB commits, the main/startup
process waits for worker-side completion at `ctx->xlogRecEndPtr`
before calling `recovery_finish_current_oxid()`, rather than
completing the main-side path prematurely.

GremSnoort:
## Summary

Hook Oriole into PostgreSQL's `RecoveryTargetReachedHook` and use the provided stop-boundary metadata to synchronize Oriole recovery state before `recovery_target_action` is applied.

The new barrier runs in the startup process and aligns:
- replay progress
- retain/undo bookkeeping
- deferred finalization/publication

with the exact recovery stop boundary selected by PostgreSQL.

## Problem

During targeted recovery, PostgreSQL can already consider the requested recovery target reached while Oriole-visible state still lags behind the same stop boundary.

As a result, a relation read immediately after recovery stops may still observe stale Oriole state even though PostgreSQL has already exposed the target-reached state.

## Changes

- register `orioledb_recovery_target_reached_hook`
- use `RecoveryTargetReachedInfo` provided by PostgreSQL
- drive startup-side deferred finalization/publication from the hook
- wait until Oriole recovery boundaries converge on the target stop boundary
- use `recordPtr` as the explicit stop boundary for both stop-before and stop-after target semantics

## Result

This makes target-point visibility for Oriole relations consistent with PostgreSQL recovery stop semantics during targeted recovery.

In particular, Oriole-visible state is now synchronized before PostgreSQL proceeds with `recovery_target_action`.

## Testing

Run from PostgreSQL's repo:

```bash
TEMP_CONFIG=${ORIOLEDB_CONFIG} \
PG_TEST_INITDB_EXTRA_OPTS="--locale=C --encoding=UTF8" \
make -C src/test/recovery installcheck PROVE_TESTS='t/003_recovery_targets.pl
```

## Issue

Closes: #820


src/recovery/recovery.c
```diff
@@ -1056,6 +1058,156 @@ orioledb_recovery_stops_before_hook(XLogReaderState *record,
...
+	/*
+	 * recordPtr is the most useful explicit stop boundary for both cases:
+	 * - for stop-before semantics it is the start of the record that must not
+	 *   become visible yet, so Oriole must at least catch up to the boundary
+	 *   immediately before it;
+	 * - for stop-after semantics it is the record that belongs to the visible
+	 *   target state, so replay/finalization must catch up through that
+	 *   boundary before PostgreSQL exposes the target-reached state.
+	 */
+	if (XLogRecPtrIsValid(info->recordPtr))
```
pashkinelfe: This check is unnecessary. Else target_ptr == InvalidXLogRecPtr anyway

src/recovery/recovery.c
```diff
@@ -1056,6 +1058,156 @@ orioledb_recovery_stops_before_hook(XLogReaderState *record,
...
+ *
+ * The hook runs in the startup process itself, so it cannot just wait for
+ * some other startup-side path to advance recovery_finished_list_ptr; it must
+ * actively drive startup-side publication while waiting for the boundaries to
+ * converge.
+ */
+void
+orioledb_recovery_target_reached_hook(const RecoveryTargetReachedInfo *info)
+{
+	XLogRecPtr	target_ptr = InvalidXLogRecPtr;
```
pashkinelfe: Assignment would be unnecessary when check at #1090 will be removed

src/recovery/recovery.c
```diff
@@ -1056,6 +1058,156 @@ orioledb_recovery_stops_before_hook(XLogReaderState *record,
...
+		elog(DEBUG4,
+			 "Recovery target reached: boundary snapshot after startup-side "
+			 "publication attempt (target_ptr=%X/%X current_ptr=%X/%X "
+			 "retain_ptr=%X/%X finished_ptr=%X/%X)",
+			 LSN_FORMAT_ARGS(target_ptr),
+			 LSN_FORMAT_ARGS(current_ptr),
+			 LSN_FORMAT_ARGS(retain_ptr),
+			 LSN_FORMAT_ARGS(finished_ptr));
+
+		if (XLogRecPtrIsValid(target_ptr) && current_ptr < target_ptr)
```
pashkinelfe: For clarity of logic I propose to convert these combined if's to a two nested ifs. External: if (XLogRecPtrIsValid(target_ptr)), and internal: all other conditions.

Is behaviour with XLogRecPtrIsValid(target_ptr)==false expected to appear and in which cases?

src/recovery/recovery.c
GremSnoort: Done in cbcb4fc9320322cc3477a3e191201c91c8e1e4a2

src/recovery/recovery.c
GremSnoort: Done in cbcb4fc9320322cc3477a3e191201c91c8e1e4a2

pashkinelfe:
@GremSnoort I've resolved conflicts in patchsets versions to run CI. Now CI fails https://github.com/orioledb/orioledb/actions/runs/25866615392/job/76010366876

src/recovery/recovery.c
```diff
@@ -1450,26 +1532,672 @@ orioledb_recovery_stops_before_hook(XLogReaderState *record,
...
 		*recordXtime = r.container.xact_info.xactTime;
 
 		if (recoveryTargetInclusive)
-			return r.container.xact_info.xactTime > recoveryTargetTime;
+			stop_here = r.container.xact_info.xactTime > recoveryTargetTime;
 		else
-			return r.container.xact_info.xactTime >= recoveryTargetTime;
+			stop_here = r.container.xact_info.xactTime >= recoveryTargetTime;
+
+		xact_time_str = pstrdup(timestamptz_to_str(*recordXtime));
```
pashkinelfe: Can we avoid all allocations like this in non-debug cases?

src/recovery/recovery.c
```diff
@@ -1450,26 +1532,672 @@ orioledb_recovery_stops_before_hook(XLogReaderState *record,
...
+
+		if (info->recoveryStopAfter &&
+			XLogRecPtrIsValid(last_replay_ptr_snapshot) &&
+			XLogRecPtrIsValid(last_visible_ptr_snapshot))
+		{
+			stop_after_replay_ptr = last_replay_ptr_snapshot;
+			stop_after_visible_ptr = last_visible_ptr_snapshot;
+		}
+
+		elog(DEBUG2,
```
pashkinelfe: Any reason to have DEBUG2 and DEBUG4 in different places?

src/recovery/recovery.c
```diff
@@ -1450,26 +1532,672 @@ orioledb_recovery_stops_before_hook(XLogReaderState *record,
...
+					 "main_retain_ptr=%X/%X current_ptr=%X/%X "
+					 "retain_ptr=%X/%X finished_ptr=%X/%X)",
+					 LSN_FORMAT_ARGS(target_ptr),
+					 LSN_FORMAT_ARGS(recovery_ptr_snapshot),
+					 LSN_FORMAT_ARGS(main_retain_ptr),
+					 LSN_FORMAT_ARGS(current_ptr),
+					 LSN_FORMAT_ARGS(retain_ptr),
+					 LSN_FORMAT_ARGS(finished_ptr));
+			else
+				elog(DEBUG4,
```
pashkinelfe: Are all debug blocks reachable? Any chance to have common helper(s) to print them?

src/recovery/recovery.c
```diff
@@ -796,11 +845,26 @@ recovery_shmem_init(Pointer ptr, bool found)
 	worker_ptrs = (RecoveryWorkerPtrs *) ptr;
 	ptr += CACHELINEALIGN(mul_size(sizeof(RecoveryWorkerPtrs), recovery_pool_size_guc + recovery_idx_pool_size_guc));
 
+	recovery_last_pair_seq = (pg_atomic_uint32 *) ptr;
```
pashkinelfe: Could it be combined to a compact block after recovery_ptr?

src/recovery/recovery.c
```diff
@@ -1450,26 +1532,672 @@ orioledb_recovery_stops_before_hook(XLogReaderState *record,
...
+				/*
+				 * For stop-before of the first post-backup transaction, the
+				 * correct visible state is the base-backup state itself.  In
+				 * that case Oriole may have no published post-backup boundary
+				 * yet, and all recovery/publication pointers remain invalid.
+				 * Treat that as a valid already-published boundary instead of
+				 * waiting forever for a boundary that must not appear before
+				 * stop.
+				 */
+				if (!XLogRecPtrIsValid(recovery_ptr_snapshot) &&
```
pashkinelfe: I'd propose to replace !XLogRecPtrIsValid -> XLogRecPtrIsInvalid everywhere

src/recovery/recovery.c
```diff
@@ -1450,26 +1532,672 @@ orioledb_recovery_stops_before_hook(XLogReaderState *record,
...
+ *
+ * Recovery-target synchronization must not compare PostgreSQL's stop record
+ * boundary directly with recovery_published_visible_ptr or current_ptr, since
+ * those pointers live in different semantic coordinate systems.  Instead,
+ * replay records the latest replay/visible pair as commit-producing events
+ * are applied, and the recovery-target hook later derives the Oriole-visible
+ * boundary relevant to the current stop condition from that latest pair.
+ */
+static inline void
+recovery_record_visible_boundary_pair(XLogRecPtr replay_ptr, XLogRecPtr visible_ptr)
```
pashkinelfe, line 2222: Both writers (replay_on_record, o_xact_redo_hook) and the only reader (the hook) run in the startup process; writes and reads never interleave on different CPUs. The seqlock looks redundant

src/recovery/recovery.c
```diff
@@ -1450,26 +1532,672 @@ orioledb_recovery_stops_before_hook(XLogReaderState *record,
...
+					 LSN_FORMAT_ARGS(stop_after_visible_ptr),
+					 LSN_FORMAT_ARGS(main_retain_ptr),
+					 LSN_FORMAT_ARGS(current_ptr),
+					 LSN_FORMAT_ARGS(retain_ptr),
+					 LSN_FORMAT_ARGS(finished_ptr));
+			break;
+		}
+
+		WakeupRecovery();
+		pg_usleep(200);
```
pashkinelfe: Are 200 not tight here and won't spin hard?

src/recovery/recovery.c
```diff
@@ -1450,26 +1532,672 @@ orioledb_recovery_stops_before_hook(XLogReaderState *record,
...
+ */
+static inline void
+recovery_record_visible_boundary_pair(XLogRecPtr replay_ptr, XLogRecPtr visible_ptr)
+{
+	uint32		seq;
+
+	if (!XLogRecPtrIsValid(replay_ptr) || !XLogRecPtrIsValid(visible_ptr))
+		return;
+
+	seq = pg_atomic_read_u32(recovery_last_pair_seq);
```
pashkinelfe: The writer reads with pg_atomic_read_u32, writes with pg_atomic_write_u32 — that's a read-modify-write without a single atomic; if another writer ever appears this races. I'd propose Asserting that there is no concurrent access ("called only on startup") and use uint32 for recovery_last_pair_seq

test/t/recovery_test.py
```diff
@@ -2880,16 +3132,1187 @@ def test_recovery_target_time(self):
...
+			replica.poll_query_until("SELECT pg_is_wal_replay_paused()",
+			                         expected=True)
+
+			self._assert_visible_series(replica, 3000, 3000)
+
+			replica_log = self._read_replica_log_until_barrier_completed(
+			    replica, stop_after=False)
+
+			self._assert_stop_before_barrier_logs(replica_log)
+			self._assert_log_contains_any(replica_log, [
```
pashkinelfe: Tests based on particular debug log messages look fragile. Not sure if there is an alternative way to expose state for the tests

src/recovery/recovery.c
GremSnoort: Done in bdb899f3c5fdc2b24211075db78076271d1e9c06

src/recovery/recovery.c
GremSnoort: Done in a8ca94ddccef0e8e7db15049662063194319d284

src/recovery/recovery.c
GremSnoort: Done in 7162817c8b6ca143b4fe2dbdb41ac8ca0894b8de

src/recovery/recovery.c
GremSnoort: Done in 3f1fb5f975ea57fd62514a601c61b741e7892dca

src/recovery/recovery.c
GremSnoort: Done in a3d361696d91d6cf2eba23ff103e9ed66b68834d

src/recovery/recovery.c
GremSnoort: Done in a3d361696d91d6cf2eba23ff103e9ed66b68834d and 83ff33928ff77921027ee69309115a39545c588a

src/recovery/recovery.c
GremSnoort: Upd in b2a0f07cd2e1f6dfc4b48615cdd56d5026fe7a53

src/recovery/recovery.c
GremSnoort: Addressed in 6790466f and b37bfe49.

b37bfe49 removes the duplicated stop-before/stop-after debug-only detach blocks and moves the common handling to `recovery_check_worker_detach` / `recovery_abort_if_worker_detached`.

6790466f also turns non-converging/impossible recovery-target boundary states into ERRORs instead of leaving them as debug wait paths. The remaining DEBUG4 blocks are regular barrier state snapshots/wait diagnostics and are reachable from the stop-before/stop-after branches.

test/t/recovery_test.py
GremSnoort: Please take a look at this commit 32285ed10982f53acc42bcac527cd2f6ed0f4c13
If this approach looks good, I can convert the remaining log-based checks the same way.

src/recovery/recovery.c
```diff
@@ -1450,26 +1537,722 @@ orioledb_recovery_stops_before_hook(XLogReaderState *record,
...
+ */
+static inline void
+recovery_record_visible_boundary_pair(XLogRecPtr replay_ptr, XLogRecPtr visible_ptr)
+{
+	Assert(AmStartupProcess() || !IsUnderPostmaster);
+
+	if (XLogRecPtrIsInvalid(replay_ptr) || XLogRecPtrIsInvalid(visible_ptr))
+		return;
+
+	pg_atomic_write_u64(recovery_last_replay_ptr, replay_ptr);
```
pashkinelfe: Do we need atomic when reads and writes can't be concurrent?

src/recovery/recovery.c
```diff
@@ -1450,26 +1537,722 @@ orioledb_recovery_stops_before_hook(XLogReaderState *record,
...
+	pg_atomic_write_u64(recovery_last_visible_ptr, visible_ptr);
```
pashkinelfe: Do we need atomic when reads and writes can't be concurrent?

src/recovery/recovery.c
```diff
@@ -2706,6 +3501,116 @@ free_run_xmin(void)
...
+ * backwards.  The startup process is the only writer.
+ */
+static void
+recovery_publish_visible_ptr(XLogRecPtr ptr)
+{
+	XLogRecPtr	old_ptr;
+
+	Assert(AmStartupProcess() || !IsUnderPostmaster);
+
+	old_ptr = pg_atomic_read_u64(recovery_published_visible_ptr);
```
pashkinelfe, line 3596: Could there be any concurrency around recovery_published_visible_ptr? If no, it could be made non-atomic. If yes, race is possible: recovery_published_visible_ptr could be invalid at the moment of reading, but already valid at L3611 and overwritten.

test/t/recovery_test.py
pashkinelfe: Looks good

src/recovery/recovery.c
pashkinelfe: Is perform_spin_delay() possible here?

src/recovery/recovery.c
pashkinelfe: I think still there are many similar elog printing blocks that call for some common formatting helpers

src/recovery/recovery.c
GremSnoort: Done in 2eab5e0d2a592e5dd3fe62949f9591127ba53875.
This pair is only accessed by the startup process, so I switched it from `pg_atomic_uint64` shmem slots to plain `XLogRecPtr` variables.

src/recovery/recovery.c
GremSnoort: Done in 2eab5e0d2a592e5dd3fe62949f9591127ba53875

src/recovery/recovery.c
GremSnoort: The model is:
- startup process: single writer
- recovery workers / startup hook: concurrent readers

There can be concurrent readers, so I think this should stay atomic.
Recovery workers may read `recovery_published_visible_ptr` while the startup process publishes it.

However, there is only one writer: `recovery_publish_visible_ptr()` is startup-process-only, and worker paths call `update_proc_retain_undo_location()` with `worker_id >= 0`, so they never publish.

Therefore the overwrite race should not be possible; atomics are needed for cross-process visibility, not for multi-writer synchronization.

src/recovery/recovery.c
GremSnoort: I would avoid `perform_spin_delay` here. This is not a spinlock-style short wait: the barrier can legitimately wait for recovery workers / retain publication progress, especially on slow CI or under valgrind. The current loop uses `pg_usleep(1000L)` and processes startup interrupts on each iteration, so it should not spin tightly and still remains responsive to shutdown/promotion.

src/recovery/recovery.c
GremSnoort: I agree there is still some repetition, but most of these DEBUG4 blocks describe different barrier branches and intentionally print branch-specific state. I already moved the duplicated detach handling into a common helper. For the remaining logs, a shared formatter would need either a **wide argument list** or a **dedicated context struct**, which seems **less readable** for this patch. I can still do that if you strongly prefer, but I would keep them explicit for now.

pashkinelfe: I understand this point. However my feelings about the patch consisting of so much debug messages formatting are mixed. And it could be hard to catch the logic of the change for the future reader.  I propose the following approach:
- Verify if all debug messages added, reachable
- Remove non-diagnostic fields in debug messages
- Combine diagnostic fields into re-usable blocks and assemble debug messages from them rather than copying formatting code. IIRC a part of differences between debug messages now is due to order of variables printed.

test/t/recovery_test.py
pashkinelfe: Pending conversion of the remaining log-based checks in tests

Issue #834, opened 2026-04-24
Invalid pages after primary key' change during perform_writeback_and_relock


bwyltcaxe:
During perform_writeback_and_relock the table is unlocked and it's possible to change the primary key for it. Hence, invalidation doesn't happen and then we fail at assertion for getting page in seq_buf_finalize.

Backtrace itself:
```
#0  0x00007f57d708d02c in __pthread_kill_implementation () from /lib64/libc.so.6
#1  0x00007f57d703fb86 in raise () from /lib64/libc.so.6
#2  0x00007f57d7029873 in abort () from /lib64/libc.so.6
#3  0x0000000000d0a9e4 in ExceptionalCondition (
    conditionName=0x7f57c6fd3578 "OInMemoryBlknoIsValid(shared->pages[shared->curPageNum])", 
    fileName=0x7f57c6fd3238 "src/utils/seq_buf.c", lineNumber=509) at assert.c:66
#4  0x00007f57c6fa6cd9 in seq_buf_finalize (seqBufPrivate=0x3218160) at src/utils/seq_buf.c:509
#5  0x00007f57c6f14100 in checkpoint_temporary_tree (flags=256, descr=0x3218090) at src/checkpoint/checkpoint.c:1663
#6  0x00007f57c6f1fb14 in checkpoint_tables_callback (type=oIndexPrimary, treeOids=..., tableOids=..., arg=0x7ffe459bd2b0)
    at src/checkpoint/checkpoint.c:5017
```

The reproducer (using gdb!):
1) Connect to checkpointer using gdb
2) in psql create temporary table and get know its oid, e.g.:
```
postgres=# create temporary table test(id int);
postgres=# insert into test values (0);
postgres=# select oid from pg_class where relname = 'test';
```
3) Set breakpoint in gdb:
```
b perform_writeback_and_relock if desc->oids.reloid == <oid>
```
4) Then when we stay after o_tables_rel_unlock_extended at gdb and in psql:
```
ALTER TABLE test ADD PRIMARY KEY (id);
```
5) `continue` at gdb and assert is triggered.

bwyltcaxe:
+ it looks like https://github.com/orioledb/orioledb/issues/784 at 329-2487764.backtrace.txt

bwyltcaxe:
Closed in https://github.com/orioledb/orioledb/pull/840

PR #840, opened 2026-04-29
Remove dead else in checkpoint_tables_callback

e-ivkov:
Shows that https://github.com/orioledb/orioledb/issues/834 should now (after local page pool introduction) be impossible


bwyltcaxe:
LGTM!
Now temporary tables are ignored at check_tree_needs_checkpointing as it's not in shared root info systree.

akorotkov (commented):
OK

PR #852, opened 2026-05-13
fix ALTER TABLE/INDEX aliasing

ereutt:
ALTER INDEX is in fact just an alias for the forms of ALTER TABLE that apply to indexes.

### DIFF
--- a/src/catalog/ddl.c
+++ b/src/catalog/ddl.c
@@ -957,7 +957,7 @@ orioledb_utility_command(PlannedStmt *pstmt,
 		if (OidIsValid(relid) && objtype == OBJECT_TABLE &&
 			(lockmode == AccessExclusiveLock || lockmode == ShareUpdateExclusiveLock))
 		{
-			Relation	rel = table_open(relid, lockmode);
+			Relation	rel = relation_open(relid, lockmode);
 
 			if (is_orioledb_rel(rel))
 			{
--- a/Makefile
+++ b/Makefile
@@ -99,6 +99,7 @@ OBJS = src/btree/btree.o \
 REGRESSCHECKS = btree_sys_check \
 				alter_type \
 				alter_storage \
+				alter_index \
 				bitmap_scan \
 				btree_compression \
 				btree_print \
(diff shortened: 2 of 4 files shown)


src/catalog/ddl.c
```diff
@@ -957,7 +957,7 @@ orioledb_utility_command(PlannedStmt *pstmt,
 		if (OidIsValid(relid) && objtype == OBJECT_TABLE &&
 			(lockmode == AccessExclusiveLock || lockmode == ShareUpdateExclusiveLock))
 		{
-			Relation	rel = table_open(relid, lockmode);
+			Relation	rel = relation_open(relid, lockmode);
 
 			if (is_orioledb_rel(rel))
```
homper, line 962: Is this expected behaviour that it just skips any index relation, because this check fails?
```C
bool
is_orioledb_rel(Relation rel)
{
	Assert(rel != NULL);

	return (rel->rd_tableam == (TableAmRoutine *) &orioledb_am_methods);
}
```
also, there is `objtype == OBJECT_TABLE` above
was test failing without this exact change?

ereutt: Actually looks like objtype == OBJECT_TABLE, when we come here with an index. Yes, test was failed without this changes right in table_open, on relkind validation, when I've done ALTER TABLE index_name SET TABLESPACE, instead of ALTER INDEX.

I will try another ALTER TABLE operation with index to answer about is_orioledb_rel check

ereutt: objtype depends on only on statement, not on relation kind. It has value OBJECT_TABLE for ALTER TABLE statement and OBJECT_INDEX for ALTER INDEX. So when ALTER TABLE index_name is executed we go into if branch and try to open index relation as table, so got an ERROR on verification.
As soon as we do not execute this if-branch for ALTER INDEX at all there is no mistake in negative check is_orioledb_rel for an index relation. Also there is no reason to do these extra checks inside  is_orioledb_rel for index, because it handle oriole specific  of alter statements doesn't applied to index at all.

Any way I've increase checks for ALTER TABLE/ALTER INDEX aliasing and move tests to separate sql script

ereutt:
Must be merged after https://github.com/orioledb/orioledb/pull/854

PR #860, opened 2026-05-21
Fix index-only scan crash on secondary indexes

fatih-akatlar:
OrioleDB secondary indexes internally store PK fields in leaf tuples, and scan.c adds matching PK columns to indextlist so the planner can use index-only scans.  However fill_itup() produced index tuples using the internal itupdesc (which includes hidden PK fields at different positions), causing a column count mismatch with the planner's slot. PostgreSQL's StoreIndexTuple() asserts slot->natts == itupdesc->natts, leading to a crash.

Reproducer:

```sql
CREATE TABLE t2 (a int4 NOT NULL, b int4 NOT NULL, c int4, d int4,
                 PRIMARY KEY(a, b) INCLUDE(c)) USING orioledb;
CREATE INDEX t2_d_idx ON t2 (d);
INSERT INTO t2 SELECT i, i*2, i*3, i*4 FROM generate_series(1, 200) AS i;
ANALYZE t2;

SET enable_seqscan = off;
EXPLAIN (COSTS OFF) SELECT d FROM t2 WHERE d > 500;
SELECT d FROM t2 WHERE d > 500 LIMIT 5;
```

Two fixes:

1. scan.c: Use nkeyfields instead of nfields when iterating PK columns to add to indextlist.  nfields includes INCLUDE columns of the PK, which should not be added since they are not part of the PK key.

2. handler.c fill_itup: For secondary indexes, build an extended index tuple matching the planner's indextlist (rd_att columns + extra PK fields) instead of using the internal itupdesc.

Assisted-by: Crush:glm-5.1

### DIFF
--- a/src/indexam/handler.c
+++ b/src/indexam/handler.c
@@ -1831,6 +1831,24 @@ fill_itup(IndexScanDesc scan, OTuple tuple, OTableDescr *descr,
 			temp_rowid_isnull[i] = slot->tts_isnull[attnum];
 		}
 
+		/*
+		 * primaryFieldsAttnums covers PK key columns only (see
+		 * add_index_fields(fillPrimary=true), nFields = nkeyfields), but
+		 * o_new_rowid() formats the rowid against primary->nonLeafTupdesc
+		 * whose natts is nkeyfields + nIncludedFields when the PK has INCLUDE
+		 * columns.  Those INCLUDE columns aren't part of the secondary's leaf
+		 * either, so we have no value to plug in -- and refind only matches
+		 * against the key columns anyway.  Fill the tail of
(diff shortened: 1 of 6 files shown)


akorotkov:
Hi, @fatih-akatlar. Thank you for your fix. I've edited it. This version manages to do without the allocation of a tuple descriptor per tuple. Please, check if it fixes all your cases.

fatih-akatlar:
> Hi, @fatih-akatlar. Thank you for your fix. I've edited it. This version manages to do without the allocation of a tuple descriptor per tuple. Please, check if it fixes all your cases.

Yes, it fixes the issues. Thank you.

homper (approved):

pashkinelfe (approved):

PR #862, opened 2026-05-22
Add pg_amcheck support function for orioledb tables

pashkinelfe:
- Add verify_orioledb sql function similar to orioledb_tbl_check but reporting to amcheck caller: one row per failed index
- Add wait_for_checkpoint mode for check_btree() mode that waits when called during checkpoint without spurious fail.
- Add tests for amcheck/pg_amcheck


src/btree/check.c
```diff
@@ -86,9 +85,21 @@ check_btree(BTreeDescr *desc, bool force_file_check)
...
+		 * Repeat until we get checkpoint number to avoid spurious failure due
+		 * to concurrent checkpoint when called by amcheck.
+		 */
+		while (!get_checkpoint_number(desc, desc->rootInfo.rootPageBlkno,
+									  &checkpoint_number, &copy_blkno))
+		{
+			CHECK_FOR_INTERRUPTS();
+			pg_usleep(1000L);
+		}
+	}
```
pgnickb, line 104: Nit: Any reason to not wait on a latch here?

sql/orioledb--1.7--1.8_prod.sql
```diff
@@ -2,3 +2,13 @@
 
 -- complain if script is sourced in psql, rather than via CREATE EXTENSION
 \echo Use "ALTER EXTENSION orioledb UPDATE TO '1.8'" to load this file. \quit
+
+CREATE FUNCTION verify_orioledb(relation regclass,
```
pgnickb, line 8: The name follows the pattern that `amcheck` uses but still can be misleading.

Do you think it's worth adding a comment or slightly adjusting the name to give a better idea on what the function does?

pgnickb (commented):
Looks good to me. I've left some minor comments.

sql/orioledb--1.7--1.8_prod.sql
pashkinelfe: Yes it follows amcheck name pattern. I'm not against renaming, what name do you propose?

pgnickb: I don't have any good ideas for a name. Maybe comment is a better way?

Also we should not forget to document this.

pashkinelfe: Done by 1d00e8aef49a36b4b78be4e1e75d4160ffd0bea6

src/btree/check.c
pashkinelfe: Done by 1d00e8aef49a36b4b78be4e1e75d4160ffd0bea6

src/tableam/func.c
```diff
@@ -1284,6 +1285,56 @@ orioledb_tbl_check(PG_FUNCTION_ARGS)
...
 
+/*
+ * amcheck entry point: verify all b-trees of the orioledb relation and emit
+ * one row per failed index as (index_name, msg).  `thorough_check` forces
+ * the on-disk map check inside check_btree.  Wait-for-checkpoint mode is
+ * always used so amcheck doesn't spuriously fail on a tree sitting on a
+ * checkpoint boundary.
+ */
+Datum
+verify_orioledb(PG_FUNCTION_ARGS)
```
za-arthur, line 1296: This function is mostly copy of `orioledb_tbl_check()`, they have difference in return results.
I propose to delete `orioledb_tbl_check()` in favor of this new function.

pashkinelfe: I'm planning to do this in a separate PR for not mixing pg_amcheck changes with tests-related refactoring.

za-arthur (approved):

src/tableam/func.c
pashkinelfe: More precisely I'm planning to switch to verify_orioledb from orioledb_tbl_check in all callers/tests. orioledb_tbl_check function itself should remain in C code and in sql for compatibility (I can't remember precisely breakup scenario for deleted functions, however). Do you agree @za-arthur ?

za-arthur: @pashkinelfe I think we can just delete it all together. I think it can break explicitly granted permissions, but it is manageable.

src/btree/check.c
akorotkov: Should we `o_tables_rel_lock_extended(&oids, AccessExclusiveLock, true)` instead?  Even if we waited for concurrent no checkpointing at some moment, there is no guarantee new one didn't started.

PR #864, opened 2026-05-25
Deprecate orioledb_tbl_check() and switch all callers to

pashkinelfe:
verify_orioledb()

Autonomous page writes during a checkpoint leave per-page state that only clears on the next explicit CHECKPOINT, so the checkpoint_split_root* tests now run that follow-up CHECKPOINT before asserting a clean verify_orioledb result (matching the concurrent_start/midle/end pattern in the same file).

orioledb_tbl_check itself is left in a deprecated but unchanged state:
- C function should be defined in .so in case it's referenced from previous SQL version
- SQL function is left for not breaking restoring previous logical backups that could reference this SQL function (e.g. some permission GRANT's)


src/tableam/func.c
```diff
@@ -1249,41 +1249,15 @@ orioledb_tbl_are_indices_equal(PG_FUNCTION_ARGS)
...
-	{
-		o_btree_load_shmem(&descr->indices[i]->desc);
-		result = check_btree(&descr->indices[i]->desc, force_map_check);
-
-		if (result == false)
-			break;
-	}
-	relation_close(rel, AccessExclusiveLock);
+	ereport(ERROR,
+			(errcode(ERRCODE_WRONG_OBJECT_TYPE),
```
za-arthur: I think `ERRCODE_FEATURE_NOT_SUPPORTED` would make more sense here:
```suggestion
			(errcode(ERRCODE_FEATURE_NOT_SUPPORTED),
```

test/t/checkpoint_split3_test.py
```diff
@@ -288,10 +286,12 @@ def test_checkpoint_split_root(self):
 
 		self.assertEqual(
 		    node.execute("SELECT COUNT(*) FROM o_checkpoint;")[0][0], 28)
-		# autonomous checkpoint write happens
-		self.assertFalse(
-		    node.execute("SELECT orioledb_tbl_check('o_checkpoint'::regclass)")
-		    [0][0])
+		node.execute("SELECT * FROM verify_orioledb('o_checkpoint'::regclass)")
```
za-arthur: What should `verify_orioledb()` return here? Should we check its output?

test/t/checkpoint_split3_test.py
```diff
@@ -343,15 +343,13 @@ def test_checkpoint_split_root_v2(self):
 
 		self.assertEqual(
 		    node.execute("SELECT COUNT(*) FROM o_checkpoint;")[0][0], 29)
-		# autonomous checkpoint page write happens
-		self.assertFalse(
-		    node.execute("SELECT orioledb_tbl_check('o_checkpoint'::regclass)")
-		    [0][0])
+		node.execute("SELECT * FROM verify_orioledb('o_checkpoint'::regclass)")
```
za-arthur: It seems we don't check the result of `verify_orioledb` here too.

za-arthur (commented):
It looks good to me, I have a couple of comments below.

test/t/checkpoint_split3_test.py
pashkinelfe: First call of orioledb_tbl_check returned transient false during checkpointing. verify_orioledb doesn't output false it waits istead and returns actual check result. The the first check remains that there is no errors.

Final check follows unchanged:
# no incomplete split after the second checkpoint
               self.assertEqual(
                   node.execute(
                       "SELECT * FROM verify_orioledb('o_checkpoint'::regclass)"), [])

test/t/checkpoint_split3_test.py
pashkinelfe: Same as previous

test/t/checkpoint_split3_test.py
pashkinelfe: @za-arthur we can check clean status in both calls, I think it's not needed as only second result is diagnostic. (First is transient, but is also clean)

za-arthur: @pashkinelfe why is `verify_orioledb()` is called if its result isn't checked in the test? Does it have any side effect to verify in the test?

AFAIU `verify_orioledb()` is a function to check an orioledb table consistency which in the end returns the result of the check. If we call it and don't check its result, why do we call it?

za-arthur: Would it make sense to delete this line if its result isn't checked?

pashkinelfe:
I'm closing this for now because orioledb-tbl-check and verify_orioledb has diffrrent attitude to transient errors: verify_orioledb considers them "clean", orioledb-tbl-check is used to detect them which is used in  tests (see 0ea982e0f46ec9b29 )

PR #870, opened 2026-05-28
Ci/nightly build publish pipeline

dliappis:
## What kind of change does this PR introduce?

CI feature — additional automated nightly build & publish pipeline

## What is the current behavior?

Builds and publishes are entirely manual: a maintainer cuts a `betaN` git tag, publishes a GitHub Release, and `docker.yml` reacts to the release event by pushing `latest-pg{16,17}{-ubuntu}` to Docker Hub. Between releases there is no automated way to consume a recent build, so anyone tracking `main` has to either build locally or wait for the next `betaN`.

## What is the new behavior?

New workflow `.github/workflows/nightly.yml`:

- Triggered by daily cron at 02:00 UTC; cron is guarded against forks via `github.repository == 'orioledb/orioledb'`.
- Builds the same matrix as `docker.yml`
- Emits two Docker tags per (PG-major, distro) combination on `orioledb/orioledb`:
  - **Immutable** — `pg<N>-nightly-<YYYYMMDD>-<sha7>` (plus `-ubuntu` variant)
  - **Rolling** — `pg<N>-nightly` (plus `-ubuntu` variant)
- After all images are in place, creates a GitHub Release named `nightly-<YYYY-MM-DD>-<sha7>` marked `prerelease: true`.  The `<sha7>` suffix means two builds on the same UTC day (different commits) never collide on a tag; a retry on the same commit is idempotent because the resolved tag is identical and the release-creation step short-circuits when the release already exists.
- Release creation runs *after* the manifest job, so the git tag never points at a build that failed to publish.

Defensive guard in `.github/workflows/docker.yml`: the `build` job now skips when the triggering release tag starts with `nightly-`, so a manually-published nightly prerelease (or one created with a PAT) can never accidentally clobber `latest-pg{16,17}`.

## Additional context

- Follow-up: gate nightly publish on a shared `workflow_call` build matrix so a failing build cannot produce a published nightly tag. Out of scope here.

- **On fixed short sha:** sliced as `${GITHUB_SHA:0:7}` rather than via `git rev-parse --short`. Adaptive abbreviation would require adding an `actions/checkout` step which adds several seconds of unnecessary CI delay just to avoid collisions that only become a practical concern at hundreds of thousands to millions of commits (the Linux kernel hit it around 2016 at ~500K commits and git made abbreviation adaptive in response); orioledb is many orders of magnitude below that threshold, and a collision would have to land on main's head on two different UTC days to even surface. Finally, fixed 7 chars matches the GitHub UI display convention and the default sha-length in docker/metadata-action.

### DIFF
--- a/.github/workflows/nightly.yml
+++ b/.github/workflows/nightly.yml
@@ -1,0 +1,87 @@
+name: nightly
+
+on:
+  # Cron disabled until the first manual workflow_dispatch from `main` is
+  # validated. Re-enable in a follow-up PR once the publish path (docker push +
+  # GitHub Release) has been confirmed end-to-end.
+  # schedule:
+  #   - cron: '0 2 * * *'  # 02:00 UTC daily
+  workflow_dispatch:
+
+concurrency:
+  group: nightly
+  cancel-in-progress: false
+
+jobs:
+  prepare:
+    # Skip cron runs in forks; allow manual dispatch anywhere for testing.
+    if: github.event_name == 'workflow_dispatch' || github.repository == 'orioledb/orioledb'
+    runs-on: ubuntu-24.04
+    outputs:
+      git_tag: ${{ steps.date.outputs.git_tag }}
+      build_id: ${{ steps.date.outputs.build_id }}
+    steps:
+      - id: date
+        run: |
+          date_iso=$(date -u +%Y-%m-%d)
+          date_compact=$(date -u +%Y%m%d)
+          sha_short=${GITHUB_SHA:0:7}
+          {
+            # build_id (compact date + 7-char SHA) makes every build uniquely
+            # taggable even if cron and a manual dispatch fire on the same UTC
+            # day from different commits.
+            echo "build_id=$date_compact-$sha_short"
+            echo "git_tag=nightly-$date_iso-$sha_short"
+          } >> "$GITHUB_OUTPUT"
+
+  build-publish:
+    needs: prepare
+    uses: ./.github/workflows/_docker-build-publish.yml
+    with:
+      mode: nightly
+      nightly_build_id: ${{ needs.prepare.outputs.build_id }}
+    secrets: inherit
+
+  # Publish the GitHub Release (and its git tag) only after Docker images are
+  # in place, so the tag never points at a build that failed to publish.
+  release:
+    needs: [prepare, build-publish]
+    runs-on: ubuntu-24.04
+    permissions:
+      contents: write
+    steps:
+      - name: Create GitHub Release (prerelease)
+        env:
+          GH_TOKEN: ${{ github.token }}
+          GH_REPO: ${{ github.repository }}
+          TAG: ${{ needs.prepare.outputs.git_tag }}
+        run: |
+          set -euo pipefail
+
+          # Nightly tags are immutable ("no moving git tag"): if the day's
+          # release already exists, treat as success rather than re-tagging.
+          if gh release view "$TAG" >/dev/null 2>&1; then
+            echo "Release $TAG already exists; skipping."
+            exit 0
+          fi
+
+          notes_file="$RUNNER_TEMP/release_body.md"
+          cat > "$notes_file" <<'EOF'
+          Automated nightly build of OrioleDB from `main` at commit `${{ github.sha }}`.
+
+          Docker images published to [`orioledb/orioledb`](https://hub.docker.com/r/orioledb/orioledb):
+
+          **Immutable (this build, ${{ needs.prepare.outputs.build_id }}):**
+          - `pg16-nightly-${{ needs.prepare.outputs.build_id }}` · `pg16-nightly-${{ needs.prepare.outputs.build_id }}-ubuntu`
+          - `pg17-nightly-${{ needs.prepare.outputs.build_id }}` · `pg17-nightly-${{ needs.prepare.outputs.build_id }}-ubuntu`
+
+          **Rolling (latest nightly):**
+          - `pg16-nightly` · `pg16-nightly-ubuntu`
+          - `pg17-nightly` · `pg17-nightly-ubuntu`
+          EOF
+
+          gh release create "$TAG" \
+            --prerelease \
+            --target "${{ github.sha }}" \
+            --title "$TAG" \
+            --notes-file "$notes_file"
--- a/.github/workflows/_docker-build-publish.yml
+++ b/.github/workflows/_docker-build-publish.yml
@@ -1,0 +1,153 @@
+name: _docker-build-publish
+
+# Reusable workflow invoked by docker.yml and nightly.yml.
+
+on:
+  workflow_call:
+    inputs:
+      mode:
+        description: '`release` | `nightly`. Drives the metadata-action flavor/tags shape.'
+        required: true
+        type: string
+      release_extra_tag:
+        description: 'Optional raw tag (release mode only). Ports docker.yml workflow_dispatch `tag` input.'
+        required: false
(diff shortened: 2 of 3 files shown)


dliappis:
## For reviewers:

To assist with the PR review I kicked off a test run of the new action by [enabling workflow trigger](https://github.com/orioledb/orioledb/pull/870/commits/1505fcea2c05d02613f096963723d2fa5301a6b2) but having the actual steps that [push new docker images and github release+tag commented out](https://github.com/orioledb/orioledb/pull/870/commits/b30771ba26f4f391a5540fe711ba3f8b32459fec).

The output of the action can be found here:

https://github.com/orioledb/orioledb/actions/runs/26566202610

.github/workflows/nightly.yml
```diff
@@ -0,0 +1,209 @@
...
+          - `pg16-nightly-${{ needs.prepare.outputs.build_id }}` · `pg16-nightly-${{ needs.prepare.outputs.build_id }}-ubuntu`
+          - `pg17-nightly-${{ needs.prepare.outputs.build_id }}` · `pg17-nightly-${{ needs.prepare.outputs.build_id }}-ubuntu`
+
+          **Rolling (latest nightly):**
+          - `pg16-nightly` · `pg16-nightly-ubuntu`
+          - `pg17-nightly` · `pg17-nightly-ubuntu`
+          EOF
+
+          gh release create "$TAG" \
+            --prerelease \
```
dliappis, line 84: I think this is necessary for a "rolling release" ; e.g. prevents users/tooling following `/releases/latest` from accidentally picking up a nightly/prerelease.

pashkinelfe (commented):
Could we re-use parts of the scripts for release docker builds in nightly builds without duplication?

.github/workflows/nightly.yml
```diff
@@ -0,0 +1,209 @@
...
+            distr_version: 3.21
+          - cpu: arm64
+            distr: ubuntu
+            runner: ubuntu-24.04-arm
+            distr_version: noble
+
+    runs-on: ${{ matrix.runner }}
+
+    steps:
+      - uses: actions/checkout@de0fac2e4500dabe0009e67214ff5f5447ce83dd # v6.0.2
```
pashkinelfe: This changed from docker.yml 
Any reason for this difference?

dliappis:
> Could we re-use parts of the scripts for release docker builds in nightly builds without duplication?

Yes sure!

The reason for this approach was mostly out of abundance caution: to reduce the blast radius from a bug that could break/rewrite existing beta releases while testing nightlies. Secondarily I felt the duplicate steps are mostly scaffolds that rarely change.

For your ask, to re-use existing docker build steps, one idea is to move much of them out to a new reusable workflow that accepts a "mode" parameter that both docker.yml and nightly.yml would use use, e.g.:

```yaml
name: reusable-docker-build-and-push  # new reusable workflow

on:
  workflow_call:
    inputs:
      mode:
        required: true
        type: string         # "release" | "nightly"
      release_extra_tag:
        required: false
        type: string
        default: ""          # ports the current docker.yml workflow_dispatch `tag` input
      nightly_build_id:
        required: false
        type: string
        default: ""          # <YYYYMMDD-sha7> from nightly.yml's prepare job
...
```

with some conditionals to satisfy the need for extra suffixes for nightlies depending on mode nightly/release
The existing `docker.yml` would be modified (shortened) to reuse this, and similarly for `nightly.yml`.

With all that said, would you rather I do this refactoring as part of this PR, or land it first and then refactor in a subsequent one?

.github/workflows/nightly.yml
```diff
@@ -0,0 +1,209 @@
...
+on:
+  # Cron disabled until the first manual workflow_dispatch from `main` is
+  # validated. Re-enable in a follow-up PR once the publish path (docker push +
+  # GitHub Release) has been confirmed end-to-end.
+  # schedule:
+  #   - cron: '0 2 * * *'  # 02:00 UTC daily
+  workflow_dispatch:
+
+permissions:
+  contents: write  # to create the GitHub Release and its git tag
```
za-arthur: It might be just a nitpicking, but I think we don't need to define `permissions` here in a broader context since it is defined specifically in the `release` job below anyway, where the workflow creates tag.

dliappis: Great catch, addressed in baee7661f0fbb4d7db5147484e0b3b7165e2355c

It was a left over from a previous structure that needed it at that point.

pashkinelfe (commented):
> With all that said, would you rather I do this refactoring as part of this PR, or land it first and then refactor in a subsequent one?

I think it's better to re-use existing docker build code in this PR from the start (and do necessary refactoring) The main goal is not risking occasional difference between existing "latest" and upcoming "nightly" build workflows

.github/workflows/nightly.yml
za-arthur: Using a hash rather than a tag is a security practice to mitigate supply chain attacks.

dliappis: Right, and we can enable automation like dependabot (if not already present) to help us keep those hashes up to date. For reference: https://docs.github.com/en/actions/reference/security/secure-use#using-third-party-actions

dliappis:
> I think it's better to re-use existing docker build code in this PR from the start (and do necessary refactoring) The main goal is not risking occasional difference between existing "latest" and upcoming "nightly" build workflows

@pashkinelfe Ok I pushed https://github.com/orioledb/orioledb/pull/870/commits/3302da18 deduplicating and refactoring as discussed. A few notes:

1. Common build / manifest steps from `docker.yml` and `nightly.yml` now live in a single **reusable workflow** at `.github/workflows/_docker-build-publish.yml`, invoked by both callers via `workflow_call`. As a result `docker.yml` and `nightly.yml` as mostly consumers.

2. We are introducing a new reusable workflow condition on `inputs.mode` (`release` | `nightly`) to emit the right tag set per caller — for releases that still resolves to `<tag>-pg<N>[-ubuntu]` + `latest-pg<N>[-ubuntu]`, identical to today.

3. I kept the pined action SHA being a best practice (it also addresses warnings about node 20 getting removed from GHA)

## Testing

Same earlier strategy with temporary commits (now reverted): used two temp commits to disable push/manifest/release inside the reusable workflow (https://github.com/orioledb/orioledb/pull/870/commits/2ed86d81) and add a push trigger to nightly (https://github.com/orioledb/orioledb/pull/870/commits/8fe1d6d7) so I could validate the build matrix without producing any external artifacts. Result: https://github.com/orioledb/orioledb/actions/runs/26635849353?pr=870

As mentioned earlier, since this refactoring also touches `docker.yml` and I haven't end-to-end tested the publish/manifest/release path (push events don't trigger docker.yml, and a `release: published` would create real public Docker Hub + GitHub Release artifacts), we should review carefully and discuss scenarios where things might break with the existing docker workflow.

## Additional Note

I opted for a reusable workflow (rather than composite actions under `.github/actions/`) because it allows  `strategy.matrix` — composite actions are step-level and can't define a `matrix:`, so that route would have needed a separate data file plus a matrix-reader action. (I attempted the composite-action approach first and looked rather bloated)

za-arthur (approved):

dliappis:
@pashkinelfe any thoughts on the [latest refactoring](https://github.com/orioledb/orioledb/pull/870#issuecomment-4575339138)?

pashkinelfe (approved):

danielmitterdorfer (approved):
LGTM, thanks!

dliappis:
Thanks everyone for the reviews! Merging.

PR #879, opened 2026-06-02
CREATE INDEX CONCURRENTLY for orioledb (v1: non-unique)

akorotkov:
CREATE INDEX CONCURRENTLY for orioledb, plus REINDEX CONCURRENTLY for native btree indexes.

**History reworked for review.** The original 19-commit history is preserved unchanged at `create_index_concurrently-orig-20260817`; before the rebase the reworked history was verified to produce a byte-identical tree (`d102ed68`).

Ten commits now, design first:

1. **doc: architecture page** — read this first; it explains the phases, the spool, and the lock dance.
2. **Index state, spool module and WAL records for phase transitions** — the on-disk and catalog groundwork.
3. **Capture concurrent DML into the spool** — the write-side hooks, including the REVERSE-op undo hook.
4. **Phase orchestration via ambuild and validate_scan** — the actual CIC driver.
5. **Survive and report parallel-build failures on a replica** — the three recovery-side fixes, together.
6. **Support UNIQUE native indexes and bridged indexes**.
7. **REINDEX CONCURRENTLY** — native btree, primary index skipped in TABLE form.
8. **Stop events at phase boundaries** — for deterministic concurrency tests.
9. **Adapt to main and regenerate expected outputs**.
10. **Address review comments from @homper** — kept separate so the re-review is just this commit.

**Rebased onto current main.** The one conflict worth knowing about: main took WAL record ids 20 and 21 for `DATABASE_CREATE_COPY` / `DATABASE_TEMPLATE_CHECKPOINT`, so the three CIC phase records were renumbered to **22, 23, 24**.

**Review comments addressed** (commit 10):

- `OIndexState` spoke of a "capture phase" and "meta installed" — neither names anything in the code. It now says what is captured and what each state means for writers.
- `tableam.sql` had lost its `-- not supported` marker when CIC became supported, leaving the still-failing statements unmarked. Marker restored in front of them.
- The reserved `BUILDING_PHASE_1/3/4` states now say *why* they exist now (the state is persisted, so a value added later would not be readable by an older binary) and what each would be for.
- The `REINDEX CONCURRENTLY` primary-index error is documented, with the reason the primary index cannot be rebuilt that way.

Local: build clean, regresscheck 49/49.


doc/architecture/concurrent-index.mdx
```diff
@@ -0,0 +1,297 @@
...
+  spool drain produces a complete index.
+- `test_cic_writer_spans_all_phases`: writer commits tiny transactions
+  spanning all phases; final `pk_cnt` must equal `idx_cnt`.
+- `test_cic_streaming_replica`: standby converges after CIC on master.
+- `test_cic_streaming_replica_failing_expr`: CIC whose expression
+  errors on some row leaves the replica caught up rather than stalled.
+- `test_cic_crash_orphan_cleanup`: planted `cic_*/spool_*.bin` is
+  removed by `rm_cleanup` on SIGKILL-induced WAL recovery.
+
+## Limitations
```
homper, line 301: "ERROR:  cannot REINDEX CONCURRENTLY primary index \"[a-z0-9_]+\" on an orioledb table" probably should be mentioned here. This is a massive doc to read by eyes for humans, but still, and it also could be parsed by llms, since we directing them to docs.

test/sql/tableam.sql
```diff
@@ -24,7 +24,6 @@ CREATE TABLE o_tableam1
 	value text
 ) USING orioledb;
 
--- not supported
```
homper, line 27: Probably should better just move `CREATE INDEX CONCURRENTLY o_tableam1_ix_concurrently ON o_tableam1 (key);` below, since this block still contains unsupported features, or comment -- not supported should be put in front of SERIALIZABLE transactions

doc/architecture/concurrent-index.mdx
```diff
@@ -0,0 +1,297 @@
...
+
+The fix: when cached state is non-VALID, re-read the OIndex state from
+the sys-tree (a cheap in-progress snapshot lookup) before falling into
+the spool branch.  If the fresh state is VALID, fall through to the
+normal index write.
+
+## Standby replay
+
+CIC emits three WAL phase records for observability
+(`WAL_REC_CIC_PHASE_3_START`, `WAL_REC_CIC_PHASE_4`,
```
homper: Can it be described here why we need all those reserved phases? Why could they be used in the future. Ok, mentioned in enum above probably

include/catalog/o_indices.h
```diff
@@ -19,6 +19,25 @@
...
+ * Lifecycle state of an index.  Non-VALID states are used by
+ * CREATE INDEX CONCURRENTLY to expose the index to writers (capture
+ * phase) before it becomes available to readers.  Persisted as one
+ * trailing byte in serialize_o_index(); records written by versions
+ * predating this field are read as OINDEX_STATE_VALID.
+ */
+typedef enum
+{
+	OINDEX_STATE_VALID = 0,		/* fully built, available to all readers */
+	OINDEX_STATE_BUILDING_PHASE_1,	/* meta installed, capture started, no DML
```
homper: capture of what, since it is not used anywhere in the code?  does it mean meta lock by word `meta`?

ci/filter_regression_diff.py
homper, line 1: Btw, do we need to enable some of tests filtered out in isolation after implementing index or should it be separate pr in your opinion?
https://github.com/orioledb/orioledb/blob/534f3ff8d5c2636b6f206a3f91ecccc22c5d3533/ci/filter_isolation_diff.py#L33

doc/architecture/concurrent-index.mdx
```diff
@@ -0,0 +1,318 @@
...
+
+## Index lifecycle: `OIndexState`
+
+A new column in the `OIndex` sys-tree record carries the index's lifecycle
+state:
+
+```c
+typedef enum
+{
+    OINDEX_STATE_VALID = 0,         /* fully built, all readers/writers OK */
```
homper, line 58: Probably enum and comments here should be updated. https://github.com/orioledb/orioledb/pull/879/changes#diff-abfd87e6c0f099ab34ef7f99009680d94a7db35ba57846bf73cd6915c2942b59R36
Btw, what's catchup mode in this context? Ok, it describes idea in paragraph below, probaly could be mentioned in comment, or it should briefly mention that info is next paragraph.
And also, stupid question, why do we have WAL_REC_CIC_PHASE_3_START in wal but don't have WAL_REC_CIC_PHASE_3_END. Shouldn't it be just a 'PHASE_3' then? or something more descriptive, like 'TURN_ON' or something. Could all those enums values be more descriptive? Or would names be picked in the future also?

include/catalog/o_indices.h
```diff
@@ -19,6 +19,26 @@
...
+ * Lifecycle state of an index.  A non-VALID state means CREATE INDEX
+ * CONCURRENTLY is still building it: the index row is already visible to
+ * writers, which record their changes into the CIC spool, but not yet to
+ * readers.  Persisted as one trailing byte in serialize_o_index(); a record
+ * written before this field existed reads as OINDEX_STATE_VALID.
+ */
+typedef enum
+{
+	OINDEX_STATE_VALID = 0,		/* fully built, available to all readers */
+	OINDEX_STATE_BUILDING_PHASE_1,	/* the OIndex row exists and writers spool
```
homper: Please mention that is never used at the moment briefly at the begining of comment. PHASE 3 and 4 also

include/catalog/o_indices.h
```diff
@@ -19,6 +19,26 @@
...
+									 * their changes; nothing is written to
+									 * the index itself */
+	OINDEX_STATE_BUILDING_PHASE_2,	/* phase 1, plus the snapshot scan that
```
homper: I think "build pending, writers capture to spool" is more descriptive for this index state.

homper (commented):
Most of comments could look like nitpicking and only correcting docs, but it was too confusing (at least for me) to figure out how this works and is it working correctly at the moment and not in the future.

Otherwise, no complains apart from slight complexity. But this is concurrent algorithm after all. (Maybe observability wal records could be not added right now, if they not used, but this breaks nothing, and is needed for future extensions I assume)

doc/architecture/concurrent-index.mdx
akorotkov: Done, and I fixed something the text was hiding: this section claimed CIC emits three WAL phase records. It emits one — `WAL_REC_CIC_INDEX_VALID`, when the index becomes visible to readers. The other two are never written; like the unused `OIndexState` values they only reserve numbering a later revision would need, since a record number added later would not be understood by an older binary. The section now says that and points back to the enum section, and the enum section itself explains the reservation in one place instead of two half-places.

include/catalog/o_indices.h
akorotkov: Right — the comment described what a future catchup mode would do, in the present tense, so "capture" had no antecedent. Rewritten: the enum header now defines both terms where they first appear — a writer **captures** a change when it appends it to the CIC spool instead of writing it into the index, and **catchup mode** is the protocol a later revision would need to build against a moving snapshot rather than a fixed one. Nothing in the block relies on "meta" any more either.

include/catalog/o_indices.h
akorotkov: Done for all three. Each unused value now leads with it:

```c
OINDEX_STATE_BUILDING_PHASE_1,  /* not reached in this revision: reserved
                                 * for catchup mode, where capture opens
                                 * before the scan starts */
```

and the same for `PHASE_3` and `PHASE_4`. The enum header states once that only `VALID` and `BUILDING_PHASE_2` are reached, and why the others exist at all: the state is persisted, so a value added later would not be understood by an older binary, whereas an unused value that already exists costs nothing and keeps the numbering stable.

include/catalog/o_indices.h
akorotkov: Taken verbatim — `PHASE_2` now reads `/* build pending, writers capture to spool */`, and the header and the architecture doc no longer disagree about what these values mean.

doc/architecture/concurrent-index.mdx
akorotkov: Three separate answers, because the question turned out to have three parts.

**The enum comments.** You were right that the header was the stale copy — it described catchup-mode semantics in the present tense while the doc already said "reserved". The header now matches the doc, each unused value leads with "not reached in this revision", and both terms are defined at first use.

**"What's catchup mode?"** Now stated where the enum is introduced instead of being left to the paragraph below: a build that runs against a moving snapshot rather than a fixed one, which is why it needs to hand writers over to the index while the scan is still running.

**`PHASE_3_START` with no `PHASE_3_END`.** The naming looked asymmetric because it was named after a protocol that does not exist here yet. The fact behind it: **only one of the three records is ever emitted** — the single `add_cic_phase_wal_record()` call site writes the flip to VALID. The other two are reserved record numbers. So I renamed them for what they mean, keeping the numbers:

| was | is | emitted? |
|---|---|---|
| `WAL_REC_CIC_PHASE_3_START` (22) | `WAL_REC_CIC_WRITERS_DIRECT` | no — reserved |
| `WAL_REC_CIC_PHASE_4` (23) | `WAL_REC_CIC_DRAIN_BARRIER` | no — reserved |
| `WAL_REC_CIC_PHASE_FLIP` (24) | `WAL_REC_CIC_INDEX_VALID` | yes |

and both the X-macro list and `add_cic_phase_wal_record()` now say which is which. Nothing in `test/` referenced the old identifiers or display strings.

ci/filter_regression_diff.py
akorotkov: Measured it rather than guessed, and the answer has a twist.

**The two entries are dead, so they are gone in this PR.** Neither `ERROR:  concurrent index creation is not supported for orioledb tables yet` nor `ERROR:  REINDEX CONCURRENTLY is not supported for orioledb tables yet` exists in the source any more, so those regexes cannot match a line. Removed from `filter_isolation_diff.py`, and the same dead REINDEX entry from `filter_regression_diff.py`.

**What actually happens to those specs now.** I ran the five upstream isolation specs that use CONCURRENTLY against orioledb (PG18, `default_table_access_method = orioledb`, `orioledb.strict_mode = on`):

| spec | run | filter output with the entries removed |
|---|---|---|
| `multiple-cic` | **passes outright** | — |
| `prepared-transactions-cic` | differs | empty |
| `reindex-concurrently` | differs | empty |
| `reindex-concurrently-toast` | differs | empty |
| `reindex-schema` | differs | empty |

So nothing new surfaces. Two things explain the "empty" column, and both are worth knowing:

* `prepared-transactions-cic` is absorbed by the existing `cannot use PREPARE TRANSACTION in transaction that uses orioledb table` entry.
* The three reindex specs differ only by an added `s3: NOTICE:  skipping reindex of index "public.reind_con_tab_pkey" concurrently` and by `step …` lines, and the script skips added `step …` and `sN: NOTICE|WARNING|DETAIL:` lines by construction. Our new PK-skip notice is invisible to it either way.

**On "should we enable some of the filtered-out tests".** There is nothing to enable: in the `pg_tests` cell `make -C src/test/isolation installcheck` runs with `|| true` and the filtered diff never sets `status` (`ci/check.sh`), so the file is an artifact for humans, not a gate. The filters do not disable tests — they decide what shows up in that artifact.

**One thing I did not touch**, so it does not get lost: the `create_index` block in `filter_regression_diff.py` still carries the comment *"some errors related to not working CONCURRENT index build right now"*, which is no longer true. Whether its three regexes are still needed I could not establish locally — `filter_regression_diff.py` itself dies with `IndexError` in `query_plan_to_tree()` on the diff of a single-test run, so it needs the CI-shaped full-schedule run to answer. Happy to take it as a follow-up.

Local runs on this branch: `regresscheck` 49/49, `isolationcheck` 35/35, and the four CIC/REINDEX testgres files 5 + 5 + 11 + 6, all green.

PR #880, opened 2026-06-03
Fix duplicate seq scan results from overlapping undo leaves

GremSnoort:
## Summary

This PR fixes a B-tree sequential scan bug where multiple adjacent current leaf pages can be read through undo and expose overlapping historical key ranges. In that case the scan may return the same logical tuples more than once.

The original symptom was observed during targeted recovery, but the new regression test shows the same invariant can also be reached in regular live execution through the waiter split path.

## Problem

A seq scan walks the current internal-page topology. If a current child leaf page has `page_csn >= imgReadCsn`, the scan reads a historical image of that leaf through undo.

When several adjacent current child leaves are read through undo, their historical images may correspond to an older tree shape where those key ranges were not yet split. As a result, different current children can unfold into overlapping historical leaf ranges.

Before the fix, the scan treated those historical pages as independent ranges and could emit duplicate logical tuples.

## Test Coverage

Adds a live multi-backend repro:

- `test_live_waiter_split_seq_scan_reads_adjacent_undo_leaves`

The test forces a waiter split path, then runs a concurrent `SELECT` and verifies that the result contains no duplicate keys.

### DIFF
--- a/include/btree/page_contents.h
+++ b/include/btree/page_contents.h
@@ -363,7 +363,8 @@ extern ReadPageResult o_btree_try_read_page(BTreeDescr *desc,
 											Pointer key, BTreeKeyType keyType,
 											PartialPageState *partial,
 											bool loadHikeysChunk,
-											CommitSeqNo *readCsn);
+											CommitSeqNo *readCsn,
+											bool *readFromUndo);
 extern UndoLocation read_page_from_undo(BTreeDescr *desc, Page img,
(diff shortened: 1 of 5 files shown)


pashkinelfe:
@GremSnoort could you look at the alternative approach to fix the problem? https://github.com/orioledb/orioledb/pull/884

pashkinelfe:
Fixed by #891

PR #881, opened 2026-06-03
Wrap pure oriole tx commit flow into CRIT_SECTION

### LINKED ISSUE #869: Primary-replica desync due to different ROLLBACK mechanic
## Summary 
Primary and replica clusters can have assymetric machinery of tx rollback within commit flow. Raising ereport(ERROR, ) within such window leads to proper primary node rollback, but the replica node stays with tx partially commited. Hence, primary and replica divergent, furthermore, replicate ends up in an inconsistent state.

## The intuition
In case of pure oriole transaction, we take [that o-o tx branch](https://github.com/orioledb/orioledb/blob/d7c8847e67667588de9b9b230b460f7fa44f6e38/src/transam/undo.c#L2179-L2201) inside `undo_xact_callback`.

```C
if (!TransactionIdIsValid(heapXid))
{
    bool		wrote_xlog;

	/* Commit o - o : independent Oriole transaction */

	flushPos = assign_xidless_commit_lsn(oxid, &wrote_xlog);
...
```

The window opens at `add_finish_wal_record(WAL_REC_COMMIT, ...)` inside `wal_commit` (where the COMMIT bytes first enter any buffer) and closes at `curOxid = InvalidOXid` inside `current_oxid_commit` (where the abort callback's early-return becomes effective).
The root cause is [`local_wal_has_material_changes = false;`](https://github.com/orioledb/orioledb/blob/d7c8847e67667588de9b9b230b460f7fa44f6e38/src/recovery/wal.c#L297) line, because it introduce the [fast-path wal_rollback](https://github.com/orioledb/orioledb/blob/d7c8847e67667588de9b9b230b460f7fa44f6e38/src/recovery/wal.c#L342-L348).

```C
if (!local_wal_has_material_changes)
{
	local_wal_buffer_offset = 0;
	local_type = oIndexInvalid;
	ORelOidsSetInvalid(local_oids);
	return;
}
```

Hence, the primary node can properly rollback (due to in-memory undo state), while replica node doesn't receive any rollback wal-record at all.

Possible error window closee only at `current_oxid_commit` function call, when [`curOxid = InvalidOXid;`](https://github.com/orioledb/orioledb/blob/d7c8847e67667588de9b9b230b460f7fa44f6e38/src/transam/oxid.c#L1478), the occurrence of an error here leads to the same state of the primary node and the replica node, but not to the correct state, since both the replica and the primary node choose a fast rollback.

### Bug window schema
Here is a summary across multiple places where errors were injected. It is possible that there are multiple errors that can be encountered during the injection to the same location, but here is a brief overview.

```
[entry: undo_xact_callback XACT_EVENT_COMMIT case body opens]
                                                ┌── SAFE before this point ──
                                                │
wal_commit() (src/recovery/wal.c):              │
    add_finish_wal_record(WAL_REC_COMMIT, ...)  ▼ ◄── WINDOW OPENS
                                                ▲
                                                │ ◄── Malformed WAL sequence
                                                │
    flush_local_wal()                           │
        log_logical_wal_container() submits to PG shared XLog
                                                │ ◄── Replica undo panic
                                                │
    [flush_local_wal returns]                   │
                                                │ ◄── Replica undo panic
                                                │
    local_wal_has_material_changes = false      │
                                                │ ◄── Primary-Replica desync from here onward
                                                │
[wal_commit returns]                            │
                                                │
set_oxid_xlog_ptr(oxid, xidless_commit_lsn)     │ Primary-Replica desync
[assign_xidless_commit_lsn returns]             │
XLogFlush(flushPos)                             │ Primary-Replica desync
SyncRepWaitForLSN(flushPos, true)               │ Primary-Replica desync
                                                │
current_oxid_precommit()                        │ Primary-Replica desync
    → set_oxid_csn(oxid, COMMITTING_SENTINEL)   │
                                                │
csn = pg_atomic_fetch_add(&nextCommitSeqNo, 1)  │ Primary-Replica desync
                                                │
current_oxid_commit(csn) (src/transam/oxid.c):  │
    set_oxid_csn(curOxid, real_csn | ...)       │ Primary-Replica desync
    pg_write_barrier()                          │ Primary-Replica desync → PK-SK desync transition
    csn_committing_set = false                  │
    xlog_ptr_committing_set = false             │
    my_proc_info->vxids[nl].oxid = InvalidOXid  │
    advance_run_xmin(curOxid)                   │ ◄── PK-SK desync (primary-internal)
    curOxid = InvalidOXid                       ▲ ◄── WINDOW CLOSES
                                                │  After this line, get_current_oxid_if_any() returns
                                                │  InvalidOXid → undo_xact_callback's early-return at
                                                │  the case entry fires → skips wal_rollback,
                                                │  current_oxid_abort, AND apply_undo_stack entirely.
    release_assigned_logical_xids()             │
[current_oxid_commit returns]                   │
                                                ├── SAFE from this point ──
[if (enable_rewind) {...}]                      │
for (i) on_commit_undo_stack(...)               │
wal_after_commit()                              │
reset_cur_undo_locations()
reset_command_undo_locations()
oxid_needs_wal_flush = false
xidless_commit_lsn = InvalidXLogRecPtr
[case ends, break]
```

## Possible fix
Wrapping whole error window into CRIT_SECTION will help to shut down both primary-node and secondary-node, leaving only one question: is it possible for the replica to commit the tx before receiving the termination signal or not?

vvtrunov:
Wrap pure oriole tx commit frow into CRIT_SECTION to prevent primary-replica desync during error handling.

CRIT_SECTION must not contain any dynamic allocations, so it was necessary to remove stopevent + test.

Closes #869


vvtrunov:
@akorotkov test_committing_bit_cleared_on_abort() test seems to be redundant now as any error can't be raised inside CRIT_SECTION, so I've removed it

pashkinelfe (approved):

PR #882, opened 2026-06-04
Fix checkpointer-related anomalies in verify_orioledb 

pashkinelfe:
In verify_orioledb:
- Avoid premature lock acquisition inside checkpointer re-lock.
- Avoid false positive due to transient unflushed dirty pages (busy/free extents)

orioledb_tbl_check behavior remains unchanged as it's used to detect BROKEN_SPLIT/unflushed dirty pages in tests. So semantics of verify_orioledb  and orioledb_tbl_check becomes different, and we shouldn't pursue replacing orioledb_tbl_check by verify_orioledb

### DIFF
--- a/src/btree/check.c
+++ b/src/btree/check.c
@@ -52,6 +52,7 @@ typedef struct
 	ExtentsArray busy;
 	BTreeDescr *desc;
 	bool		hasError;
+	bool		emit_transient_notices;
 	OBTreeFindPageContext context;
 } BTreeCheckStatus;
 
@@ -92,6 +93,13 @@ check_btree(BTreeDescr *desc, bool force_file_check, bool wait_for_checkpoint)
 	/* get busy file extents */
 	status.desc = desc;
 	status.hasError = false;
+
+	/*
+	 * Legacy debug entry points (wait_for_checkpoint == false) keep the
+	 * BROKEN_SPLIT NOTICE; the user-facing verify_orioledb path stays silent
+	 * on that transient.
+	 */
+	status.emit_transient_notices = !wait_for_checkpoint;
 	init_page_find_context(&status.context, desc, COMMITSEQNO_INPROGRESS, BTREE_PAGE_FIND_MODIFY);
 
 	/*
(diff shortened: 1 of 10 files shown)


src/btree/check.c
```diff
@@ -73,32 +75,55 @@ static bool is_sorted_by_off(ExtentsArray *array);
 static bool is_sorted_by_len_off(ExtentsArray *array);
 
 bool
-check_btree(BTreeDescr *desc, bool force_file_check)
+check_btree(BTreeDescr *desc, ORelOids *oids, bool force_file_check)
```
akorotkov: I think we can use desc->oids, and IS_SYS_TREE_OIDS().

src/btree/check.c
```diff
@@ -73,32 +75,55 @@ static bool is_sorted_by_off(ExtentsArray *array);
...
+	 * per-tree state.  With oids supplied, wait the checkpointer out; without
+	 * oids the caller is a debug entry point and gets the legacy NOTICE.
+	 */
+	while (!get_checkpoint_number(desc, desc->rootInfo.rootPageBlkno,
+								  &checkpoint_number, &copy_blkno))
 	{
-		/* Error is possible only when calling check_btree() without lock */
-		elog(NOTICE, "Tree is under checkpoint now");
-		return false;
+		if (oids == NULL)
```
akorotkov: Can we wait on checkpoint_state->oSysTreesLock for finish of sys trees checkpointing?

akorotkov (approved):

src/btree/check.c
pashkinelfe: Done as of https://github.com/orioledb/orioledb/pull/882/changes/e96c57ade5c803f0aad354f1ebf20bad80cda853

src/btree/check.c
pashkinelfe: Done as of https://github.com/orioledb/orioledb/pull/882/changes/e96c57ade5c803f0aad354f1ebf20bad80cda853

PR #884, opened 2026-06-04
Take 2: Fix duplicate seq scan results from overlapping undo leaves

pashkinelfe:
Same as https://github.com/orioledb/orioledb/pull/880 but fix limited to logic of check_in_memory_leaf_page without introducing multi-abstraction-level logic


pashkinelfe:
Fixed by #891

PR #890, opened 2026-06-08
Keep xidmap and undo ring buffers populated across checkpoints

akorotkov:
## Summary

The `xidBuffer` (CSN map) and the three undo ring buffers used to be drained at checkpoint time:

- `fsync_xidmap_range()` called `write_xidsmap()`, which atomically swapped every slot to `COMMITSEQNO_FROZEN` and advanced `writtenXmin`. After a checkpoint, `map_oxid()` of any oxid in the just-flushed range went to the disk path until new traffic warmed the ring back up.
- `fsync_undo_range()` called `evict_undo_to_disk()`, which advanced `writtenLocation`. After a checkpoint, `undo_read()` of any location in the flushed range routed through `o_buffers_read()` instead of the in-memory ring, hurting rollback/snapshot fetch latency.

This PR teaches both subsystems to **persist data to disk without disturbing the ring**, using a per-page dirty bitmap.

## Design

Per-page atomic bitmap in shared memory (`pg_atomic_uint64` array, 1 bit per `ORIOLEDB_BLCKSZ` page of the ring):

- **Writers** mark their page dirty *after* storing the slot/record, with the release implied by `pg_atomic_fetch_or_u64`.
- **Checkpoint flush** (new `flush_dirty_xidsmap_range` / `flush_dirty_undo_range`) walks the pages in the requested range, does `pg_atomic_fetch_and_u64` to atomically test-and-clear each bit, and pushes pages whose bit was set out to `o_buffers`. Does *not* advance `writtenXmin` / `writtenLocation`; does *not* touch slot/record contents.
- **Slot-pressure eviction** (`write_xidsmap`, `evict_undo_to_disk`) is unchanged in its core behaviour but now also clears dirty bits for pages fully covered by the disk write, so subsequent checkpoint flushes can skip them.

Race semantics: a writer that re-sets the bit after the checkpointer cleared it leaves the bit set, so the next flush catches the update. Worst case is one redundant page write -- correctness holds without the writer ever taking the eviction lock.

## What changes

- `src/transam/oxid.c` (~180 lines): per-page dirty bitmap for `xidBuffer`, `mark_xid_buffer_dirty` / `test_clear_xid_buffer_page_dirty` helpers, new `flush_dirty_xidsmap_range`, `fsync_xidmap_range` switches to it, `write_xidsmap` + `advance_global_xmin` reset path clear bits for fully-covered pages.
- `src/transam/undo.c` (~225 lines): same pattern for each of the three undo logs; `mark_undo_range_dirty` covers multi-page records; `flush_dirty_undo_range`; `fsync_undo_range` switches to it; `evict_undo_to_disk` clears bits.

## Test plan

- [x] `make regresscheck` — 44/44 pass (subtransactions, rewind, toast, indices_build all exercise xidmap/undo paths)
- [x] `make isolationcheck` — 30/30 pass (isol_rr / rll_mix / concurrent_update_delete cover concurrent commit+rollback)
- [x] Focused smoke scenario: INSERT 5k → CHECKPOINT → ROLLBACK of UPDATE → savepoint ROLLBACK → UPDATE + CHECKPOINT mid-txn + ROLLBACK — all rollbacks correctly revert
- [x] `pgindent` clean

The CI run on this branch will exercise it across PG16/17/18 + sanitize/valgrind/check_page.

### DIFF
--- a/src/utils/o_buffers.c
+++ b/src/utils/o_buffers.c
@@ -420,6 +420,32 @@ o_buffers_write_if_exists(OBuffersDesc *desc, Pointer buf, uint32 tag,
 	return o_buffers_rw(desc, buf, tag, offset, size, true, true);
 }
 
+/*
+ * Write one ORIOLEDB_BLCKSZ-aligned page straight to the on-disk file,
+ * bypassing the o_buffers cache.
+ *
+ * Useful at checkpoint time when the caller knows the same data will
+ * stay hot in another in-memory copy (e.g. the xidmap / undo circular
+ * buffers, which keep the slot/record contents themselves and answer
+ * subsequent reads from there rather than from the o_buffers cache).
+ * Going through o_buffers_write() in that case would only pollute the
+ * cache with a copy that nobody will ever read.
+ *
+ * Caller must not call this concurrently for the same OBuffersDesc with
+ * other paths that mutate desc->curFile (the process-local open-file
+ * cache); the typical pattern is to hold the same write lock that
+ * serialises the eviction path.
+ */
+void
+o_buffers_write_page_direct(OBuffersDesc *desc, char *data, uint32 tag,
+							int64 offset)
+{
+	Assert(OBuffersMaxTagIsValid(tag));
+	Assert(offset >= 0 && offset % ORIOLEDB_BLCKSZ == 0);
+
+	write_buffer_data(desc, data, tag, offset / ORIOLEDB_BLCKSZ);
+}
+
 static void
 o_buffers_flush(OBuffersDesc *desc,
 				uint32 tag,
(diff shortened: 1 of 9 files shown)


pashkinelfe:
@akorotkov , I've added commit with asserts and small refactoring and commit with additional tests.

pgnickb:
I've added myself just to let you know I'm having a look and am testing a hypothesis that we have a race condition related to multi-page transactions. I will build some infra to be able to reproduce this in a separate branch and add a test. Will link to it later.

src/utils/o_buffers.c
```diff
@@ -394,30 +402,46 @@ void
...
+void
+o_buffers_write_clean(OBuffersDesc *desc, Pointer buf, uint32 tag,
+					  int64 offset, int64 size)
+{
+	Assert(OBuffersMaxTagIsValid(tag) && offset >= 0 && size > 0);
+	o_buffers_rw(desc, buf, tag, offset, size, true, false, true);
 }
 
 bool
 o_buffers_write_if_exists(OBuffersDesc *desc, Pointer buf, uint32 tag,
```
pashkinelfe, line 440: Do we need to consider "clean" in this call path?

akorotkov: I don't see a use case right now.

pashkinelfe (approved):

PR #891, opened 2026-06-08
Fix waiter-insert CSN sampling + cover pre-downlink-update split scan

akorotkov:
## Summary

Builds on the live-repro from #880 (commit c0d057b7) with two changes:

1. **`btree/insert.c`** — `o_btree_insert_item_with_waiters()` was sampling `TRANSAM_VARIABLES->nextCommitSeqNo` with a plain atomic read when stamping the CSN of a newly-created leaf-image undo item.  Concurrent waiter-inserts could therefore land different undo images under the same CSN.  A later seq scan walking the post-split parent would observe `leaf.csn == imgReadCsn` on every sibling, fall through the `csn >= imgReadCsn` branch and reconstruct the same pre-split leaf through undo for every downlink — the duplicate-key shape covered by the live test.  Switched to `pg_atomic_fetch_add_u64(..., 1)` so each needs-undo path consumes a distinct CSN.

2. **`test/t/transaction_test.py`** — two complementary coverage scenarios for the case where the reader's parent copy *predates* the new sibling downlink:

   - `test_seq_scan_internal_page_read_before_split_downlink`  (in-memory)
   - `test_seq_scan_internal_page_read_before_split_downlink_evicted` (after `orioledb_evict_pages` — goes through `load_next_disk_leaf_page`)

   Both halt the splitter on `STOPEVENT(page_split)` so the leaf is already split (L1 + new L2) but the parent still carries the pre-split single downlink.  The reader follows that single downlink to L1, reconstructs pre-split L through undo, never visits L2 (not in its copy of I), and must see exactly the 12 pre-split keys with no duplicates and no loss.

   While here, drop the diagnostic-log asserts (`source=btree-insert-with-waiters`, `scan-debug: ...`) and the `log_min_messages = DEBUG2` line from `test_live_waiter_split_seq_scan_reads_adjacent_undo_leaves`: the corresponding `elog(DEBUG2,...)` calls were never committed to the source tree, so those checks could only pass against a developer build with local instrumentation.  The real-invariant `assertEqual(duplicated_keys, [])` is preserved.

## Test plan

- [x] `python3.12 -m unittest test.t.transaction_test.TransactionTest` — all 5 tests pass, 3 of them are the seq-scan/waiter-split coverage above; ran each of the new tests in a loop, 3/3 stable
- [x] Without the CSN fix, the existing live repro still requires the seq-scan trim from #880; with both, every coverage scenario passes

### DIFF
--- a/src/btree/insert.c
+++ b/src/btree/insert.c
@@ -1061,7 +1061,7 @@ o_btree_insert_item_with_waiters(BTreeInsertStackItem *insert_item,
 
 	/* Get CSN for undo item if needed */
 	if (needsUndo)
-		csn = pg_atomic_read_u64(&TRANSAM_VARIABLES->nextCommitSeqNo);
+		csn = pg_atomic_fetch_add_u64(&TRANSAM_VARIABLES->nextCommitSeqNo, 1);
 	else
 		csn = COMMITSEQNO_INPROGRESS;
 
--- a/test/t/transaction_test.py
+++ b/test/t/transaction_test.py
@@ -54,6 +54,9 @@ Test orchestration:
 	and the cluster PANICs.
 """
 
+from collections import Counter
+import inspect
+import os
 import time
 
 from .base_test import BaseTest, ThreadQueryExecutor, wait_stopevent, wait_for_wait_event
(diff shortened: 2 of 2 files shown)


pashkinelfe:
@akorotkov fix looks good to me.

Added commit with test drop-out https://github.com/orioledb/orioledb/pull/891/changes/15c27e61bf0d36d3ed87478cae9897a9a14b0c5c

pashkinelfe (approved):

Issue #876, opened 2026-06-02
Replica stucks in live-lock after primary node shutdown


vvtrunov:
## Summary
If primary aborts in-flight tx due to force shutdown, replica stays in desync state even after primary reboot. The end-of-crash-recovery in-memory abort of in-flight transactions (`recovery_finish`) is **not propagated into the WAL stream**, so a streaming standby -- which applied those transactions' changes eagerly and never runs `recovery_finish` itself -- can never resolve their oxids, and livelocks when a later modify conflicts with one.

## Primary side abort process

When the primary reaches end-of-WAL in crash recovery, `recovery_finish` aborts every still-`INPROGRESS` oxid by applying its undo stack -- **emitting no WAL** [`src/recovery/recovery.c:1682–1693`](https://github.com/orioledb/orioledb/blob/main/src/recovery/recovery.c#L1681-L1693).

## Replica side abort process

Replica does not shutdown at all, it simply reconnects to primary and starts receiving new wal records upon previos internal state. Thus, it still has some `IN_PROGRESS` tx which actually is aborted during primary `recovery_finish`.
Thus, when replica receives some wal associated with tuples that have been modified by incomplete oxids, it simply locks in live-lock mode because replica is trying to get such a COMMIT/abort record from primary (primary itself has already marked in memory such oxids as aborted).

## Steps to repro

Problem fires with low probability, so it is better to run test in for-loop.

```
git checkout add_stress_bank_account_test
```

```
timeout 180 env \
  RR_REPLICA_MODE=streaming \   # second node = pg_basebackup'd hot-standby (this is essential)
  RR_INSTANCE=wedge_repro \     # names the saved logs under test/t/crash/results/
  RR_WRITERS=20 \               # 20 concurrent writers onto...
  RR_ACCOUNTS=100 \             # ...100 accounts -> hot, contended key space
  RR_DURATION=15 \              # workload seconds
  RR_ASSERT_POINTS=orioledb-commit-assert \  # which injection to weaponize
  RR_ASSERT_FIRINGS=2 \         # arm it ~2x over the run -> 2 PANIC/crash-recover cycles
  RR_INJECTION_POINTS=NONE \    # no other (non-fatal) injection chaos
  RR_PANIC_FATAL=0 \            # tolerate PANIC + allow cluster restart (must be 0: we need the post-crash recovery)
  RR_SAVE_ALL_LOGS=0 \          # keep server logs only when a bug/panic is detected
  python3 -m unittest -v \
[...]

PR #892, opened 2026-06-08
Defer free_run_xmin to after WAL_REC_ROLLBACK emission

### LINKED ISSUE #889: Recovery livelock due to the standby regresses globalXmin below writtenXmin after primary-node shutdown
## Summary 
When one of the worker processes fails, OrioleDB aborts all active transactions by reverting them by applying undo, so no WAL records are provided for the standby replica node. After the primary cluster restart, WAL_ROLLBACK records are produced to signal the replica node to abort such transactions as well (see #878). However, such behavior does not respect minXid invariants, because after the primary restart the min oxid may already be greater than the oxid of the aborted transaction, so the WAL record with the rollback will be malformed. Therefore, the replica node can lower the Xmin frontier below the actually committed transaction; such a scenario leads to a replica live-lock, because a transaction that has actually already finished is marked as IN-PROGRESS.

## Bug scenario

Bug scenario is built according to the buggy run logs

```
SIGKILL primary mid-transaction (oxid 248 in-flight, no terminator in WAL)
  → restarted primary recovery_finish() aborts 248 in memory, but runXmin has raced to 468
  → wal_emit_recovery_finish_rollback(248) emits ROLLBACK stamped xmin=runXmin=468 (≫ 248)  [wal.c:405]
  → standby replays ROLLBACK(248) → recovery_switch_to_oxid → pairingheap_add(xmin_queue)    [recovery.c:1899]
       (inserts oxid 248 below the standby's writtenXmin=352)
  → update_run_xmin reads queue-min=248 → writes globalXmin=248 < writtenXmin=352            [recovery.c:2502]
  → oxid_get_csn skips the oxid<globalXmin fast-path, reads FROZEN slot, maps it to IN_PROGRESS
  → recovery workers spin on committed-frozen oxids in [248,352) → listPtr pinned → livelock
```

### Confirmed with logs
A spin-site trace was added at the exact point a recovery worker reads `conflictOxid` as in-progress (`o_btree_modify_handle_conflicts`, `src/btree/modify.c`): it logs the **raw `xidBuffer` slot value** plus `globalXmin`/`writtenXmin` via the `oxid_debug_raw_slot()` accessor (`src/transam/oxid.c`). A trial caught on 2026-06-05 (`CAUGHT_noinj_replica.log`, 6.4 GB, 5,109,500 spins) captured the whole chain **on the replica**.

**The cause — the replica's own `update_run_xmin()` writes `globalXmin` backward, below `writtenXmin`:**
```
17:35:48.992 [8513] runxmin-backward func=update_run_xmin
             old_globalXmin=352 → new_globalXmin=248  writtenXmin=352  nextXid=468  recovery_xmin=468  queueEmpty=0  belowWritten=1
```
`queueEmpty=0` ⇒ the backward value came from the `xmin_queue`'s oldest entry (a committed/frozen oxid re-entered the in-flight set and dragged the horizon back), **not** the `nextXid` fallback.

**The effect — 1 ms later, every spinning worker reads a FROZEN slot and gets `IN_PROGRESS`:**
```
17:35:48.993 [8516] conflict-inprogress opOxid=378 conflictOxid=320 rawSlotCsn=3 globalXmin=248 writtenXmin=352 oxidGEglobalXmin=1 belowWritten=1
17:35:48.994 [8517] conflict-inprogress opOxid=427 conflictOxid=337 rawSlotCsn=3 globalXmin=248 writtenXmin=352 oxidGEglobalXmin=1 belowWritten=1
17:35:48.998 [8518] conflict-inprogress opOxid=469 conflictOxid=351 rawSlotCsn=3 globalXmin=248 writtenXmin=352 oxidGEglobalXmin=1 belowWritten=1
```

Every field corroborates the mechanism:

- **`rawSlotCsn=3` = `COMMITSEQNO_FROZEN`** on *all* spinners. The slot is **never** rewritten to `INPROGRESS (0x0)` — it holds the legitimate `FROZEN` value `advance_global_xmin` stamped when the oxid committed. **No clobber, no revert.**
- **`globalXmin=248`** equals the value the regression just wrote (352→248), with **1 ms causal ordering** (`.992` regress → `.993` first spin).
- conflictOxids **320 / 337 / 351 all lie in the band `[248, 352)`** = the re-exposed frozen region:
  - `≥ globalXmin (248)` ⇒ `oxidGEglobalXmin=1` ⇒ the `oxid < globalXmin ⇒ FROZEN` fast-path in `oxid_get_csn` is **skipped**;
  - `< writtenXmin (352)` ⇒ `belowWritten=1` ⇒ the slot is already frozen.

So `oxid_get_csn(320)` skips its fast-path, reads the slot (`FROZEN`, a *special* value), maps the special value to **`IN_PROGRESS`**, and the worker spins forever. It is not one oxid but the **entire `[248,352)` band** of committed-and-frozen oxids mis-resolved at once (here, 3 workers on 3 oxids).

## Repro steps

Streaming-standby crash loop, **no injection**:

```bash
git checkout add_stress_bank_account_test

...

# Drive the bank-account stress harness against a streaming standby and SIGKILL the
# primary postmaster on a fixed interval; watch the replica fail to shut down.
RR_REPLICA_MODE=streaming \
RR_INJECTION_POINTS=NONE \
RR_ASSERT_FIRINGS=0 \
RR_KILL_POSTMASTER=1 \
RR_KILL_POSTMASTER_INTERVAL=6 \
RR_PANIC_FATAL=0 \
RR_SAVE_ALL_LOGS=1 \
RR_DURATION=45 \
RR_WRITERS=8 \
python3 -m unittest test.t.crash.rr_stress_test.RrStressTest.test_bank_account_invariant
```

A caught trial is identified by: replica `postgresql.log` growing without bound, `handle_conflicts` spin count climbing into the millions, and `pg_ctl stop` on the replica timing out ("server does not shut down"). Last buggy run replica logs was ~ 6.5 GB in a few minutes.

akorotkov:
## Summary

Fixes the standby livelock from #889.

### What was wrong

`recovery_finish()` called `free_run_xmin()` at the very end of crash
recovery, lifting master's `runXmin` to `nextXid` before the
post-recovery checkpoint ran.  That advanced horizon was then
persisted as `control.checkpointRetainXmin` in the new control file.

When `o_emit_recovery_finish_rollbacks()` later emitted the deferred
`WAL_REC_ROLLBACK` records from `o_after_checkpoint_cleanup_hook()`,
the standby's WAL stream was:

  1. post-recovery checkpoint (claims a high `checkpointRetainXmin`),
  2. then `WAL_REC_ROLLBACK(oxid, xmin=high)` for each in-flight
     transaction master aborted in memory.

The standby replayed the checkpoint, slid `globalXmin` forward to the
checkpoint's claimed horizon, then ingested `WAL_REC_ROLLBACK` whose
`recovery_switch_to_oxid()` pushed the rolled-back oxid into
`xmin_queue` — dragging `globalXmin` back across a band where master's
`advance_global_xmin` had already stamped FROZEN.

`oxid_get_csn()`'s fast-path (`oxid < globalXmin => COMMITSEQNO_FROZEN`)
no longer covered those slots; the wrapper's
`COMMITSEQNO_IS_SPECIAL(csn) => INPROGRESS` fallback then turned the
legitimate FROZEN values into `INPROGRESS`, and recovery workers in
`o_btree_modify_handle_conflicts()` spun forever on
committed-and-frozen oxids — the livelock.

### The fix

Move the `free_run_xmin()` call out of `recovery_finish()` and into
`o_emit_recovery_finish_rollbacks()`, after every WAL record has been
emitted.  The post-recovery checkpoint now runs with master's
`runXmin` still at the pre-recovery floor, so
`control.checkpointRetainXmin` matches the horizon the subsequent
`WAL_REC_ROLLBACK` records carry.  The standby never observes a
`checkpointRetainXmin` that would justify pushing `globalXmin` past a
stamped-FROZEN band, and `oxid_get_csn()`'s fast-path stays valid.

## Tests

- `test_parallel_worker_rollback_undo_lost_to_compaction` — new test
  exercising the parallel-worker side of the invariant via stopevents
  (parks rollback undo while concurrent workers revisit the deleted
  tuple's page).  Passes with the fix.
- `test_recovery_finish_rollback_does_not_regress_replica_xmin` —
  restored from a previous iteration.  Currently fails on a separate
  \"stuck low globalXmin\" symptom (replica's `globalXmin` does not
  catch up to master's after recovery completes) that is orthogonal
  to the livelock fix; left in place so the diagnostic surface
  remains visible.
- Existing #876 covering test
  `test_recovery_finish_aborts_propagate_to_replica` still passes.

## Test plan

- [x] \`make regresscheck\` — 44/44 OK
- [x] \`recovery_test\` testgres suite — 56/56 OK
- [x] \`replication_test\` testgres suite — 35/36 OK (one expected
      failure documented above)

Fixes #889.

### DIFF
--- a/src/checkpoint/checkpoint.c
+++ b/src/checkpoint/checkpoint.c
@@ -355,15 +355,40 @@ checkpoint_shmem_init(Pointer ptr, bool found)
 			pg_atomic_write_u64(&undo_meta->cleanedCheckpointEndLocation, undo_info->checkpointRetainEndLocation);
 		}
 
-		pg_atomic_init_u64(&xid_meta->nextXid, control.lastXid);
-		pg_atomic_init_u64(&xid_meta->runXmin, control.lastXid);
-		pg_atomic_init_u64(&xid_meta->globalXmin, control.lastXid);
-		pg_atomic_init_u64(&xid_meta->lastXidWhenUpdatedGlobalXmin, control.lastXid);
-		pg_atomic_init_u64(&xid_meta->writtenXmin, control.lastXid);
-		pg_atomic_init_u64(&xid_meta->writeInProgressXmin, control.lastXid);
+		/*
+		 * Seed xid horizons from the checkpoint's retain range:
+		 *
+		 * - nextXid / writtenXmin / writeInProgressXmin =
+		 * checkpointRetainXmax (master's nextXid at checkpoint time).
+		 * writtenXmin stays at the high mark so the on-disk xidmap range --
+		 * which may contain FROZEN slots that master's advance_global_xmin
+		 * stamped for oxids < master.runXmin -- is still served via
+		 * o_buffers_read.
+		 *
+		 * - runXmin / globalXmin / cleanedXmin / lastXidWhenUpdatedGlobalXmin
+		 * = checkpointRetainXmin (master's runXmin at checkpoint time, the
+		 * actual floor of in-flight oxids).  This is the conservative
+		 * starting horizon that matches what subsequent WAL records will
+		 * reference.
+		 *
+		 * On a clean shutdown master writes checkpointRetainXmin ==
+		 * checkpointRetainXmax == nextXid, so all horizons collapse to a
+		 * single value -- behaviour identical to the previous lastXid-based
+		 * init.
+		 *
+		 * control.lastXid is now redundant with checkpointRetainXmax; it's
+		 * retained in the control file for backward compatibility with
+		 * existing data directories.
+		 */
+		pg_atomic_init_u64(&xid_meta->nextXid, control.checkpointRetainXmax);
+		pg_atomic_init_u64(&xid_meta->runXmin, control.checkpointRetainXmin);
+		pg_atomic_init_u64(&xid_meta->globalXmin, control.checkpointRetainXmin);
+		pg_atomic_init_u64(&xid_meta->lastXidWhenUpdatedGlobalXmin, control.checkpointRetainXmin);
+		pg_atomic_init_u64(&xid_meta->writtenXmin, control.checkpointRetainXmax);
+		pg_atomic_init_u64(&xid_meta->writeInProgressXmin, control.checkpointRetainXmax);
 		pg_atomic_init_u64(&xid_meta->checkpointRetainXmin, control.checkpointRetainXmin);
 		pg_atomic_init_u64(&xid_meta->checkpointRetainXmax, control.checkpointRetainXmax);
-		pg_atomic_init_u64(&xid_meta->cleanedXmin, control.lastXid);
+		pg_atomic_init_u64(&xid_meta->cleanedXmin, control.checkpointRetainXmin);
 		pg_atomic_init_u64(&xid_meta->cleanedCheckpointXmin, control.checkpointRetainXmin);
 		pg_atomic_init_u64(&xid_meta->cleanedCheckpointXmax, control.checkpointRetainXmax);
 
--- a/src/recovery/recovery.c
+++ b/src/recovery/recovery.c
@@ -1766,10 +1766,24 @@ recovery_finish(int worker_id)
 		pairingheap_free(retain_undo_queues[i]);
 	}
 	if (worker_id < 0)
-	{
 		pairingheap_free(xmin_queue);
-		free_run_xmin();
-	}
+
+	/*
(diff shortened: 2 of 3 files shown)


pashkinelfe:
Maybe it's worth to invalidate nextXid in the future checkpoints https://github.com/orioledb/orioledb/commit/96dafd592c121ae3eadd2d97ececb765b7d79edb

PR #895, opened 2026-06-10
btree skip scan: tuple-aware probe-then-narrow for PG18 skip arrays

akorotkov:
## Summary

Wires PG18's btree skip-scan optimization ([upstream commit 92fe23d93aa](https://github.com/postgres/postgres/commit/92fe23d93aa)) into OrioleDB so a scan that lacks an equality on the leading column actually skips to each distinct value present in the index, instead of scanning the entire `low_compare..high_compare` span with only a trailing-column filter.

## Approach

OrioleDB drives index scans from key-range bounds rather than per-tuple btree callbacks, so the upstream entry point `_bt_advance_array_keys` (which requires an `IndexTuple`) is not usable from `switch_to_next_range`.  Instead this PR vendors the small static array-advancement primitives upstream uses internally and wires them into a probe-then-narrow loop driven by OrioleDB's existing iterator.

Flow per scan:

1. **Sentinel probe.**  `switch_to_next_range` opens an iterator over the broad sentinel range (MINVAL/MAXVAL/SearchNull-initial); `o_iterate_index` marks `skipScanProbePending`.
2. **Pin from tuple.**  First non-NULL tuple is consumed by `o_skip_arrays_observe_tuple`, which pins `sk_argument` to its leading-column value via the vendored `o_bt_skiparray_set_element_from_tuple`.
3. **Narrow scan.**  The iterator is dropped and reopened inline with a point-bound `[sk_argument, sk_argument]` range; trailing-column predicates filter via `is_tuple_valid` as before.
4. **Tuple-aware advance.**  When the narrow iterator exhausts, `o_bt_advance_array_keys_increment` sets `SK_BT_NEXT` (forward) or `SK_BT_PRIOR` (backward) instead of using `SkipSupport.increment`.  This is the key correctness/termination decision: `sk_argument` is **not** moved by `SkipSupport`, so the next iterator opens with an exclusive bound past `sk_argument` and descends to the **next actual distinct value present in the index** rather than blindly stepping through discrete next-of-type values (which would infinite-loop on sparse data — e.g. an int8 column with only 100 rows would otherwise step through int8 max before terminating).
5. **Natural termination.**  If the probe iterator finds no tuple past `sk_argument` within `low_compare..high_compare`, it returns NULL and the scan ends.

## Commits

1. `test: regression coverage for PG18 btree skip-scan queries` — baselines for int / text / backward / multi-distinct cases.
2. `btree skip scan: vendor upstream array primitives + sentinel-safe advance` — `o_bt_array_increment` / `_decrement` / `_set_low_or_high` / `_skiparray_set_isnull` / `_skiparray_set_element_from_tuple` (kept structurally identical to upstream for mechanical diffs on future PG rebases).  Reworks `o_bt_advance_array_keys_increment` to honor upstream's MINVAL/MAXVAL sentinel contract.
3. `btree skip scan: handle post-advance state in o_key_data_to_key_range` — point bound for concrete `sk_argument`, exclusive bounds for `SK_BT_NEXT` / `SK_BT_PRIOR`, sentinel fallback (including `SK_ISNULL`-initial backward+nulls-last case) keeps the pre-skip-scan broad-range behavior.
4. `btree skip scan: tuple-aware probe-then-narrow iteration` — the iterator integration: probe-pending flag in `OScanState`, inline iterator rebuild post-probe (bypasses `switch_to_next_range` to avoid `_bt_start_array_keys` overwriting the pin), advance via `SK_BT_NEXT`/`PRIOR`.
5. `pgindent` — mechanical reformat across the three files touched.

## Test plan

- [x] `make regresscheck` — all 44 OrioleDB regression tests green locally.
- [x] `make isolationcheck ISOLATIONCHECKS=\"btree_iterate isol_rc\"` — green.
- [x] New test cases in `test/sql/indices.sql` exercise:
  - `int` leading (SkipSupport-aware opclass).
  - `text` leading (no SkipSupport; `SK_BT_NEXT` path is the only route).
  - Backward scan with `int` leading (`SK_BT_PRIOR` path).
  - Multi-distinct iteration: 10 distinct `x` × 10 matching `y` = 100 rows, verifies the full sentinel→narrow→probe→narrow cycle end-to-end.
- [ ] CI on this PR.

## Notes

- Pre-PG18 builds compile out the new paths via `#if PG_VERSION_NUM >= 180000`; behavior on PG16/PG17 is unchanged.
- Non-skip-array scans are unaffected: `skipScanProbePending` is only set when at least one skip array is in sentinel/transition state.

### DIFF
--- a/include/tableam/index_scan.h
+++ b/include/tableam/index_scan.h
@@ -33,6 +33,20 @@ typedef struct OScanState
 	bool		curKeyRangeIsLoaded;
 	int			numPrefixExactKeys;
 	bool		exact;
+#if PG_VERSION_NUM >= 180000
+
+	/*
+	 * Skip-scan probe state.  Set when the current key range was opened over
+	 * a skip array still in its sentinel (MINVAL/MAXVAL/ SearchNull-initial)
+	 * or transition (SK_BT_NEXT/SK_BT_PRIOR) state; in that case the iterator
+	 * is positioned at a probe range (broad slice over
+	 * low_compare..high_compare, or exclusive past sk_argument), and the next
+	 * non-NULL fetched tuple's leading-column value is used to pin
+	 * sk_argument and reopen the iterator with a narrow point-bound range. If
+	 * no tuple is found, the scan terminates naturally.
+	 */
+	bool		skipScanProbePending;
+#endif
 	OBTreeKeyRange curKeyRange;
 	BTreeIterator *iterator;
 	List	   *indexQuals;
--- a/src/btree/iterator.c
+++ b/src/btree/iterator.c
@@ -328,9 +328,22 @@ o_btree_find_tuples_start(BTreeDescr *desc, void *key,
 	it->curKeyReturned = false;
 	it->startKey = NULL;
 	it->startKind = BTreeKeyNone;
+	it->pageCount = 1;
+	BTREE_PAGE_LOCATOR_SET_INVALID(&it->undoLoc);
 #ifdef USE_ASSERT_CHECKING
 	O_TUPLE_SET_NULL(it->prevTuple.tuple);
 #endif
+
+	/*
+	 * combinedPage is only assigned on the combined-result branch below (via
+	 * load_page_from_undo()); the plain non-combined path returns without
+	 * touching it.  A reused iterator would then keep the recycled chunk's
+	 * garbage value, and o_btree_iterator_fetch_internal() reads it to decide
+	 * whether to merge the (here uninitialized) undo image into the result --
+	 * a garbage-true value dereferences that stale image and returns a bogus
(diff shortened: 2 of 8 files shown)


pashkinelfe:
@akorotkov I've added a couple of commits to a separate branch https://github.com/orioledb/orioledb/tree/pg18-btree-skip-scan-add. Please merge them to this PR if you agree

pashkinelfe (commented):
I've added a couple of commits to a separate branch https://github.com/orioledb/orioledb/tree/pg18-btree-skip-scan-add. Please merge them to this PR if you agree

za-arthur (approved):

pashkinelfe (approved):

homper (approved):

PR #899, opened 2026-06-10
recovery: keep replica horizons honest under fast-path master aborts

akorotkov:
## Summary

Two fixes to the recovery xid-horizon machinery that together close the symptom the user reported in #889: replica's globalXmin/runXmin sailing past the master's actual floor when the master has long-running oxids whose modify records have not yet flushed off the backend's local_wal buffer.

### Commit 1: pin `recovery_xmin` to `checkpointRetainXmin` before `read_xids()`

`read_xids()` runs an `update_run_xmin()` pass that **unconditionally** overwrites `xid_meta->runXmin` with the smallest oxid in `xmin_queue`, falling back to `nextXid` when the heap is empty.  On a streaming standby this happens whenever the master's only in-flight oxid X has all its modify records still buffered in `local_wal` -- the master's `xids` file names X but the standby's recovery xid hash (and therefore its own restartpoint's `xids` file) doesn't.  After a clean standby restart `update_run_xmin()` takes the empty-heap branch, `recovery_xmin` is still at its static-init `InvalidOXid` (effectively unbounded), the `Min(xmin, recovery_xmin)` cap is inert, and `runXmin` jumps to `nextXid`.

Fix: seed `recovery_xmin` from `xid_meta->checkpointRetainXmin` *before* `read_xids()` runs.  The `Min(...)` cap now keeps `runXmin` pinned at the checkpoint floor on the empty-heap path; subsequent `WAL_REC_COMMIT` / `WAL_REC_ROLLBACK` / `WAL_REC_JOINT_COMMIT` records bump `recovery_xmin` via the existing `Max()` exactly as before.

### Commit 2: drop fast-path-aborted oxids off `xmin_queue` in `update_run_xmin`

After the floor seeding above, `runXmin` is still pinned by oxids that were named in the checkpoint's xids file but never produced a `WAL_REC_XID` on the wire -- the master's `wal_rollback()` fast path takes them without writing anything when the transaction had no material changes.  Such an oxid sits at the top of `xmin_queue` forever (no `WAL_REC_COMMIT` / `WAL_REC_ROLLBACK` will ever arrive to remove it), even though `recovery_xmin` has already been bumped past it by a later finish record carrying the master's post-abort runXmin.

Recognise the combination `(checkpoint_xid && !wal_xid && state->oxid < recovery_xmin)` in `update_run_xmin()` and process the oxid as aborted right there:
- `set_oxid_csn(state->oxid, COMMITSEQNO_ABORTED)` so visibility checks see a settled txn rather than `COMMITSEQNO_INPROGRESS`.
- `walk_checkpoint_stacks(state, COMMITSEQNO_ABORTED, ...)` to apply any lock-only undo the master captured (a fast-path abort can carry no-modify-record but non-empty lock undo).
- Reset `recovery_oxid` / `curUndoLocations` / `oxid_needs_wal_flush` back to the no-current-oxid state.
- Drop the entry from `xmin_queue`.

Also turn the `globalXmin` handling into a one-way invariant.  The pre-existing `if (xmin < globalXmin) globalXmin = xmin;` was the last remaining write that could regress a published horizon; with the checkpoint-era floor now seeded by `checkpoint_shmem_init()` and `recovery_xmin` pinned by `recovery_init()`, every legitimate value of `xmin` in `update_run_xmin()` is `>= globalXmin`.  Replace the write with an `Assert(xmin >= globalXmin)` so any future code path that violates monotonicity is caught up front.

## Test plan

- [x] **New** `test/t/replication_test.py::test_replica_restart_seeds_runxmin_from_checkpoint_floor` -- opens a long-running tx X whose UPDATE fits in the 8 KB local_wal buffer (so it never flushes), runs ~200 short commits past it, checkpoints master, restarts the replica, and asserts the replica's runXmin / globalXmin do not exceed master's runXmin by more than a small constant.  Pre-fix: gap ≈ 200.  Post-fix: passes.
- [x] Existing #876 cover: `test_recovery_finish_aborts_propagate_to_replica` -- still passes.
- [x] Existing #889 cover: `test_recovery_finish_rollback_does_not_regress_replica_xmin` -- still passes.
- [x] `make regresscheck` (all 44 pass)
- [x] `replication_test` + `recovery_test` testgres suites (all 94 pass)

### DIFF
--- a/src/recovery/recovery.c
+++ b/src/recovery/recovery.c
@@ -1272,6 +1272,7 @@ o_recovery_finish_hook(bool cleanup)
 		}
 	}
 
+	update_proc_retain_undo_location(-1);
 	recovery_finish(-1);
 
 	if (!recovery_single)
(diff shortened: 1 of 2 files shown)


pashkinelfe (approved):

pashkinelfe:
@akorotkov I've added some tests for update_run_xmin invariant checked by Assert in some edge cases of primary workload https://github.com/orioledb/orioledb/pull/899/commits/dd5f2a4bb5526e75fcd8a598df0c2130f225a245

src/recovery/recovery.c
```diff
@@ -2507,6 +2534,93 @@ static void
...
+	 */
+	while (!pairingheap_is_empty(xmin_queue))
+	{
+		RecoveryXidState *state;
+
+		state = pairingheap_container(RecoveryXidState, xmin_ph_node,
+									  pairingheap_first(xmin_queue));
+
+		if (state->oxid >= recovery_xmin)
+			break;
```
pashkinelfe, line 2599: Does it need to continue to check all other entries in xmin_queue?

src/recovery/recovery.c
```diff
@@ -2507,6 +2534,93 @@ static void
...
+		if (!state->checkpoint_xid || state->wal_xid)
+			break;
```
pashkinelfe, line 2601: Does it need to continue to check all other entries in xmin_queue?

src/recovery/recovery.c
akorotkov: We can safely rollback only transactions less than recovery_xmin. They are guaranteed to be finished on primary, and no more WAL for them.

src/recovery/recovery.c
```diff
@@ -2507,6 +2534,93 @@ static void
 update_run_xmin(void)
 {
 	OXid		xmin;
+	int			i;
+	bool		found;
+
```
pashkinelfe, line 2561: Worth an `Assert(worker_id < 0) ` to avoid anyone adding calls of update_run_xmin without this condition in the future?

src/recovery/recovery.c
akorotkov: We can't do a cheap ordered scan of a heap.  The current approach is to shift the horizon as much as possible each time.  update_run_xmin() gets called after each transaction finishes and recovery_xmin shift.  So, the runXmin horizon must be kept up-to-date.

src/recovery/recovery.c
akorotkov: Do you propose to add worker_id as an argument to update_run_xmin()?

pashkinelfe: Maybe comment for the update_run_xmin() that it should be caller only from recovery leader would be better

za-arthur: There is a static global `xmin_queue` variable:
https://github.com/orioledb/orioledb/blob/5e836bf07b9fd6ab7315094e5ccd68f6684c3da0/src/recovery/recovery.c#L1545

It is used only in the case of `worker_id = -1`, but for some reason allocated in all cases. We can assign `NULL` when it isn't needed.

za-arthur: The assert can be in that case:
```
Assert(xmin_queue != NULL);
```

akorotkov: @za-arthur, thank you. I did it as you proposed!

pashkinelfe (approved):

PR #900, opened 2026-06-11
Don't emit wal_rollback record for oxid below recoveryXmin

vvtrunov:
Txs aborted without any persistent wal records (e.g. has_material_changes = false wal_rollback path) are considered as an in-flight during the deferred wal emiting. Thereby, if the replica node has already received some commit/rollback wal records  with rised oxid horizon, it would have lowered globalXmin on replica under the horizon, breaking the invariants.

recovery_xmin extracts such oxid horizon invariant from the whole wal sequence, it fully reflects the replica side horizon at the crash-recovery time.


akorotkov:
I'm closing this as issue must be already fixed by #899.

PR #902, opened 2026-06-12
Skip page-level undo when no concurrent readers or local deletes

pashkinelfe:
Page-level undo (pre-split/pre-compact page image) is needed only to
serve concurrent readers or rollback of in-flight changes on the same
page.  On bulk-insert workloads it dominates undo bandwidth and is
rarely consulted.


pashkinelfe:
Superseded by #907

PR #904, opened 2026-06-15
Preload jemalloc in the Alpine Docker image

pashkinelfe:
Replace musl's allocator can with jemalloc that scales better


pashkinelfe:
Performance effect is present but not pronounced. Alpine is an inferior target currently.

PR #907, opened 2026-06-17
Differential page-level undo images for B-tree merge and split

akorotkov:
## What

A B-tree page merge or split writes a copy of the pre-operation page(s) into
the page-level undo log so older snapshots can still see the original page(s).
Originally that was always a full copy — **2×8 KB for a merge, 8 KB + split
key for a split** — even though the operation only redistributes tuples and
physically drops nothing in the common case.

This series makes that page-level image **as cheap as the situation allows**:

- **No image at all** in the common case — a no-drop split/merge that no
  active snapshot is old enough to need.  The result page(s) are frozen and
  read live; per-tuple MVCC does the rest.
- A **differential** image (tens of bytes instead of kilobytes) when nothing
  is dropped but a snapshot *is* held over the page.
- A **full** image only when a tuple is physically dropped, or when a
  concurrent sequential scan may reconstruct whole pre-op pages.

This is a big win for the page-level undo produced during bulk load, where
splits are frequent and almost never need any image.

## How it works

Each operation classifies itself at op time.

**Skip the image entirely** when all of:

- the op drops no tuple (`page_op_drops_tuple()`, mirroring the leaf drop
  condition of `merge_pages()` / `make_split_items()`);
- the page's own pre-op `undoLocation` is already below the retain horizon
  (`minProcRetainLocation` / `minRewindRetainLocation`) or invalid — i.e. no
  active snapshot is older than the page (a merge requires this of *both*
  pages);
- no sequential scan is active on the tree (`meta_page_get_num_seq_scans()`).

The result page(s) then get `csn = COMMITSEQNO_FROZEN` and an invalid undo
location, exactly like a freshly initialized page: every reader reads them
live (a frozen csn precedes any snapshot, so the undo chain is never walked)
and per-tuple MVCC handles tuple visibility.  A backward index scan that steps
onto such a frozen split half takes the page's lokey from its smallest stored
key (valid precisely because nothing was dropped).

**Otherwise** write a page-level image:

- **`UndoPageImageMergeDiff`** / **`UndoPageImageSplitDiff`** (differential)
  when nothing is dropped but a snapshot is held: store the boundary/split key
  + the relevant `(csn, undoLocation)` continuation, and reconstruct from the
  newer page already sitting in the reader's buffer.  A differential split
  half only covers one leaf's key range — fine for point lookups (MVCC) and
  index/range scans (which clip per-tuple to the current leaf).
- a **full** image when a tuple is dropped, or when a sequential scan is
  active (it reconstructs whole pre-op pages spanning the full key range,
  which a differential half cannot provide across multi-level operations).

Why a dropped tuple forces an image: droppability is xid-based (the deleting
xid is older than `runXmin`), but an active snapshot whose csn precedes the
delete's *commit* can still need the tuple — only the image preserves it.

## Commits

1. **always seed page image before walking page-level undo** — foundation:
   copy the live page into the destination before walking the undo chain
   (differential images reconstruct in place from it).
2. **materialize partial page before walking page-level undo** — the
   fetch/index/bitmap paths read leaves by chunk; differential reconstruction
   needs every chunk, so `partial_load_full_page()` fully materializes a
   partially-loaded page first.
3. **differential page-level undo images for merge and split** — the
   `UndoPageImageMergeDiff` / `UndoPageImageSplitDiff` records and their read
   paths.
4. **mask undo location in sys-tree structure dump** — absolute undo
   locations shift when undo allocation shrinks; mask them so the dump test is
   stable.
5. Review follow-ups (per-type undo image headers, asserts, zeroing,
   `null_unused_bytes` reuse, shared test setup, header-layout StaticAssert,
   PG18 undo-buffer-floor test sizing).
6. **skip the page-level undo image for invisible splits and merges** — the
   no-image fast path described above (frozen pages + smallest-key lokey for
   backward scans).

## Testing

- `regresscheck` 44/44, `isolationcheck` 30/30.
- `merge_test`: no-image **and** differential merge/split exercised through
  seq, index, point, bitmap and UPDATE-driven paths under old snapshots.
- `load_refind_page` / `rightlink` / `btree_iterate` / `temp_local_pool_test`:
  concurrent eviction + seq/range scans, including backward scans over frozen
  no-image split halves.

## Known limitation

The seq-scan guard checks `numSeqScans` at op time, but a transaction takes
its snapshot before the orioledb seq-scan node registers.  A no-drop op that
lands in that window while a snapshot is held writes a **differential** image,
which a later seq scan (with the older snapshot) cannot fully reconstruct
across multi-level operations.  The no-image (frozen) fast path is unaffected —
it is read live, so a seq scan over it never reconstructs.  Closing the
differential-image window fully (publishing a seq-scan snapshot horizon at
snapshot acquisition) is left as a follow-up.

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -167,6 +167,7 @@ ISOLATIONCHECKS = bitmap_hist_scan \
 				  rll_deadlock \
 				  rll_mix \
 				  rll_subtrans \
+				  skipundo \
 				  table_lock_test \
 				  concurrent_truncate \
 				  uniq
--- a/Makefile
+++ b/Makefile
@@ -215,6 +215,7 @@ TESTGRESCHECKS_PART_2 = test/t/checkpoint_concurrent_test.py \
 						test/t/reindex_test.py \
 						test/t/s3_test.py \
 						test/t/schema_test.py \
+						test/t/seq_scan_undo_test.py \
 						test/t/temp_local_pool_test.py \
 						test/t/toast_index_test.py \
 						test/t/trigger_test.py \
--- a/include/btree/page_chunks.h
+++ b/include/btree/page_chunks.h
@@ -31,6 +31,7 @@ typedef struct
 } BTreePageItem;
 
 extern bool partial_load_hikeys_chunk(PartialPageState *partial, Page img);
+extern bool partial_load_full_page(PartialPageState *partial, Page img);
 extern bool partial_load_chunk(PartialPageState *partial, Page img,
 							   OffsetNumber chunkOffset,
 							   BTreePageItemLocator *loc);
--- a/include/btree/undo.h
+++ b/include/btree/undo.h
@@ -172,6 +172,19 @@ typedef struct
 /* maximum size of undo record */
 #define O_MAX_UNDO_RECORD_SIZE O_MERGE_UNDO_IMAGE_SIZE
 
+/*
+ * O_UNDO_GET_IMAGE_LOCATION() and get_page_from_undo() locate the page image
+ * at a fixed offset past the header, derived from UndoPageImageHeader, for
+ * every full-page image type: Compact and Merge use UndoPageImageHeader, while
+ * Split uses UndoPageImageSplitHeader.  That single offset is only correct if
+ * all full-image headers MAXALIGN to the same size.  Guard it so that, e.g.,
+ * adding a field to UndoPageImageSplitHeader cannot silently shift the split
+ * page image out from under those readers.
+ */
+StaticAssertDecl(MAXALIGN(sizeof(UndoPageImageHeader)) ==
+				 MAXALIGN(sizeof(UndoPageImageSplitHeader)),
+				 "full-page undo image headers must MAXALIGN to the same size");
+
 extern bool page_item_rollback(BTreeDescr *desc, Page p, BTreePageItemLocator *locator,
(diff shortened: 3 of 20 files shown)


pashkinelfe:
I removed my commit "Size max_procs by autovacuum_worker_slots on PG18" from this branch with force-push as it is already in main branch https://github.com/orioledb/orioledb/commit/5568d55be95acf6b237f7ef47e0470979799808b. All other commits unchanged.

pashkinelfe (commented):
I've added a couple of minor commits that are necessary for CI to pass to a current branch:
https://github.com/orioledb/orioledb/pull/907/changes/20fec190f1f98c8699d4d75f056d133f9aea0994
https://github.com/orioledb/orioledb/pull/907/changes/6135d99396ca234129d6c2551780d2743eb94759

Also I made several granular proposals for this PR in a separate branch:
https://github.com/orioledb/orioledb/commits/page_level_diff_undo_records-add1/
@akorotkov if you agree, please add them to the branch in this PR.

pashkinelfe:
@akorotkov I've added tests for some page-skip-scan logic and removing of a redundant check (covered by one of new tests) https://github.com/orioledb/orioledb/tree/page_level_diff_undo_records-add2  Please look at them and merge these changes to this PR if you agree

pashkinelfe (approved):

Issue #830, opened 2026-04-20
Handle `IS_DEV` guard and `inspect_sys_tree_structure` helper properly


za-arthur:
Problem:
- `inspect_sys_tree_structure` is a useful C-side debugging helper that has been reused informally multiple times for sys tree inspection during debugging.
- The `#ifdef IS_DEV` guard is currently a no-op since `IS_DEV` is not passed to C compilation flags (only used by Makefile for `.sql` file generation).

Proposed Work:
- Remove the `IS_DEV` guard so helpers are always compiled in, or
- Properly wire `-DIS_DEV` into C compilation flags for test/dev builds
and:
- Expose helpers as SQL functions for consistency with other debug utilities

Reference: https://github.com/orioledb/orioledb/pull/815#discussion_r3111095637

PR #912, opened 2026-06-18
Handle `IS_DEV` guard and `inspect_sys_tree_structure` helper properly

pgnickb:
- added `-DIS_DEV` flag to `PG_CPPFLAGS` when it is passed to the Makefile
- fixed all relevant places where the flag is or should be used in C code
- fixed .dockerignore being out of sync with .gitignore.
- bumped extension version to 1.9 (no user-visible changes in this one)

Addresses #830


src/catalog/sys_trees.c
```diff
@@ -560,6 +560,21 @@ inspect_sys_tree_structure(int systree, int depth)
...
 	return DatumGetTextP(res);
 }
+
+/*
+ * SQL wrapper to expose the helper for consistency with other debug utilities.
+ */
+PG_FUNCTION_INFO_V1(inspect_sys_tree_structure_sql);
+
+Datum
+inspect_sys_tree_structure_sql(PG_FUNCTION_ARGS)
```
pashkinelfe: inspect_sys_tree_structure() has no existing callers, so IMO it could be modified to output Datum and be called from sql interface directly without inspect_sys_tree_structure_sql wrapper

sql/orioledb--1.8--1.9_dev.sql
```diff
@@ -0,0 +1,11 @@
+/* contrib/orioledb/sql/orioledb--1.8--1.9_dev.sql */
+
+-- complain if script is sourced in psql, rather than via CREATE EXTENSION
+\echo Use "ALTER EXTENSION orioledb UPDATE TO '1.9'" to load this file. \quit
+
+-- While most OrioleDB functions use the "orioledb_" prefix, it is
+-- omitted here for brevity since this is an internal, non-user-facing helper.
+CREATE FUNCTION inspect_sys_tree_structure(systree int, depth int)
```
pashkinelfe: Any reason to call inspect_sys_tree_structure_sql -> inspect_sys_tree_structure -> orioledb_sys_tree_structure instead of calling orioledb_sys_tree_structure() directly?

pgnickb: no. Was just following the linked issue's proposal. Will remove as per offline discussion.

src/btree/io.c
```diff
@@ -1183,6 +1183,8 @@ get_free_disk_extent_copy_blkno(BTreeDescr *desc, off_t page_size,
 	return FileExtentIsValid(*extent);
 }
 
+#ifdef IS_DEV
+
 /* Functions for eviction_page_checkpoint_numbers test included under IS_DEV */
 PG_FUNCTION_INFO_V1(reset_read_page_checkpoint_stats);
```
pashkinelfe, line 1189: We previously agreed that functions already existed in .so should not be removed. Following functions were in previous **prod** builds and wil be absent in the future prod builds.

pashkinelfe: They are not referenced from prod SQL, so maybe still good

src/catalog/sys_trees.c
```diff
@@ -560,6 +560,7 @@ inspect_sys_tree_structure(int systree, int depth)
 
 	return DatumGetTextP(res);
 }
+
```
pashkinelfe: not needed

pgnickb: Yeah some artifacts. I'll clean it up in a bit

pashkinelfe (approved):

PR #919, opened 2026-06-22
FETCH-mode partial B-tree iterator (backward scans + fastpath downlink search)

akorotkov:
## Overview

Implements a **FETCH-mode partial B-tree iterator**: the iterator reads only the
parts of each page it actually needs (partial reads) instead of copying whole
page images, and switches to full IMAGE reads after a few pages. Backward scans
are supported via a separate parent-partial state, and intra-page navigation
uses a cache-friendly fastpath downlink search.

## What's in here

- **Iterator FETCH mode + backward scans.** Defaults to FETCH (partial leaf
  reads), switches to IMAGE after the 3rd page. Backward scans keep a stable
  per-page lokey (`leafLokey`) and a separate parent-partial image so
  `find_left_page`/`find_right_page` can navigate siblings.

- **Fastpath downlink search** (`fastpath.c`): walks a fixed-stride array that
  mirrors the page layout to resolve a downlink without copying chunks, with
  lock-free validation (state-bits change count **and** `pageChangeCount`, so a
  reused/evicted/merged page is detected). Now also supports **DESC-ordered key
  columns** — the array-search routines take a `descending` flag and bounds are
  expressed as storage directions, so DESC and `NULLS FIRST/LAST` resolve
  correctly.

- **Load the hikeys chunk on demand** (`loadHikeys = !fastpath`): the
  chunk-descriptor array is loaded only when the iterator actually crosses a
  chunk boundary / steps to a sibling, not eagerly on every leaf.

- **`convert_fastpath_parent_to_img()`**: when the fastpath located the parent's
  downlink, materialize the parent into `parentImg` by *finishing the partial
  read the descent already started* (`partial_load_hikeys_chunk` +
  `partial_load_chunk`) rather than re-reading the page — re-snapshotting could
  leave `parentImg` inconsistent with the downlink the fastpath already chose.
  Forward scans defer this until `find_right_page()` actually needs it.

- **Defensive assert** in `try_copy_page()` (assert builds only): a copied
  shared-pool page must physically belong to the descended tree (`datoid` +
  `relnode`).

## Testing

- `regresscheck` 45/45, `isolationcheck` 30/30.
- `test_replication_hot_read` (hot standby) green.
- 64× `valgrind_1` matrix green (temporary CI fan-out, since removed).
- New regression test `fastpath_desc` cross-checks DESC / DESC-NULLS-FIRST /
  mixed `(ASC, DESC)` composite index scans against a heap oracle.

### DIFF
--- a/src/btree/iterator.c
+++ b/src/btree/iterator.c
@@ -86,11 +86,10 @@ struct BTreeIterator
 	bool		curKeyReturned;
 
 	/*
-	 * The scan's original start key/kind.  Contract: the caller must keep
-	 * the key alive and unmodified for the iterator's lifetime; the
-	 * !curKeySet branch of iterator_refind_partial_leaf() reads it to re-find
-	 * the scan start when a partial read fails before any tuple has been
-	 * returned.
+	 * The scan's original start key/kind.  Contract: the caller must keep the
+	 * key alive and unmodified for the iterator's lifetime; the !curKeySet
+	 * branch of iterator_refind_partial_leaf() reads it to re-find the scan
+	 * start when a partial read fails before any tuple has been returned.
 	 */
 	void	   *startKey;
 	BTreeKeyType startKind;
@@ -934,8 +933,8 @@ iterator_refind_partial_leaf(BTreeIterator *it)
 	 * no tuple on the leaf is >= curKey -- e.g. curKey got deleted and the
 	 * leaf's max is now strictly less than curKey.  For forward we leave the
 	 * locator invalid so the caller's outer loop steps to the right sibling.
-	 * For backward the largest tuple <= curKey on this leaf is the last
-	 * item, so decrement to it; mirrors the seed and !curKeySet paths.
+	 * For backward the largest tuple <= curKey on this leaf is the last item,
+	 * so decrement to it; mirrors the seed and !curKeySet paths.
 	 */
 	if (!BTREE_PAGE_LOCATOR_IS_VALID(context->img, loc))
 	{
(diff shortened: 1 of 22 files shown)


test/expected/fastpath_desc.out
```diff
@@ -0,0 +1,187 @@
...
+	(SELECT a, b, id FROM o WHERE a = 42 AND b >= 10000 EXCEPT ALL SELECT a, b, id FROM h WHERE a = 42 AND b >= 10000)
+	UNION ALL
+	(SELECT a, b, id FROM h WHERE a = 42 AND b >= 10000 EXCEPT ALL SELECT a, b, id FROM o WHERE a = 42 AND b >= 10000)
+) d;
+ comp_range_mismatch 
+---------------------
+                   0
+(1 row)
+
+-- Ordered samples: the index drives the order (forward = DESC, backward = ASC).
```
homper: may be useful to display plans here, so it will show that it scans using expected indices

akorotkov: Added `EXPLAIN (COSTS OFF)` before the ordered-sample queries — the plans now show each one is served by the expected index (forward or backward index-only scan), e.g. `Index Only Scan using o_v_desc` / `Index Only Scan Backward using o_v_desc` / `o_v_desc_nl` / `o_ab`, never a seqscan. I also made the second single-column index `DESC NULLS LAST` (it was identical to the first) so each `ORDER BY ... NULLS ...` maps unambiguously to one index.

src/btree/iterator.c
```diff
@@ -1029,8 +1319,34 @@ btree_iterate_raw_internal(BTreeIterator *it, void *end, BTreeKeyType endKind,
...
 			BTREE_PAGE_READ_LEAF_ITEM(*tupHdr, result, context->img, loc);
-			IT_NEXT_OFFSET(it, loc);
+
+			/*
+			 * Remember this position in case the page changes later.  This
+			 * tuple is being consumed, so mark it returned: a later re-find
+			 * resumes after it.
+			 */
+			copy_fixed_key(context->desc, &it->curKey, result);
+			it->curKeySet = true;
```
pashkinelfe: Is combination of `it->curKeySet = true` with `O_TUPLE_SET_NULL(result);` (L1360) valid in a case the caller continues to use iterator (not drops it immedialtey)?

pashkinelfe:
@akorotkov I've added a couple of whitebox tests and fixes in separate branch https://github.com/orioledb/orioledb/tree/iterator-fetch-add  If you agree, please merge them into this PR

pashkinelfe:
Force-pushed https://github.com/orioledb/orioledb/tree/refs/heads/iterator-fetch-add branch Changes to new regress tests that prevented them from passing in CI.

pashkinelfe:
One more force-push to branch https://github.com/orioledb/orioledb/tree/refs/heads/iterator-fetch-add

pashkinelfe:
Added commit https://github.com/orioledb/orioledb/pull/919/changes/86b3052fa394f69fedfef34548006f73f4951d88 mirroring logic into btree_iterate_raw_internal

pashkinelfe (approved):

PR #923, opened 2026-06-24
Adjust logic of CI tests to allow manual dispatch and reduce extra runs

pgnickb:
Currently CI runs trigger on a PR and on a Push to any branch. This duplicates the runs when PR is created wasting resources and complicating run result analysis.

Remove CI runs for `Push` events on `test`, `pgindent` and `static` for non-`main` branches and instead only run on PRs and allow manual dispatch.

This way developers can manually run CI on dev branches without needing a PR.

### DIFF
--- a/.github/workflows/check.yml
+++ b/.github/workflows/check.yml
@@ -1,8 +1,6 @@
 name: check
 
-on:
-  push:
-  pull_request:
+on: [workflow_dispatch, pull_request]
 
 jobs:
   matrix:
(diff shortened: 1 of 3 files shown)


.github/workflows/check.yml
```diff
@@ -1,8 +1,6 @@
 name: check
 
-on:
-  push:
-  pull_request:
+on: [workflow_dispatch, pull_request]
```
homper: IMO, all of them also need `push: ["main"]`?

PR #925, opened 2026-06-26
Implementation of moving database files

ereutt:
Implement moving database files via ALTER DATABASE ... SET TABLESPACE command.
For moving using file system directory copy. In order to support replication this DDL command new WAL record introduced.

### DIFF
--- a/include/recovery/wal.h
+++ b/include/recovery/wal.h
@@ -189,6 +189,13 @@ typedef struct
 	uint8		origin_lsn[sizeof(XLogRecPtr)];
 } WALRecOriginInfo;
 
+typedef struct
+{
+	uint8		recType;
+	uint8		datid[sizeof(Oid)];
+	uint8		src_tblspc[sizeof(Oid)];
+	uint8		dst_tblspc[sizeof(Oid)];
+} WALRecDbCopy;
 
 #define LOCAL_WAL_BUFFER_SIZE	(8192)
 #define ORIOLEDB_WAL_PREFIX	"o_wal"
@@ -205,6 +212,7 @@ extern void add_switch_logical_xid_wal_record(TransactionId logicalXid_top, Tran
 extern void add_savepoint_wal_record(SubTransactionId parentSubid,
 									 TransactionId prentLogicalXid);
 extern void add_rollback_to_savepoint_wal_record(SubTransactionId parentSubid);
+extern void add_database_copy_wal_record(Oid dboid, Oid src_tblspc, Oid dst_tblspc);
 extern bool local_wal_is_empty(void);
 extern XLogRecPtr flush_local_wal(bool isCommit, bool withXactTime);
 extern XLogRecPtr wal_commit(OXid oxid, TransactionId logicalXid,
--- a/src/btree/io.c
+++ b/src/btree/io.c
@@ -3104,6 +3104,10 @@ retry:
 	return evict ? OWalkPageEvicted : OWalkPageWritten;
 }
 
+/*
+ * Recursively write pages in the tree. Stop reqursion if we reach maxLevel,
+ * when it has non-negtive value. To write all pages, set maxLevel to -1.
+ */
 static bool
 write_tree_pages_recursive(UndoLogType undoType,
 						   OInMemoryBlkno blkno, uint32 loadId,
@@ -3164,7 +3168,7 @@ write_tree_pages_recursive(UndoLogType undoType,
 										  maxLevel,
 										  evict);
 
-	if (level <= maxLevel)
+	if (level <= maxLevel || maxLevel == -1)
 	{
 		while (true)
 		{
@@ -3179,7 +3183,7 @@ write_tree_pages_recursive(UndoLogType undoType,
 	return true;
 }
 
-static void
+void
 write_tree_pages(BTreeDescr *desc, int maxLevel, bool evict)
 {
 	o_btree_load_shmem(desc);
--- a/src/catalog/ddl.c
+++ b/src/catalog/ddl.c
@@ -23,6 +23,7 @@
 #include "catalog/o_sys_cache.h"
 #include "storage/lockdefs.h"
 #include "tableam/descr.h"
+#include "storage/copydir.h"
 #include "tableam/operations.h"
 #include "catalog/pg_am.h"
 #include "tableam/toast.h"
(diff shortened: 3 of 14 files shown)


src/btree/io.c
```diff
@@ -3403,6 +3403,8 @@ perform_writeback(IOWriteBack *writeback)
 				char	   *filename;
 
 				filename = btree_filename(cur.key, segno, chkpNum);
+				if (cur.key.oids.datoid > 1000)
```
homper: Debugging log leftover I assume

include/recovery/wal.h
```diff
@@ -205,6 +212,7 @@ extern void add_switch_logical_xid_wal_record(TransactionId logicalXid_top, Tran
 extern void add_savepoint_wal_record(SubTransactionId parentSubid,
 									 TransactionId prentLogicalXid);
 extern void add_rollback_to_savepoint_wal_record(SubTransactionId parentSubid);
+extern void add_o_database_copy_record(Oid dboid, Oid src_tblspc, Oid dst_tblspc);
```
homper:
```suggestion
extern void add_database_copy_wal_record(Oid dboid, Oid src_tblspc, Oid dst_tblspc);
```

src/catalog/ddl.c
```diff
@@ -4335,6 +4355,110 @@ orioledb_object_access_hook(ObjectAccessType access, Oid classId, Oid objectId,
...
+					/*
+					* The directory exists but is empty. We must remove it before using
+					* the copydir function.
+					*/
+					if (rmdir(dst_dbpath) != 0)
+						elog(ERROR, "could not remove directory \"%s\": %m",
+							dst_dbpath);
+				}
+
+				PG_ENSURE_ERROR_CLEANUP(o_movedb_failure_callback,
```
homper, line 4431: Not sure how to reproduce such situation yet, but isn't it possible that it will send wal record, fail somewhere and then it will not be "rolled back" on replica?

ereutt: Actually I can't reproduce this as well, but bring this code from vanilla move database handling

src/tableam/descr.c
```diff
@@ -1284,6 +1284,44 @@ o_invalidate_descrs(Oid datoid, Oid reloid, Oid relfilenode)
 	o_stop_saving_inval_messages(was_saving);
 }
 
+List* o_evict_db_relations(Oid datoid)
```
za-arthur:
```suggestion
List*
o_evict_db_relations(Oid datoid)
```

src/catalog/ddl.c
```diff
@@ -4753,6 +4877,68 @@ o_createdb(ParseState *pstate, const CreatedbStmt *stmt)
 	return result;
 }
 
+static void o_check_movedb(const AlterDatabaseStmt *stmt, movedb_params *movedb)
```
za-arthur:
```suggestion
static void
o_check_movedb(const AlterDatabaseStmt *stmt, movedb_params *movedb)
```

src/btree/io.c
```diff
@@ -3156,7 +3156,7 @@ write_tree_pages_recursive(UndoLogType undoType,
 										  maxLevel,
 										  evict);
 
-	if (level <= maxLevel)
+	if (level <= maxLevel || maxLevel == -1)
```
za-arthur, line 3171: Why was this change added? In what cases can `maxLevel` be `-1`?

src/catalog/ddl.c
```diff
@@ -4335,6 +4355,110 @@ orioledb_object_access_hook(ObjectAccessType access, Oid classId, Oid objectId,
...
+			OidIsValid(o_movedb_data.src_tsoid) &&
+			OidIsValid(o_movedb_data.dest_tsoid) &&
+			o_tables_num(o_movedb_data.dest_dboid))
+		{
+			char *src_dbpath = NULL;
+			char *dst_dbpath = NULL;
+			char *dst_prefix = NULL;
+			DIR  *dir;
+			struct dirent *xlde;
+			bool skipMoving = false;
```
za-arthur:
```suggestion
			bool skipMoving = true;
```

src/catalog/ddl.c
```diff
@@ -4335,6 +4355,110 @@ orioledb_object_access_hook(ObjectAccessType access, Oid classId, Oid objectId,
...
+					break;
+				}
+
+				FreeDir(dir);
+				skipMoving = !hasFiles;
+			}
+			else
+			{
+				skipMoving = true;
+			}
```
za-arthur:
```suggestion
```

src/catalog/o_tables.c
```diff
@@ -1228,6 +1237,71 @@ o_tables_drop_all(OXid oxid, CommitSeqNo csn, Oid database_id)
...
+		o_invalidate_descrs(ix_table->oids.datoid, ix_table->oids.reloid, ix_table->oids.relnode);
+	}
+	if (table_moved)
+	{
+		o_tables_update(o_table, move_arg->oxid, move_arg->csn);
+	}
+	o_tables_after_update(o_table, move_arg->oxid, move_arg->csn);
+}
+
+void o_tables_move_all(OXid oxid, CommitSeqNo csn, Oid database_id, Oid old_tspcoid, Oid new_tspcoid)
```
za-arthur:
```suggestion
void
o_tables_move_all(OXid oxid, CommitSeqNo csn, Oid database_id, Oid old_tspcoid, Oid new_tspcoid)
```

src/recovery/wal_reader.c
```diff
@@ -299,6 +300,19 @@ wal_parse_rec_modify(WalReaderState *r, WalRecord *rec)
 	return WALPARSE_OK;
 }
 
+/* Parser for WAL_REC_DATABASE_COPY */
+WalParseResult wal_parse_rec_dbcopy(WalReaderState *r, WalRecord *rec)
```
za-arthur:
```suggestion
WalParseResult
wal_parse_rec_dbcopy(WalReaderState *r, WalRecord *rec)
```

src/btree/io.c
ereutt: Initially this evict implementation used only for evict relation pages via sql-interface, which has max_level as input parameter. I use this eviction code on replica side to evict pages relative to moving database, all pages must be evicted -1 for max_level here means "evict all levels"

za-arthur: Please add comments to explain this "magic number".

src/catalog/ddl.c
ereutt: Vanilla implementation has two separate way records for moving database: copy files and remove old. May be I need do something like this on replica node:
- WAL record about start moving: evict pages + copy files
- WAL records which update system tree with new table space
- WAL record about end moving: update evicted data and remove old directory.

What do you think about rolling back in this schema?

But I still couldn't come up with reproduce case for reproduction rollback situation

homper: Hm, after another look, probably current state is good enough for now, if we catch or come up with possible situation in the future, will probably go this way

PR #934, opened 2026-06-30
Don't try to save extended hash proc if it not exist for type

### LINKED ISSUE #928: "cache lookup failed for amproc" error when using ulid column type
I use the latest version of [pgx_ulid](https://github.com/pksunkara/pgx_ulid) to support the ulid type in my db, but when trying to create a table using orioledb, I get the mentioned error. To confirm it I just tried to create two tables,
`CREATE TABLE test_ulid_heap (id ulid PRIMARY KEY) USING heap;`
and
`CREATE TABLE test_ulid_oriole (id ulid PRIMARY KEY) USING orioledb;`
and the second one couldn't be created

homper:
Fixes #928

Also thought that I had wrong assumption and we don't need default hash function for types, but it is still needed for calling lookup_type_cache during recovery

### DIFF
--- a/src/catalog/o_sys_cache.c
+++ b/src/catalog/o_sys_cache.c
@@ -1377,6 +1377,7 @@ o_cache_type_opclasses(Oid datoid, Oid typoid,
 		Oid			sys_datoid;
 		Oid			hash_opf;
 		Oid			hash_opintype;
+		Oid			hash_extended_proc = InvalidOid;
 
 		hash_opf = get_opclass_family(hash_opclass);
 		hash_opintype = get_opclass_input_type(hash_opclass);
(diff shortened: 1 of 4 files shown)


pashkinelfe (approved):

pashkinelfe:
Is it possible to also commit test similar to #928 but without using other extensions?

src/catalog/o_sys_cache.c
```diff
@@ -1391,9 +1392,14 @@ o_cache_type_opclasses(Oid datoid, Oid typoid,
 		o_amproc_cache_add_if_needed(datoid, hash_opf, hash_opintype,
 									 hash_opintype, HASHSTANDARD_PROC,
 									 insert_lsn, NULL);
-		o_amproc_cache_add_if_needed(datoid, hash_opf, hash_opintype,
-									 hash_opintype, HASHEXTENDED_PROC,
-									 insert_lsn, NULL);
+		hash_extended_proc = get_opfamily_proc(hash_opf,
+											   hash_opintype,
+											   hash_opintype,
+											   HASHEXTENDED_PROC);
```
za-arthur, line 1398: I think it make sense to add new argument `missingOk` (similar to other postgres and orioledb functions), which will ignore key if it doesn't exist and if `missingOk == true`.

pashkinelfe: The downside of this, if I understand you correctly, would be modifying PG patchsets with just moderate benefit gained

za-arthur: I proposed to add `missingOk` argument into `*_cache_add_if_needed` function, pass it through and check it in `*_cache_fill_entry` function. These are not Postgres internal functions. It can skip adding new entry instead of raising an error:
https://github.com/orioledb/orioledb/blob/main/src/catalog/o_amproc_cache.c#L77-L77

homper: @za-arthur wait, no, I still not sure if it conforms with missing_ok in posgres, because it used mostly when getting or dropping something, not for inserts; and also missing is not ok when caching metadata, caching only if it exists for type makes more sense IMO

---
Hm, and also, probably, it needs to check for existing of regular functions for btree and hash ams and throw sane error for them if not as a safeguard for weird wrong types created by users

homper:
> Is it possible to also commit test similar to #928 but without using other extensions?

Yes, added test with such type

PR #935, opened 2026-06-30
Implement multi-insert optimization

pashkinelfe:
Drain ordered rows into one locked primary leaf

orioledb_multi_insert previously did one find_page + lwlock cycle per row. For ordered bulk loads (COPY into a CTID-PK table, or any sorted explicit-PK COPY) most rows share a primary leaf with their neighbors, so the per-row descent and lock acquisition are pure overhead.

Now o_tbl_multi_insert() is the four-phase:

  Phase 1  per-slot ctid + bridge ctid + in-row TOAST + form primary tuple
           + key bound.

  Phase 2  optimistically assume keys[] are monotone and just verify with
           an O(n) scan; the common cases (CTID-PK by construction,
           ordered explicit-PK COPY) hit this fast path.  Only on the first
           out-of-order pair do we fall back to qsort_arg-ing by key,
	   and materializing sorted views of tuples / tuplens / keyptrs / cb_args.
           Sorting is required because the leaf probe detects "key past hikey"
	   but not "key before lokey" (lokey lives in the parent, not the leaf),
	   so non-monotone input could corrupt downlinks.

  Phase 3  stream sorted keys through primary leaves, holding each
           leaf's lwlock for as many adjacent keys as fit.  Two safety
           bounds on the per-batch row-undo reservation:

           - Tail slack of one max-row size.  get_undo_record() deducts
             size again on its circular-buffer boundary-cross retry, so
             an exact nitems * per_row reservation underflows on the last
             call when one allocation straddles the wrap.

           - Cap at 2 * O_MAX_UNDO_RECORD_SIZE -- the per-backend share
             the buffer is sized for (see undo_shmem_needs).  Without it,
             max_procs concurrent multi_insert backends could collectively
             reserve more than the buffer holds.  Larger inputs are split
             across successive iterations; a Fits-with-i<ntuples exit
             branch covers the cap path, since it isn't a bail to
             slow-path.

           NoFit / Duplicate -> slow-path one row via o_tbl_index_insert
           and resume; HikeyCrossed -> re-find the next leaf.  Non-orioledb
           input slots fall back to per-row o_tbl_insert.

  Phase 4  per-slot TOAST values + WAL (order-independent).

orioledb_multi_insert() routes to the batched path when orioledb.debug_disable_multi_insert is off (default) and ntuples > 1

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -120,6 +120,7 @@ REGRESSCHECKS = btree_sys_check \
 				ioc \
 				iterator \
 				joins \
+				multi_insert \
 				nulls \
 				opclass \
 				parallel_scan \
@@ -212,6 +213,7 @@ TESTGRESCHECKS_PART_2 = test/t/checkpoint_concurrent_test.py \
 						test/t/logical_test.py \
 						test/t/logical_xid_subxacts_test.py \
 						test/t/merge_into_test.py \
+						test/t/multi_insert_test.py \
 						test/t/not_supported_yet_test.py \
 						test/t/pg_dump_restore_test.py \
 						test/t/parallel_test.py \
--- a/include/btree/insert.h
+++ b/include/btree/insert.h
@@ -16,6 +16,7 @@
 
 #include "btree.h"
 #include "btree/find.h"
+#include "btree/modify.h"
 
 extern void o_btree_split_fix_and_unlock(BTreeDescr *descr,
 										 OInMemoryBlkno left_blkno);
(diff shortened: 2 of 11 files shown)


pashkinelfe:
Rebased on fresh main and made minor renaming (force-pushed)

akorotkov:
Looks correct. The possible further optimizations include:

 * Multi-insert for secondary indexes (requires core patching)
 * Multi-insert for toast

akorotkov (approved):

PR #938, opened 2026-07-01
Optimize array key lookup in is_tuple_valid() with binary search

fatih-akatlar:
Replace O(n) linear scan with O(log n) binary search for checking tuple membership in sorted array key elements.  Also hoist the coercibility check and comparator selection out of the loop, avoiding repeated IsBinaryCoercible() calls per tuple.

Assisted-by: Crush:glm-5.2


akorotkov:
Similar fix merged as ee5906dec8a92966a6f230b63cb075c8f32ae53f.

PR #940, opened 2026-07-02
Don't wait on in-progress re-inserter of a concurrently-deleted key

akorotkov:
## Problem

`SELECT ... FOR UPDATE` could block on the **in-progress transaction that
re-inserted a tuple with the same primary key** whose `DELETE` had already
**committed after the waiter's snapshot**.

OrioleDB links a delete plus a later same-key re-insert into a single undo
chain. When resolving a lock conflict, the code walked up to the in-progress
top version and waited on it, instead of recognizing the committed delete
below it in the chain and treating the row as gone. Under `REPEATABLE READ` /
`SERIALIZABLE` this should be an immediate serialization error, not a wait on
a re-insert that is invisible to us.

Deterministic repro (3 sessions): s1 fixes an RR snapshot while key 5 is alive;
s2 deletes key 5 and commits; s3 re-inserts key 5 (in-progress); s1's
`... ORDER BY id LIMIT 1 FOR UPDATE` blocks on s3 until timeout. Expected:
immediate `could not serialize access due to concurrent delete`.

## Fix

In `o_btree_modify_handle_conflicts`, before `wait_for_tuple` and **only for
`BTreeOperationLock`**, walk the version chain via the new
`tuple_is_deleted_for_csn()`: if the first committed version is a delete, or
the undo chain breaks before reaching our snapshot's csn, raise a
serialization error (RR/SERIALIZABLE) or return `OBTreeModifyResultDeleted`
instead of blocking on the invisible re-inserter.

Uses `oxid_get_csn()` for per-version csn (not `XACT_INFO_MAP_CSN`, which
returns 0 for in-progress).

## Testing

- New isolation test `forupdate_reinsert` (fails before the fix — blocks;
  passes after — serialization error).
- `isolationcheck` 31/31 and `regresscheck` 46/46 pass locally.
- Validated on an optimized (`--disable-cassert`) build in **both** the FETCH
  and IMAGE iterator modes.

## Notes / open items (draft)

- Scope is limited to `BTreeOperationLock`. Extending to `Update` reintroduced
  `concurrent_update_delete` failures (update-over-deleted / pkey-change
  semantics) and a WAL assert, so it is left out for now.
- The chain walk fires on every in-progress `FOR UPDATE` lock conflict and adds
  ~13% overhead to the TPCC delivery convoy under the FETCH iterator (no cost
  under IMAGE). It could be gated to `IsolationUsesXactSnapshot()` (TPCC
  delivery is READ COMMITTED, where the spurious wait is a liveness rather than
  correctness issue) to remove that overhead — not done here pending review.
- This bug is not the cause of the FETCH-vs-IMAGE TPCC throughput difference
  (new_order never reuses a primary key, so same-key delete+reinsert does not
  occur there); it is a separate correctness bug.

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -149,6 +149,7 @@ ISOLATIONCHECKS = bitmap_hist_scan \
 				  btree_scan \
 				  concurrent_update_delete \
 				  fkeys \
+				  forupdate_reinsert \
 				  included \
 				  insert_fails \
 				  ioc_deadlock \
--- a/include/btree/undo.h
+++ b/include/btree/undo.h
@@ -208,6 +208,7 @@ extern UndoLocation make_merge_undo_image(BTreeDescr *desc, Pointer left,
 										  Pointer right, CommitSeqNo imageCsn);
 extern bool page_op_drops_tuple(BTreeDescr *desc, Pointer p,
 								CommitSeqNo imageCsn);
+extern bool tuple_is_deleted_for_csn(UndoLogType undoType, BTreeLeafTuphdr *pageTuphdr, CommitSeqNo my_csn);
 extern bool row_lock_conflicts(BTreeLeafTuphdr *pageTuphdr,
 							   BTreeLeafTuphdr *conflictTupHdr,
 							   UndoLogType undoType,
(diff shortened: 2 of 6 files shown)


pashkinelfe:
@akorotkov I've added a couple of proposal commits to a separate branch https://github.com/orioledb/orioledb/tree/fix-forupdate-wait-on-reinserted-key-add1  Merge them into this PR if you agree

PR #647, opened 2025-11-18
Added lockstep scan for arrays and fixed coercible calls

Serge-sudo:
During Array scans like (SELECT ... WHERE id IN (1,3,4...)) orioledb does full tree descend for each element of the array. This is inefficient when the array is large and the elements are close to each other in the index. This commit adds lockstep scan for arrays, which descends the tree only once for all elements of the array and uses iterators.

Also fixed the problem with too many coercible calls during index scans (perf showed that it was having significant impact on performance).
Even if the key was not coercible, orioledb was trying to check that again and again during index scans.

PR #945, opened 2026-07-04
Back the bitmap scan with an adaptive radix tree; support composite PKs

akorotkov:
## Summary

Reimplements the orioledb bitmap-scan key set on top of PostgreSQL's
adaptive radix tree (instead of a bespoke red-black tree) and extends
bitmap scans to **composite primary keys** — including row-array `IN()`
on the primary key, the shape that degraded to whole-warehouse scans in
TPCC delivery.

## What's here

- **`include/lib/o_radixtree.h`** — vendored copy of PG's `lib/radixtree.h`
  extended with an optional `RT_KEY_SIZE`: when defined, the tree is keyed
  by a fixed-length big-endian byte string instead of a `uint64`, keeping
  the adaptive-node machinery untouched. Vendoring also supplies the
  template on PG16.
- **`key_bitmap.c`** — the densified uint64 chunk map now lives in a radix
  tree (behaviour-equivalent to the old rbtree); a second, fixed-key mode
  stores whole order-preserving encoded keys for composite PKs. Ordered
  seeks come from a sorted key array built lazily once the bitmap stops
  being mutated. Handle type is the opaque `OKeyBitmap`.
- **`bitmap_scan.c`** — order-preserving encode/decode of integer PK
  columns (int2/int4/int8, ascending, INCLUDE columns excluded);
  `o_keybitmap_pk_mode()` classifies a PK (uint64 / fixed / ineligible)
  and is shared by planner and executor. Bitmap index scans over the
  primary index itself are now supported, so `(a,b,…) IN (...)` on a
  composite PK is planned as a `BitmapOr` of per-tuple primary-index
  scans — on a large table ~20x faster than the common-prefix scan, while
  selective single-key quals still use an index scan on default costs.
- **`radix_selftest.c`** — dev-only SQL self-tests for the fixed-key radix
  tree, the encoding, and the fixed-key set.
- **`bitmap_scan`** regression test gains a composite-PK row-array `IN()`
  case checked against a sequential scan.

`make regresscheck` 46/46 and `make isolationcheck` 31/31 pass locally
(PG17). All 7 touched regression tests changed only *plans*, not results.

## Follow-ups (to polish via CI)

- Only the **PG17** expected variants were regenerated locally
  (`*_1.out` / `stats.out`); the PG16 and PG18 variants of the same 7
  tests will show the identical plan changes and should be updated from CI.
- **Cost refinement:** on very small / unanalyzed tables the bitmap-heap
  path can still be chosen over an index scan because core
  `cost_bitmap_heap_scan` models a random heap fetch rather than
  orioledb's primary-index scan. Making that exact needs a patched-PG
  cost change.

---

## Update — read efficiency & array-scan reuse

Further commits on this branch make the scan itself cheaper, not just the plan choice.

### Partial (FETCH-mode) page reads
- Bitmap scans read only the needed chunks of a leaf and skip empty internal
  pages via bitmap-driven descent, instead of copying whole page images into
  local memory. Includes a fix for partial-state leaking across leaves (which
  re-emitted rows), a crash fix for a composite seek key over a covering
  primary key, and unifying the seq-scan `getNextKey` / `getNextPageKey`
  callbacks.

### Array-scan iterator reuse (`btree/iterator.c`, `tableam/index_scan.c`, `tableam/key_range.c`)
- An ordered `col = ANY(array)` / `col IN (...)` scan now reuses a single
  iterator — and its already-loaded page — across array elements, instead of
  re-descending from the root for every element. Adds
  `o_btree_iterator_advance()` (range path) and
  `o_btree_find_tuples_start()` / `o_btree_find_tuples_continue()` (exact-key
  path). Inspired by #647.
- Caches the "not coercible" fact for bound values
  (`O_VALUE_BOUND_NON_COERCIBLE`) to avoid repeated `IsBinaryCoercible()`
  calls that `perf` showed were significant.
- **Optimistic next-item check:** when the next matching tuple is the item the
  scan is already positioned on (adjacent / dense arrays), skip the per-key
  binary search; sparse gaps, chunk crossings and backward scans fall through
  to the search unchanged.

### Streaming primary bitmap scans (`tableam/bitmap_scan.c`, `tableam/key_bitmap.c`)
- A single primary `BitmapIndexScan`, or a `BitmapOr` of only such scans (the
  composite-PK row-array `(a,b,c) IN (...)` shape TPCC delivery uses), is now
  executed by driving the primary index scan(s) directly and handing their
  **live** tuples to the output — instead of materializing a key bitmap and
  re-reading the primary tree. Full orioledb row identity (csn / hint / rowid)
  is preserved, so row locking, EPQ, foreign keys and toast behave exactly as
  the second-pass path, and the second pass over the primary tree is removed.
- A `BitmapOr`'s branches can yield the same pk (overlapping / duplicate
  row-tuples, which the planner does not dedup), so a dedup key bitmap carries
  an "already-emitted" bit (`o_keybitmap_emit()` / `o_keybitmap_emit_key()`).
  Any other shape (`BitmapAnd`, a non-primary or bridged branch) falls back to
  the existing key-bitmap build.
- This also softens the cost follow-up above: for a primary bitmap scan the
  expensive second pass no longer happens, so choosing the bitmap-heap path is
  much closer to an index scan in practice.

### Benchmarks (VM, release build, sysbench / go-tpc)
- sysbench `id IN (50001..50100)` on 1M rows: **+14–19%** vs the key-bitmap
  path, reaching **~86–92% of heap** (was ~74–80%).
- Composite-PK row-array `IN` (TPCC delivery shape): **+14%**, reaching
  **~93–97% of heap** (was ~82–85%).
- go-tpc TPC-C tpm (w=22): **flat, no regression** — the row-array `IN`
  delivery `SELECT` is a small, non-bottleneck fraction of the transaction mix.

`make regresscheck` 46/46 and `make isolationcheck` 31/31 still pass (PG17),
verified against a sequential-scan baseline for overlapping / duplicate `OR`
branches and composite-PK row-array `IN`.

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -76,6 +76,7 @@ OBJS = src/btree/btree.o \
 	   src/tableam/key_range.o \
 	   src/tableam/key_bitmap.o \
 	   src/tableam/operations.o \
+	   src/tableam/radix_selftest.o \
 	   src/tableam/scan.o \
 	   src/tableam/tree.o \
 	   src/tableam/vacuum.o \
--- a/ci/cppcheck-suppress
+++ b/ci/cppcheck-suppress
@@ -3,3 +3,5 @@ uselessAssignmentPtrArg:*
 incorrectStringBooleanError:*
 nullPointerRedundantCheck:*
 integerOverflowCond:*/atomics/generic.h
+# Vendored verbatim from PostgreSQL's lib/radixtree.h; not ours to audit.
+*:*/lib/o_radixtree.h
--- a/ci/static.sh
+++ b/ci/static.sh
@@ -10,7 +10,6 @@ cd orioledb
 if [ "$COMPILER" = "clang" ]; then
 	scan-build-$LLVM_VER --status-bugs --use-cc=clang-$LLVM_VER \
 		-disable-checker deadcode.DeadStores \
-		--exclude include/lib/o_radixtree.h \
 		make CLANG=clang-$LLVM_VER USE_PGXS=1 IS_DEV=1 USE_ASSERT_CHECKING=1 || status=$?
 
 elif [ "$COMPILER" = "gcc" ]; then
--- a/ci/static.sh
+++ b/ci/static.sh
@@ -10,6 +10,7 @@ cd orioledb
 if [ "$COMPILER" = "clang" ]; then
 	scan-build-$LLVM_VER --status-bugs --use-cc=clang-$LLVM_VER \
 		-disable-checker deadcode.DeadStores \
+		--exclude include/lib/o_radixtree.h \
 		make CLANG=clang-$LLVM_VER USE_PGXS=1 IS_DEV=1 USE_ASSERT_CHECKING=1 || status=$?
 
 elif [ "$COMPILER" = "gcc" ]; then
--- a/include/btree/btree.h
+++ b/include/btree/btree.h
@@ -38,16 +38,88 @@ typedef uint64 OTupleXactInfo;
 #define TOASTIndexNumber (0xFFFE)
 #define InvalidIndexNumber (0xFFFF)
 
+/*
+ * BTreeKeyType
+ *
+ * Defines the type of key used in B-tree operations.  The type affects how
+ * keys are compared, hashed, and interpreted during tree traversal and
+ * modification operations.
+ */
 typedef enum BTreeKeyType
 {
+	/*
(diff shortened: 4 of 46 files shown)


pashkinelfe (approved):

pashkinelfe:
Approved after agreement of all participants of Orioledb team in sync call

PR #949, opened 2026-07-07
Refactor tablespace directory traversal

fatih-akatlar:
Consolidates duplicated  `pg_tblspc`  directory iteration code into a single reusable API, and moves tablespace path routines out of the overloaded  `o_sys_cache`  module into dedicated  `o_tablespaces`  files.

### DIFF
--- a/src/s3/headers.c
+++ b/src/s3/headers.c
@@ -20,6 +20,7 @@
 
 #include "btree/btree.h"
 #include "btree/io.h"
+#include "catalog/o_tablespaces.h"
 #include "checkpoint/checkpoint.h"
 #include "s3/headers.h"
 #include "s3/worker.h"
--- a/src/s3/headers.c
+++ b/src/s3/headers.c
@@ -992,7 +992,8 @@ s3_headers_error_cleanup(void)
 typedef void (*IterateFilesCallback) (S3HeaderTag tag);
 
 static void
-iterate_tablespace_files(Oid tablespace, char *path, IterateFilesCallback callback)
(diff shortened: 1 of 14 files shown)


src/s3/headers.c
```diff
@@ -1058,77 +1060,15 @@ iterate_tablespace_files(Oid tablespace, char *path, IterateFilesCallback callba
...
-
-		tablespace = pg_strtoint64(file->d_name);
-		Assert(OidIsValid(tablespace));
-		path[0] = '\0';
-		pg_snprintf(path, MAXPGPATH,
-					PG_TBLSPC "/%s/" TABLESPACE_VERSION_DIRECTORY,
-					file->d_name);
-		if (lstat(path, &st) < 0)
-		{
-			ereport(PANIC,
```
fatih-akatlar, line 1097: The unified `o_tablespaces_foreach_prefix()` uses `ERROR` for all filesystem errors, matching the severity already used in `ddl.c` and `recovery.c`. However, the original `s3/headers.c` code used `PANIC` for the `lstat()` failure — so this is a behavioral change for the S3 checkpoint path.

That said, the original code was already inconsistent: it used `PANIC` for `lstat()` but `ERROR` for `readlink()` failures in the very next lines. This makes the `PANIC` look unintentional rather than a deliberate safety measure.

I can add an `elevel` parameter to `o_tablespaces_foreach_prefix()` to let each caller choose its severity if the `PANIC` is indeed intentional. But given the inconsistency, I suspect it isn't. Would appreciate confirmation either way.

homper:
Added commit from #1163 to also generalize database iteration

Issue #591, opened 2025-10-07
`DROP DATABASE` leaves artifacts of the dropped database


za-arthur:
`DROP DATABASE` should:
- delete relevant directories `<PGDATA>/orioledb_data/<db_oid>`, `<TABLESPACE>/orioledb_data/<db_oid>`, check that the directories are empty if necessary
- delete relevant entries from `SYS_TREES_*_CACHE`, `SYS_TREES_EXTENTS_OFF_LEN`, `SYS_TREES_EXTENTS_LEN_OFF`, `SYS_TREES_EVICTED_DATA`, `SYS_TREES_CHKP_NUM`, `SYS_TREES_MULTIRANGE_CACHE`, `SYS_TREES_TABLESPACE_CACHE`

PR #950, opened 2026-07-08
Remove empty orioledb_data directory after DROP DATABASE

fatih-akatlar:
`DROP DATABASE` removed table metadata and scheduled per-relation file cleanup via undo drop records, but left the per-database `orioledb_data/<dbOid>` directory behind.  Register a transaction callback that runs after undo processing (registered before `undo_xact_callback` so LIFO ordering places it last) to rmdir the empty directory from the default tablespace and all non-default tablespaces, under `pg_tblspc`.


za-arthur:
Relevant issue https://github.com/orioledb/orioledb/issues/591

PR #957, opened 2026-07-08
Fix the draining of grouping stack for many bitmapscans

bwyltcaxe:
Before this patch, EXPLAIN (FORMAT JSON) could fail with empty grouping stack if the query exists of many (>=7) bitmap scans.


akorotkov:
Can we add the test case reproducing the problem to this PR?

bwyltcaxe:
> Can we add the test case reproducing the problem to this PR?

Yes, I'll do it. I already have the manual reproducer, but I'll have to take a look at our tests about `EXPLAIN`.

PR #964, opened 2026-07-10
Introduce pages checksums

### LINKED ISSUE #956: Checksums for pages
In principle the implementation should follow the standard PG approach.

ORIOLEDB_PAGE_VERSION will very like be bumped to 2.

Older pages, without checksums must be still readable as well.

https://github.com/postgres/postgres/blob/master/src/include/storage/checksum_impl.h

cc @pashkinelfe

deadtrickster:
Uses FNV via pg_checksum_block, takes care of alignment. 
Skips validation for v1 pages. Zeroes compressed tails.
Doesn't do its own ERROR because load_page has some cleanup. 
Consequetly failed checksum validation can pick up strange errnos. 
^= blkno is deferred.

closes #956

### DIFF
--- a/doc/usage/configuration.mdx
+++ b/doc/usage/configuration.mdx
@@ -81,6 +81,14 @@ How OrioleDB handles a client request for SERIALIZABLE isolation. OrioleDB does 
 - `error`: reject SERIALIZABLE transactions with `ERRCODE_FEATURE_NOT_SUPPORTED` (the legacy behavior).
 - `repeatable_read`: treat SERIALIZABLE as `REPEATABLE READ` for OrioleDB tables only — no extra locks are taken, since OrioleDB's CSN-based snapshot already provides per-transaction stable reads. Heap tables in the same transaction continue to use PG's SSI as usual.
 
+### `orioledb.checksums_enabled`
+
+|             |    |
+| ----------- | -- |
+| **Default** | on |
+
+Read-only. Shows whether OrioleDB page checksums are enabled. OrioleDB always stores and verifies page checksums independently of the PostgreSQL `data_checksums` cluster setting.
+
 ## Advanced config options
 
 ### `orioledb.free_tree_buffers`
--- a/include/checkpoint/control.h
+++ b/include/checkpoint/control.h
@@ -32,7 +32,7 @@ typedef struct
  * Also on-the-flight conversion routine should be added to
  * check_checkpoint_control()
  */
-#define ORIOLEDB_CHECKPOINT_CONTROL_VERSION	1
+#define ORIOLEDB_CHECKPOINT_CONTROL_VERSION	2
 
 /*
  * To ensure correct reading of controlFileVersion, changes in struct layout
@@ -58,6 +58,11 @@ typedef struct
 	OXid		checkpointRetainXmax;
 	uint32		binaryVersion;
 	bool		s3Mode;
+
+	/*
+	 * orioledb_checksums_enabled
+	 */
+	bool		checksums_enabled;
 	/* CRC of all fields above. It should be last */
 	pg_crc32c	crc;
 } CheckpointControl;
--- a/include/orioledb.h
+++ b/include/orioledb.h
@@ -98,7 +98,7 @@
  * these values makes sense only within one ORIOLEDB_BINARY_VERSION value.
  */
 #define ORIOLEDB_VERSION "OrioleDB beta 16"
-#define ORIOLEDB_BINARY_VERSION 9
+#define ORIOLEDB_BINARY_VERSION 10
 #define ORIOLEDB_SYS_TREE_VERSION	1	/* Version of system catalog */
 #define ORIOLEDB_PAGE_VERSION		1	/* Version of binary page format */
 #define ORIOLEDB_COMPRESS_VERSION	1	/* Version of page compression (only
@@ -385,7 +385,8 @@ typedef struct
 	 * pages it should be used for conversion of uncompressed images
 	 */
 	uint8		page_version;
-	uint32		reserved1;
+	uint16		checkSum;
+	uint16		reserved1;
 	uint32		reserved2;
 } OrioleDBOndiskPageHeader;
 
@@ -470,6 +471,7 @@ extern bool orioledb_table_description_compress;
 extern char *max_bridge_ctid_string;
 extern BlockNumber max_bridge_ctid_blkno;
 extern bool orioledb_s3_mode;
+extern bool orioledb_checksums_enabled;
 extern int	s3_num_workers;
 extern int	s3_desired_size;
 extern int	s3_queue_size_guc;
--- a/src/btree/io.c
+++ b/src/btree/io.c
@@ -47,6 +47,8 @@
 #include "access/relation.h"
 #include "pgstat.h"
 #include "storage/bufmgr.h"
+#include "storage/checksum.h"
+#include "storage/checksum_impl.h"
 #include "utils/memutils.h"
 #include "utils/syscache.h"
 #include "funcapi.h"
@@ -76,6 +78,21 @@ typedef struct IOWriteBack
 	int			extentsAllocated;
 	TreeOffset *extents;
 } IOWriteBack;
+
+typedef union
+{
+	OrioleDBOndiskPageHeader phdr;
+	uint32		data[ORIOLEDB_BLCKSZ / (sizeof(uint32) * N_SUMS)][N_SUMS];
+} OrioleDBChecksummablePage;
+
+
+StaticAssertDecl(sizeof(OrioleDBChecksummablePage) == ORIOLEDB_BLCKSZ,
+				 "OrioleDBChecksummablePage size doesn't match ORIOLEDB_BLCKSZ");
+
+StaticAssertDecl(ORIOLEDB_BLCKSZ % (sizeof(uint32) * N_SUMS) == 0,
+				 "ORIOLEDB_BLCKSZ is not a multiple of sizeof(uint32) * N_SUMS");
+StaticAssertDecl(ORIOLEDB_COMP_BLCKSZ % (sizeof(uint32) * N_SUMS) == 0,
+				 "ORIOLEDB_COMP_BLCKSZ is not a multiple of sizeof(uint32) * N_SUMS");
 
 static IOWriteBack io_writeback =
 {
@@ -1249,6 +1266,69 @@ convert_orioledb_page_version(Pointer img)
 	elog(FATAL, "Page version conversion is not implemented");
 }
 
+static inline uint32
+oriole_checksum_block_len(uint32 blk_size)
+{
+	return (uint32) (blk_size / (sizeof(uint32) * N_SUMS));
+}
+
+/*
+ * Block checksum algorithm.  The page must be adequately aligned
+ * (at least on 4-byte boundary).
+ */
+static inline uint32
+oriole_checksum_block(const OrioleDBChecksummablePage *page, uint32 len)
+{
+	uint32		sums[N_SUMS];
+	uint32		result = 0;
+	uint32		i,
+				j;
+
+	/* initialize partial checksums to their corresponding offsets */
+	memcpy(sums, checksumBaseOffsets, sizeof(checksumBaseOffsets));
+
+	/* main checksum calculation */
+	for (i = 0; i < len; i++)
+		for (j = 0; j < N_SUMS; j++)
+			CHECKSUM_COMP(sums[j], page->data[i][j]);
+
+	/* finally add in two rounds of zeroes for additional mixing */
+	for (i = 0; i < 2; i++)
+		for (j = 0; j < N_SUMS; j++)
+			CHECKSUM_COMP(sums[j], 0);
+
+	/* xor fold partial checksums together */
+	for (i = 0; i < N_SUMS; i++)
+		result ^= sums[i];
+
+	return result;
+}
+
+static bool
+check_orioledb_page_checksum(OrioleDBOndiskPageHeader ondisk_page_header,
+							 char *buf, const char *label, uint32 len)
+{
+	uint16		computedCheckSum = 0;
+
+	((OrioleDBOndiskPageHeader *) buf)->checkSum = 0;
+
+	computedCheckSum = (uint16) ((oriole_checksum_block((const OrioleDBChecksummablePage *) buf, len) % 65535) + 1);
+
+	elog(DEBUG1, "Read %s disk page: stored checksum %u, computed checksum %u",
+		 label, ondisk_page_header.checkSum, computedCheckSum);
+
+	if (computedCheckSum != ondisk_page_header.checkSum)
+	{
+		ereport(WARNING,
+				(errcode(ERRCODE_DATA_CORRUPTED),
+				 errmsg("%s page checksum mismatch: stored %u, computed %u",
+						label, ondisk_page_header.checkSum, computedCheckSum)));
+		/* TODO: maybe ereport(ERROR) here once caller cleanup is safe */
+		return false;
+	}
+	return true;
+}
+
 /*
  * Now we have only one compresss version (1). When we have
  * different versions we'll need to bump
@@ -1269,7 +1349,7 @@ check_orioledb_compress_version(OrioleDBOndiskPageHeader ondisk_page_header)
  * Reads a page from disk to the img from a valid downlink. It's fills an empty
  * array of offsets for the page.
  */
-bool
+OReadPageResult
 read_page_from_disk(BTreeDescr *desc, Pointer img, uint64 downlink,
 					FileExtent *extent)
 {
@@ -1282,6 +1362,8 @@ read_page_from_disk(BTreeDescr *desc, Pointer img, uint64 downlink,
 	OrioleDBOndiskPageHeader ondisk_page_header = {0};
 	bool		needs_page_version_convert;
 
+	/* for the checksum computation we need aligned buffer */
+	Assert(((uintptr_t) img % sizeof(uint32)) == 0);
 	Assert(FileExtentOffIsValid(offset));
 	Assert(FileExtentLenIsValid(len));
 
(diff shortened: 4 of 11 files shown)


include/orioledb.h
```diff
@@ -385,7 +385,7 @@ typedef struct
 	 * pages it should be used for conversion of uncompressed images
 	 */
 	uint8		page_version;
-	uint32		reserved1;
+	uint32		checkSum;
```
pashkinelfe: PG shortens checksum to uint16 to write it on the page. While Orioledb has enough space in page header for uint32 checksum, I'd suggest we stick to uint16.

src/btree/io.c
```diff
@@ -1310,13 +1315,36 @@ read_page_from_disk(BTreeDescr *desc, Pointer img, uint64 downlink,
 			return false;
 
 		ondisk_page_header = *((OrioleDBOndiskPageHeader *) img);
+
 		needs_page_version_convert = check_orioledb_page_version(ondisk_page_header);
 
+		if (ondisk_page_header.page_version > 1)
```
pashkinelfe: - I think it's worth to add #define ORIOLEDB_PAGE_CHECKSUM_VERSION       2  /* Page version when saving page checksums became possible */
- Make repeated block as function (check_orioledb_page_checksum)
- Need to consider case when checksums are off by GUC

src/btree/io.c
```diff
@@ -1236,7 +1238,7 @@ store_read_page_checkpoint_stats(uint32 checkpointNum)
 static bool
 check_orioledb_page_version(OrioleDBOndiskPageHeader ondisk_page_header)
 {
-	if (ondisk_page_header.page_version != ORIOLEDB_PAGE_VERSION)
+	if (ondisk_page_header.page_version != 1 && ondisk_page_header.page_version != ORIOLEDB_PAGE_VERSION)
```
pashkinelfe: I propose following logic: 
ondisk_page_header.page_version == 1 && ORIOLEDB_PAGE_VERSION == 2 returns false (1->2 needs only checksum calculation, not conversion)
ondisk_page_header.page_version == ORIOLEDB_PAGE_VERSION - returns false
ondisk_page_header.page_version > ORIOLEDB_PAGE_VERSION - elog FATAL ("page from the future version")
ondisk_page_header.page_version < ORIOLEDB_PAGE_VERSION returns true (not implemented yet)

src/btree/io.c
```diff
@@ -1400,11 +1470,18 @@ write_page_to_disk(BTreeDescr *desc, FileExtent *extent, uint32 curChkpNum,
 				   Pointer page, off_t page_size)
 {
 
+	StaticAssertDecl(ORIOLEDB_BLCKSZ == BLCKSZ, "OrioleDB BLCKSZ doesn't match PG BLCKSZ");
```
pashkinelfe: Is this condition necessary? To date I think we could build Orioledb with non-strandard PG BLCKSZ. I'd propose to use OrioleDBChecksummablePage type to untie them.

src/btree/io.c
```diff
@@ -1416,82 +1493,74 @@ write_page_to_disk(BTreeDescr *desc, FileExtent *extent, uint32 curChkpNum,
 		byte_offset &= S3_OFFSET_MASK;
 	}
 
-	if (!OCompressIsValid(desc->compress))
```
pashkinelfe, line 1419: Could you turn back this case on top for easy reading of this PR diff?

src/btree/io.c
```diff
@@ -1416,82 +1493,74 @@ write_page_to_disk(BTreeDescr *desc, FileExtent *extent, uint32 curChkpNum,
...
+		if (use_device)
+			byte_offset *= (off_t) ORIOLEDB_COMP_BLCKSZ;
+		else
+			byte_offset *= (off_t) ORIOLEDB_BLCKSZ;
+
+		memcpy(&buf[O_PAGE_HEADER_SIZE], page + O_PAGE_HEADER_SIZE, ORIOLEDB_BLCKSZ - O_PAGE_HEADER_SIZE);
+		/* now buf has header then page in full, can be written in one call */
 	}
 
+	ondisk_page_header->checkSum = pg_checksum_block(&cp);
```
pashkinelfe: I'd propose to null ondisk_page_header->checkSum explicitly just before calculation not relying on memset above to be sure

pashkinelfe:
As an update to previous comments, Orioledb is planning to bump ORIOLEDB_BINARY_VERSION in another PR. Therefore, we will not find a situation when we need to read pages with ORIOLEDB_PAGE_VERSION before checksums introduced. So you can leave ORIOLEDB_PAGE_VERSION 1 and not bother yourself with code to support reading pre-checksum pages. (Support for writing/reading pages when page_checksums GUC is off still needed) @deadtrickster

deadtrickster:
Thank you, I dedicated today's evening to this PR so this note is welcome.

src/btree/io.c
deadtrickster: pg_checksum_block wants PGChecksummablePage, with these lines:

```

	/* ensure that the size is compatible with the algorithm */
	Assert(sizeof(PGChecksummablePage) == BLCKSZ);
```

My undertanding is that decoupling here means having our copy of the pg_checksum_block. Thoughts?

deadtrickster: Also the union itself:

```
/* Use a union so that this code is valid under strict aliasing */
typedef union
{
	PageHeaderData phdr;
	uint32		data[BLCKSZ / (sizeof(uint32) * N_SUMS)][N_SUMS];
} PGChecksummablePage;
```

deadtrickster:
Pushed, checksum now uint16, so old reserved1 is back, but as uint16. 
Oriole's checksummable page type is on hold - I asked a question.

On GUC naming - some existing ones have oriole_ prefix, some do not. I went without prefix.

deadtrickster:
So, thinking more about the GUC behaviour and my python tests - I'm very uncomfortable with silently "skipping" that corrupted page when `page_checksum` is `off`. So, maybe always check the checksum when it is not 0? preserving the original intent here. Just not computing it when it is `off`. 

UPD: pushed another commit - checksum is going to be verified on reads if not 0 regardless of GUC.

pashkinelfe:
@deadtrickster my thoughts is that we can stick to PG checksum behavior https://www.postgresql.org/docs/current/checksums.html i.e. 
- Checksums can be disabled when the cluster is initialized using [initdb](https://www.postgresql.org/docs/current/app-initdb.html#APP-INITDB-DATA-CHECKSUMS). 
- They can also be enabled or disabled at a later time as an offline operation. **UPDATE: we'll need hook for adding page checksums to all Oriole pages on turning checksums on, see [comment below](https://github.com/orioledb/orioledb/pull/964#discussion_r3594577218)** (implementation for this is not needed immediately, we can make separate PR for https://www.postgresql.org/docs/current/app-pgchecksums.html app.

When page checksums are off it's done to save a little on the page IO performance. We'd better not process non-zero values as valid.

src/btree/io.c
pashkinelfe: Yes, OrioleDBChecksummablePage is similar to PGChecksummablePage but sized by ORIOLEDB_BLCKSZ, not BLCKSZ. Probably this calls for having slightly modified version of pg_checksum_block in Orioledb repo, that should not be a problem.

doc/usage/configuration.mdx
```diff
@@ -221,6 +221,16 @@ Display compression column in orioledb_table_description.
 Try to allocate files as `sparse` at filesystem level. Saves occupied space on disk by excluding non-allocated
 regions in file from occupying disk space. It's an experimental option.
 
+### `orioledb.page_checksum`
```
pashkinelfe: If we accidentally turn it on, will all reads of previously written pages fail?
Maybe we can use existing data_checksums GUC? https://www.postgresql.org/docs/current/runtime-config-preset.html#GUC-DATA-CHECKSUMS

It's read-only and have necessary guards from accidentally turning it on, without modifying all pages checksums. But this means that we need to patch pg_checksums with hook to process all orioledb pages and write checsums when pg_checksums turns checksums on.

deadtrickster: no, when turned on reads will not fail. When checksums are turned off, checksum field is zeroed which distinguishes it from having _any_ checksum (see +1 in uint16 truncation). So old pages without checksums will not be checked.

za-arthur:
> So, thinking more about the GUC behaviour and my python tests - I'm very uncomfortable with silently "skipping" that corrupted page when page_checksum is off. So, maybe always check the checksum when it is not 0? preserving the original intent here. Just not computing it when it is off.

Agree with @pashkinelfe that we should stick with the Postgres behavior and do not check checksum if it is turned off.

doc/usage/configuration.mdx
deadtrickster:
```
	computedCheckSum = (uint16) ((pg_checksum_block((const PGChecksummablePage *) buf) % 65535) + 1);

``` 

stole it from checksum_impl.h where it had this comment:

```
	/*
	 * Reduce to a uint16 (to fit in the pd_checksum field) with an offset of
	 * one. That avoids checksums of zero, which seems like a good idea.
	 */
```

deadtrickster: ok I just reread what I wrote - you asked what will happen when checksums are turned on accidentally. So, my comment should read as "when checksums are turned on, the old pages have checksum field zeroed..." Sorry for the confusion.

deadtrickster:
```
	/* page was written with checksums disabled */
	if (((OrioleDBOndiskPageHeader *) buf)->checkSum == 0)
	{
		return true;
	}
```

deadtrickster:
Checksum is no longer verified on read if GUC is off

doc/usage/configuration.mdx
pashkinelfe: In PG when checksums are on PageIsVerifiedExtended() treats page with 0 pd_checksum as checksum_failure. I'd propose we do the same

pashkinelfe: It looks tricky and you are right @deadtrickster that we probably can safely detect page created without a checksum and skip checking in principle. Nevertheless, I think we don't want behavior different from what PG does with its heap pages. 

Another thing that may need attention is exclusion for all_zeroes page in PageIsVerifiedExtended() even if checksums are on.

deadtrickster:
I changed the Page type, and switch to strict checksum verification on reads, expecting pg_checksums to write checksums if needed.

pashkinelfe:
@deadtrickster the following logic is possible for turning checksums on by pg_checksums:
- Check that cluster was shut down cleanly (so .map file from the before-shutdown-checkpoint contains a relevant array of free blocks)
- Iterate offset in relation file by 8K (compression off), or 512B (compression on). 
- If current block is free according to map, skip it
- If current block is not free, calculate and write checksum. Length of compressed block in multiples of 512 is read from its header.

deadtrickster:
noted

@pashkinelfe should it be hard coded 8k though? in the context of this PR we already discussed different block sizes? so I assume 8k as a reference to whatever Oriole's block size currently is. Agree?

pashkinelfe:
@deadtrickster You can use ORIOLEDB_COMP_BLCKSZ (512) and ORIOLEDB_BLCKSZ (8192). Maximum page size in both cases is ORIOLEDB_BLCKSZ so OrioleDBChecksummablePage will have place for both cases.

deadtrickster:
would you mind looking at the current code in this branch? it already has OrioleDBChecksummablePage.

it is reddish now, after pg_checksums approach adoption, but that going to be fixed soonish.

include/orioledb.h
```diff
@@ -385,10 +385,81 @@ typedef struct
...
-	uint32		reserved1;
+	uint16		checkSum;
+	uint16		reserved1;
 	uint32		reserved2;
 } OrioleDBOndiskPageHeader;
 
+/* number of checksums to calculate in parallel */
+#define N_SUMS 32
+/* prime multiplier of FNV-1a hash */
+#define FNV_PRIME 16777619
```
pashkinelfe: inlcude src/include/storage/checksum_impl.h

include/orioledb.h
```diff
@@ -385,10 +385,81 @@ typedef struct
...
+
+typedef union
```
pashkinelfe: I'd move things below to patchset as they are needed for pg_checksums util

src/btree/io.c
```diff
@@ -1416,81 +1469,78 @@ write_page_to_disk(BTreeDescr *desc, FileExtent *extent, uint32 curChkpNum,
...
+		ondisk_page_header->compress_version = ORIOLEDB_COMPRESS_VERSION;
 
-		/* Write header first */
-		ondisk_page_header.compress_page_size = page_size;
-		ondisk_page_header.checkpointNum = curChkpNum;
-		ondisk_page_header.compress_version = ORIOLEDB_COMPRESS_VERSION;
-		ondisk_page_header.page_version = ORIOLEDB_PAGE_VERSION;
-
-		write_size = O_PAGE_HEADER_SIZE;
-		err = btree_smgr_write(desc, (char *) &ondisk_page_header, chkpNum, write_size, byte_offset) != write_size;
```
pashkinelfe, line 1464: I can't remember what was exact reason to do double btree_smgr_write when compression in on (header then remaining page contents), but I'd propose not changing this in this PR

pashkinelfe: Before calculating checksums we'll need full page image, but when checksums are off we can save on memcpy-ing full block if we leave code as it was with separate btree_smgr_write for on-disk-header and for headerless page

src/btree/io.c
```diff
@@ -1416,81 +1469,78 @@ write_page_to_disk(BTreeDescr *desc, FileExtent *extent, uint32 curChkpNum,
...
-		elog(DEBUG1, "Wrote disk page: checkpoint %u size %d", curChkpNum, (int) page_size);
+		/* now buf has header then page in full, can be written in one call */
+	}
 
+	ondisk_page_header->checkSum = 0;
+	/* if enabled, ->checkSum won't be zero (+1) */
+	if (DataChecksumsEnabled())
+	{
+		/* mirror checksum_impl.h */
+		ondisk_page_header->checkSum = (uint16) ((oriole_checksum_block(&cp) % 65535) + 1);
```
pashkinelfe: When compression is on we may save on calculating checksums only for `n * 512 `bytes for existing chunks, not for full 8192 bytes. We'll need to adapt oriole_checksum_block() to do so, discarding everything after `n*512` bytes from calculation, whether it is null-ed or not

deadtrickster:
@pashkinelfe I made the diff as small as possible, also compressed tests were added. memcopy for checksumming now happens only if checksums are enabled. 

If looks good, I will move the relevant parts to Oriole's PG and continue on `pg_checksums`. Sounds good?

src/btree/io.c
```diff
@@ -1460,6 +1516,22 @@ write_page_to_disk(BTreeDescr *desc, FileExtent *extent, uint32 curChkpNum,
 		ondisk_page_header.compress_version = ORIOLEDB_COMPRESS_VERSION;
 		ondisk_page_header.page_version = ORIOLEDB_PAGE_VERSION;
 
+		if (DataChecksumsEnabled())
+		{
+			off_t		image_size = extent->len * ORIOLEDB_COMP_BLCKSZ;
```
pashkinelfe: I think this could split on a condition https://github.com/orioledb/orioledb/pull/964/changes#diff-7dc6c778039133a43bc9cfe0757e64c284a1a9a74c816f5bf5e6e4108e0411a5L1471: 
if (page_size == ORIOLEDB_BLCKSZ)
 image_size = ORIOLEDB_BLCKSZ

i.e. when compression freed less that 512 bytes (one ORIOLEDB_COMP_BLCKSZ) page is still written full and compression result discarded

deadtrickster: did the split

src/btree/io.c
```diff
@@ -1333,6 +1366,10 @@ read_page_from_disk(BTreeDescr *desc, Pointer img, uint64 downlink,
 			ondisk_page_header = *((OrioleDBOndiskPageHeader *) buf);
 			needs_page_version_convert = check_orioledb_page_version(ondisk_page_header);
 
+			if (DataChecksumsEnabled() &&
+				!check_orioledb_page_checksum(ondisk_page_header, buf, "compressed", oriole_checksum_block_len(read_size)))
```
pashkinelfe, line 1425: Postgres outputs ERROR: ERRCODE_DATA_CORRUPTED if PageIsVerifiedExtended returns false. Likely we should do the same in all calls of check_orioledb_page_checksum

include/orioledb.h
```diff
@@ -385,10 +385,87 @@ typedef struct
...
+{
+	OrioleDBOndiskPageHeader phdr;
+	uint32		data[ORIOLEDB_BLCKSZ / (sizeof(uint32) * N_SUMS)][N_SUMS];
+} OrioleDBChecksummablePage;
+
+StaticAssertDecl(sizeof(OrioleDBChecksummablePage) == ORIOLEDB_BLCKSZ,
+				 "OrioleDBChecksummablePage size doesn't match ORIOLEDB_BLCKSZ");
+
```
pashkinelfe: Maybe add also 
 StaticAssertDecl(ORIOLEDB_COMP_BLCKSZ %
  (sizeof(uint32) * N_SUMS) == 0
StaticAssertDecl(ORIOLEDB_BLCKSZ %
  (sizeof(uint32) * N_SUMS) == 0

src/btree/io.c
deadtrickster: I think I'm doing it already - https://github.com/orioledb/orioledb/pull/964/changes#diff-7dc6c778039133a43bc9cfe0757e64c284a1a9a74c816f5bf5e6e4108e0411a5R1268

deadtrickster: in the latest commit I added return codes for read_page_from_disk. so we can surface different errors

deadtrickster:
@pashkinelfe Per Contributors chat - readonly guc (on by default, following perf test result) gates checksums for forward compatibility. Also brought back checksum_impl - constants copypasta gone now.

include/checkpoint/control.h
```diff
@@ -32,7 +32,7 @@ typedef struct
  * Also on-the-flight conversion routine should be added to
  * check_checkpoint_control()
  */
-#define ORIOLEDB_CHECKPOINT_CONTROL_VERSION	1
+#define ORIOLEDB_CHECKPOINT_CONTROL_VERSION	2
```
pashkinelfe, line 35: Reading control file from ORIOLEDB_CHECKPOINT_CONTROL_VERSION	1 with binary from this PR likely ends up with elog(ERROR, "Wrong CRC in control file"); because offsetof(CheckpointControl, crc) changed in this PR. I'm not sure if we need proper conversion of ORIOLEDB_CHECKPOINT_CONTROL_VERSION	1 -> 2 or we may not care. @akorotkov what do you think?

deadtrickster: in https://github.com/orioledb/orioledb/pull/964/changes/432b8526c9f8330b2c4e5e252b2ab5d23a4d76d5  I moved version check before crc check. At least something for now.

deadtrickster: In a way it is a chicken-eggs thing - corrupted control file can have the corruption exactly at version bytes so yeah, interesting.

pashkinelfe: I think this is correct. The question if we need conversion support for reading version 1 is still relevant

include/orioledb.h
```diff
@@ -470,6 +471,13 @@ extern bool orioledb_table_description_compress;
 extern char *max_bridge_ctid_string;
 extern BlockNumber max_bridge_ctid_blkno;
 extern bool orioledb_s3_mode;
+
+/*
+ * Read-only GUC "orioledb.checksums_enabled".
+ * read_page_from_disk still checks it for forward compatibility
+ */
+extern bool orioledb_checksums_enabled;
+#define ORIOLEDB_DEFAULT_CHECKSUMS_ENABLED	true
```
pashkinelfe: I think this define is redundant, propose just set true.

include/checkpoint/control.h
pashkinelfe: Conversion from version 1 is not needed, current state is right.

pashkinelfe:
@deadtrickster I squashed commits with minor changes (force-pushed)

pashkinelfe (approved):

PR #968, opened 2026-07-14
Support cross-major pg_upgrade of OrioleDB clusters

akorotkov:
Make `pg_upgrade` work end to end across PostgreSQL major versions
(16→17, 16→18, 17→18) for OrioleDB clusters, plus a CI matrix that
exercises it.

## Why
`pg_upgrade` migrates the catalog and the heap relfiles it knows about,
but OrioleDB keeps its table/index data in its own `orioledb_data/` tree
and persists version-specific metadata (PostgreSQL node trees and
`FormData_pg_attribute`) in its system trees. A naive upgrade therefore
came up empty or crashed on the foreign on-disk format.

## What the support commit does
**Binary-upgrade restore**
- Keep preserved relfilenodes in `assign_new_oids` and skip the physical
  DDL OrioleDB would normally perform, so the schema restore neither
  allocates new storage nor rejects preserved relnodes.
- `orioledb_ambuild` builds nothing under `IsBinaryUpgrade`; data is
  carried over wholesale from the old cluster.
- Fix "cache lookup failed for database" when restoring `template1` by
  advancing the command counter before refreshing the database cache.

**Foreign on-disk metadata**
- Stamp the checkpoint control file with the writing PostgreSQL major
  (`ORIOLEDB_CHECKPOINT_CONTROL_VERSION 2`, with v1 conversion) and a
  `resetSysCaches` flag when the running major differs.
- Rebuild the version-specific database sys-cache after a cross-major
  upgrade instead of reading the foreign-format entry.
- Version-stamp persisted node trees (index expressions/predicates): on
  read, trees written by another major are tolerated and dropped rather
  than crashing `stringToNode`.
- `orioledb_upgrade_refresh()` rewrites index expressions into the new
  server's node-tree format.

## CI
A `pg_upgrade` workflow builds OrioleDB against both the old and new
majors, populates an old-version cluster (incl. expression + partial
indexes) with the regression schema, runs `pg_upgrade --check` and the
real upgrade, copies the `orioledb_data/`/`orioledb_undo/` trees, checks
that touching an expression index before `orioledb_upgrade_refresh()`
errors cleanly (no crash), refreshes, then verifies via pre/post
schema+data dumps and a re-run of the regression schedule. Matrix:
{16→17,16→18,17→18} × {amd64,arm64} × {gcc,clang}.

Rebased onto current main; the `btree_sys_check` expected output is
regenerated across all three PG-major variants for the slightly larger
catalog records this adds.

### DIFF
--- a/ci/pg_upgrade.sh
+++ b/ci/pg_upgrade.sh
@@ -386,12 +386,10 @@ fi
 	-o "-p $PORT_NEW" -l "$GITHUB_WORKSPACE/pg${NEW_VERSION}.log" -w start
 
 # ----------------------------------------------------------------------
-# 7a. Before anything triggers the automatic refresh, confirm that a
-#     plain read of an index whose expression trees were written by the old
-#     major raises a clean error instead of crashing.  A bare SELECT does not
-#     go through ProcessUtility, so it does not trip the automatic refresh
-#     (step 7b) and still exercises the read-time fallback guard.  Only
-#     meaningful across majors; same-version carries compatible trees.
+# 7a. Before refreshing, confirm that touching an index whose expression
+#     trees were written by the old major raises a clean error instead of
+#     crashing the backend.  Only meaningful across majors; same-version
+#     "upgrades" carry compatible trees and need no refresh.
 # ----------------------------------------------------------------------
 if [ "$OLD_VERSION" != "$NEW_VERSION" ] && db_exists $PORT_NEW "$TEST_DB"; then
 	set +e
@@ -400,7 +398,7 @@ if [ "$OLD_VERSION" != "$NEW_VERSION" ] && db_exists $PORT_NEW "$TEST_DB"; then
 	refresh_rc=$?
 	set -e
 	if [ $refresh_rc -eq 0 ]; then
-		echo "ERROR: expression-index table was readable before the automatic refresh"
+		echo "ERROR: expression-index table was readable before orioledb_upgrade_refresh()"
 		echo "$refresh_err"
 		exit 1
 	fi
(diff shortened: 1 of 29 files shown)


src/catalog/o_tables.c
```diff
@@ -2227,6 +2289,14 @@ deserialize_o_table(Pointer data, Size length)
...
 		}
 	}
+
+	/*
+	 * If any index expression/predicate was written by a different PG major
+	 * version it was skipped above (its nodeToString format is unreadable
+	 * here); flag the table so the expressions get re-derived from the
+	 * catalog and rewritten before use.
+	 */
+	o_table->refresh_exprs = o_node_deserialize_format_changed;
```
pgnickb, line 2334: Shouldn't we do this for proc cache as well?
There we also call `o_deserialize_node`, so if the first access is via a proc and not direct query we might crash when `o_deserialize_node` silently returns NULL.

ci/pg_upgrade.sh
```diff
@@ -0,0 +1,494 @@
...
+#     skips on read (see o_deserialize_node); orioledb_upgrade_refresh()
+#     re-derives them from the catalog and rewrites them.  Run once per
+#     database that has the extension, before anything reads those tables.
+# ----------------------------------------------------------------------
+for db in "$TEST_DB" regression; do
+	db_exists $PORT_NEW "$db" || continue
+	"$NEW_PREFIX/bin/psql" -p $PORT_NEW -d "$db" -v ON_ERROR_STOP=1 \
+		-c "SELECT orioledb_upgrade_refresh();" \
+		&& echo "Refreshed OrioleDB expressions in $db" \
+		|| echo "[WARN] orioledb_upgrade_refresh() failed in $db"
```
pgnickb: Should this really be just a `WARN`?

Feels like enough reason to throw an error here already instead of hoping that pg_dump diff will catch it later.

src/checkpoint/control.c
```diff
@@ -82,26 +82,43 @@ check_checkpoint_control(CheckpointControl *control)
 {
 	pg_crc32c	crc;
 
-	INIT_CRC32C(crc);
-	COMP_CRC32C(crc, control, offsetof(CheckpointControl, crc));
-	FIN_CRC32C(crc);
+	/*
+	 * Version 1 had no pgVersion field, so its CRC covered fewer bytes and
```
pgnickb: Shall we add some asserts to enforce this? That'd help detect incompatibility with v1 in case we later decide to change the structure of the `CheckpointControl`.

Something like what hstore does in a similar case:
```
	/* New format uses an HEntry for key and another for value */
	StaticAssertStmt(sizeof(HOldEntry) == 2 * sizeof(HEntry),
					 "old hstore format is not upward-compatible");
```

pgnickb: Claude proposes something like this:
```
/*
 * Mirrors the pre-v2 (ORIOLEDB_CHECKPOINT_CONTROL_VERSION == 1) layout of
 * CheckpointControl, i.e. the struct as it was before pgVersion was added.
 * It exists solely so the static assertions below can catch a future
 * reorder of CheckpointControl's fields that would silently break
 * check_checkpoint_control()'s v1-file conversion: that conversion relies
 * on pgVersion occupying exactly the byte range the v1 struct's crc field
 * used to.
 */
typedef struct
{
	uint64		controlIdentifier;
	uint32		lastCheckpointNumber;
	uint32		controlFileVersion;
	CommitSeqNo lastCSN;
	OXid		lastXid;
	UndoLocation lastUndoLocation;
	XLogRecPtr	toastConsistentPtr;
	XLogRecPtr	replayStartPtr;
	XLogRecPtr	sysTreesStartPtr;
	uint64		mmapDataLength;
	CheckpointUndoInfo undoInfo[NUM_CHECKPOINTABLE_UNDO_LOGS];
	UndoLocation checkpointRetainStartLocation;
	UndoLocation checkpointRetainEndLocation;
	OXid		checkpointRetainXmin;
	OXid		checkpointRetainXmax;
	uint32		binaryVersion;
	bool		s3Mode;
	pg_crc32c	crc;
} CheckpointControlV1Layout;

StaticAssertDecl(offsetof(CheckpointControlV1Layout, crc) ==
				 offsetof(CheckpointControl, pgVersion),
				 "pgVersion must occupy the byte range the v1 struct's crc field used to; "
				 "see check_checkpoint_control()'s v1 conversion");
StaticAssertDecl(sizeof(((CheckpointControlV1Layout *) 0)->crc) ==
				 sizeof(((CheckpointControl *) 0)->pgVersion),
				 "pgVersion and the v1 struct's crc field must be the same size");
```

To me it seems like an overkill, but has an advantage that we don't need to be "smart" about possible subtle differences between different compilers/architectures.

pgnickb (approved):
With proc cache being fixed we're in a good shape to move forward with this 👍

PR #969, opened 2026-07-15
Increase multi_insert tests portability

pashkinelfe:
Existing workflow with COPY FROM PROGRAM was not shell-independent. Replace COPY FROM PROGRAM with COPY TO/FROM temp file. Overall temp files size is little bit over 1MB, so they are left to be cleaned with other CI aftifacts.

### DIFF
--- a/test/expected/multi_insert.out
+++ b/test/expected/multi_insert.out
@@ -1,6 +1,7 @@
 CREATE SCHEMA multi_insert;
 SET SESSION search_path = 'multi_insert';
 CREATE EXTENSION IF NOT EXISTS orioledb;
+\getenv abs_builddir PG_ABS_BUILDDIR
 -- The batched same-leaf primary insert path must not change visible
 -- results regardless of the debug GUC.  Each scenario runs both with
 -- the batched path enabled and disabled so a mishap in either branch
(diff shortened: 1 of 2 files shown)


test/expected/multi_insert.out
```diff
@@ -110,15 +110,16 @@ SELECT orioledb_tbl_check('t_copy_unsorted'::regclass);
...
 -- keys in reverse order to keep the monotonicity-fallback path exercised
 -- across multiple leaves.
 CREATE TABLE t_copy_reverse (id integer PRIMARY KEY, val int) USING orioledb;
 INSERT INTO t_copy_reverse SELECT i * 10, i FROM generate_series(1, 5000) i;
+COPY (SELECT 50001 + i, i FROM generate_series(999, 1, -2) i)
+  TO '/tmp/orioledb_test_copy_reverse.data';
 SET orioledb.debug_disable_multi_insert = off;
-COPY t_copy_reverse FROM PROGRAM
-  'for i in $(seq 999 -2 1); do echo "$((50001 + i))\t$i"; done';
+COPY t_copy_reverse FROM '/tmp/orioledb_test_copy_reverse.data';
```
za-arthur: It might be good to re-use logic of Postgres tests and store this file in `PG_ABS_BUILDDIR`:
https://github.com/orioledb/postgres/blob/patches17_20/src/pl/plpgsql/src/sql/plpgsql_copy.sql#L7

pashkinelfe: Fixed in https://github.com/orioledb/orioledb/commit/68ed1c98963674613f316867110f7be0f66c6803

za-arthur (approved):

PR #974, opened 2026-07-17
Fix VACUUM FULL half-created index

bwyltcaxe:
Since pg18 pg_index keeps TOAST tables (see pg_index.h);  hence, when pg_index is under VACUUM FULL it also goes through orioledb_object_access_hook() which calls for CommandCounterIncrement() and makes the already-swapper but not-yet-rebuild index visible;  Hence, we should ignore VACUUM FULL/CLUSTER, this approach makes orioledb_object_access_hook() really no-op for VACUUM FULL for heap tables.

### DIFF
--- a/src/catalog/ddl.c
+++ b/src/catalog/ddl.c
@@ -132,6 +132,7 @@ static Oid	o_saved_reltablespace = InvalidOid;
 List	   *o_reuse_indices = NIL;
 static ORelOids saved_oids;
 static bool in_rewrite = false;
+static bool in_cluster_rebuild = false;
 List	   *reindex_list = NIL;
 static Query *savedDataQuery = NULL;
 static IndexBuildResult o_pkey_result = {0};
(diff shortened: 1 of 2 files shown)


src/catalog/ddl.c
```diff
@@ -919,6 +920,7 @@ orioledb_utility_command(PlannedStmt *pstmt,
 		ORelOidsSetInvalid(saved_oids);
 		savedDataQuery = NULL;
 		in_nontransactional_truncate = false;
+		in_cluster_rebuild = false;
```
pashkinelfe, line 923: Should we better guard this and all similar places under `#if PG_VERSION_NUM >= 180000` ?

bwyltcaxe: Hi, Pavel!
I am not sure, to be honest. Now we throw the error for orioledb' tables, so  only heap-tables goes this way. During a `VACUUM FULL` heap rewrite there's never any legitimate orioledb TOAST work. Even if they don't fail (do not trigger the Assert), I don't think we should go with `CommandCounterIncrement()` for them at `orioledb_object_access_hook()`.

But if we want to highlight the specific problem since the 18th core, I can do that.

pashkinelfe:
Force-pushed with pgindent

pashkinelfe (approved):

PR #977, opened 2026-07-19
Parallel bitmap heap scan support

akorotkov:
Parallel bitmap heap scan, native key-bitmap path plus a bridged-index path. Stacked on #1011.

**History reworked for review.** The original 12-commit history is preserved unchanged at `parallel-bitmap-scan-orig-20260817`; the tree it produced was verified byte-identical to the reworked one before the rebase (tree `7fab0bc1`), and the diff against the base branch is the same 16 files / +1886 / -36.

Five commits, in review order:

1. **Fix uninitialised csn on the bridged index-fetch planner-probe path** — a pre-existing bug that this branch's valgrind runs surfaced, unrelated to parallel scan. It was the last commit before; it belongs first, and it can be reviewed and merged on its own.
2. **Key bitmap: flat finalized form that can be serialized and attached** — the data-structure work, no behaviour change on its own.
3. **btree scan: parallel callback seq scan constructor** — the scan-side hook the next commit needs.
4. **Parallel bitmap heap scan (native key-bitmap path)** — the feature, with the recovery skip, regression, isolation and testgres tests, and the CI diff filter folded in.
5. **Parallel bitmap scan for a single bridged index (DSA TIDBitmap)** — the bridged variant with its concurrency test.

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -169,6 +169,7 @@ ISOLATIONCHECKS = bitmap_hist_scan \
 				  merge \
 				  parallel_idx_scan \
 				  parallel_ordered_scan \
+				  parallel_bitmap_scan \
 				  partition_move \
 				  rightlink \
 				  rll \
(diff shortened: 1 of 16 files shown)


pgnickb (approved):
Aside from some minor test coverage concerns the code looks good to go.

The tests can be added in a later commit.

PR #980, opened 2026-07-22
FATAL on corrupted page file during checkpoint tree load

pashkinelfe:
When read_page_from_disk fails in evictable_tree_init_meta, the ERROR came before the sharedRootInfo was inserted into
SYS_TREES_SHARED_ROOT_INFO. On the next checkpoint cycle o_find_shared_root_info returns NULL, tree_is_under_checkpoint returned true (the checkpointer matches its own state), and o_btree_load_shmem_internal gave up, so the corrupted tree was silently skipped and the checkpoint succeeded.


pashkinelfe:
Rebased on fresh main branch

pgnickb (approved):
LGTM

Issue #663, opened 2025-12-01
Streaming replication error on subscriber: PGRES_FATAL_ERROR [1]


pashkinelfe:
Orioledb: https://github.com/orioledb/orioledb/commit/7a6e095a1c51d53e973832b121ca80dab8388de6

Apply patch:

[0001-Enable-running-PG-subscription-tests-on-Oriole-table.patch](https://github.com/user-attachments/files/22277304/0001-Enable-running-PG-subscription-tests-on-Oriole-table.patch)

Run

cd src/test/subscription
make installcheck TEMP_CONFIG="./orioledb.conf" PG_TEST_INITDB_EXTRA_OPTS="--locale=C" PROVE_TESTS="t/015_stream.pl"

Test hangs with repeated error on subscriber: ```could not access status of transaction 4286578688```

```
2025-12-01 14:10:57.491 +04 [678391] LOG:  logical replication apply worker for subscription "tap_sub" has started
2025-12-01 14:10:57.492 +04 [678393] 015_stream.pl LOG:  statement: SELECT count(1) = 0 FROM pg_subscription_rel WHERE srsubstate NOT IN ('r', 's');
2025-12-01 14:10:57.495 +04 [678395] LOG:  logical replication table synchronization worker for subscription "tap_sub", table "test_tab_2" has started
2025-12-01 14:10:57.506 +04 [678395] LOG:  logical replication table synchronization worker for subscription "tap_sub", table "test_tab_2" has finished
2025-12-01 14:10:57.506 +04 [678397] LOG:  logical replication table synchronization worker for subscription "tap_sub", table "test_tab" has started
2025-12-01 14:10:57.515 +04 [678397] LOG:  logical replication table synchronization worker for subscription "tap_sub", table "test_tab" has finished
2025-12-01 14:10:57.604 +04 [678400] 015_stream.pl LOG:  statement: SELECT count(1) = 0 FROM pg_subscription_rel WHERE srsubstate NOT IN ('r', 's');
2025-12-01 14:10:57.627 +04 [678408] 015_stream.pl LOG:  statement: SELECT count(*), count(c), count(d = 999) FROM test_tab
2025-12-01 14:10:57.743 +04 [678391] ERROR:  could not receive data from WAL stream: ERROR:  could not access status of transaction 4286578688
        DETAIL:  Could not open file "pg_xact/0FF8": No such file or directory.
[...]

PR #981, opened 2026-07-23
Fix logical decoding of transactions with savepoint before first change

### LINKED ISSUE #936: Logical decoding emits nothing for a transaction inside a SAVEPOINT
Hello, after adding OrioleDB to [supabase/realtime](https://github.com/supabase/realtime) tests matrix, we noticed the CI started failing and I believe the root cause is because it's failing to produce logical decoding output when changes are made inside a `SAVEPOINT` (it works fine in a plain transaction).

Here's a script to reproduce:

`orioledb_logical_decoding_test.sh`

```sh
#!/usr/bin/env bash

# OrioleDB logical decoding test.
#
# Requires Docker. Run with:
#
#   ./orioledb_logical_decoding_test.sh

set -euo pipefail

run() {
  local image="$1"
  local out="stream-$(printf '%s' "$image" | tr '/:' '__').bin"
  echo "==> $image"
  : > "$out"
  docker run --rm -i -e POSTGRES_PASSWORD=postgres -e OUT="/out/$out" -v "$PWD:/out" "$image" bash -s <<'IN'
set -e
docker-entrypoint.sh postgres -c config_file=/etc/postgresql/postgresql.conf >/dev/null 2>&1 &
for i in $(seq 1 90); do pg_isready -q -h /var/run/postgresql -U postgres && break; sleep 1; done
sleep 2
export PGHOST=/var/run/postgresql PGUSER=postgres

psql -X -q <<'SQL'
CREATE TABLE messages (id bigint PRIMARY KEY, t text);
CREATE PUBLICATION repro_pub FOR TABLE messages;
SQL

pg_recvlogical -d postgres --slot repro_slot --create-slot -P pgoutput >/dev/null 2>&1
pg_recvlogical -d postgres --slot repro_slot --start \
  -o proto_version=2 -o publication_names=repro_pub -o binary=true \
  -f "$OUT" --no-loop >/dev/null 2>&1 &
RECV=$!
sleep 2

psql -X -q <<'SQL'
BEGIN; INSERT INTO messages VALUES (1, 'plain'); COMMIT;
BEGIN; SAVEPOINT s; INSERT INTO messages VALUES (2, 'savepoint'); RELEASE SAVEPOINT s; COMMIT;
SQL
sleep 4
kill $RECV 2>/dev/null || true
sleep 1
IN

  echo "bytes streamed: $(wc -c < "$out")"
  echo "rows streamed :"
  grep -ao -E 'plain|savepoint' "$out" | sort -u | sed 's/^/  - /'
  echo
}

run supabase/postgres:17.6.1.127
run supabase/postgres:17.6.0.098-orioledb
```

Result:

```sh
❯ ./orioledb_logical_decoding_test.sh
==> supabase/postgres:17.6.1.127
bytes streamed:      214
rows streamed :
  - plain
  - savepoint

==> supabase/postgres:17.6.0.098-orioledb
bytes streamed:      129
rows streamed :
  - plain
```

I'm not sure if that's a bug but it seems like it.

DmitryNFomin:
Fixes #936.

## Problem

If a transaction's first OrioleDB modification occurs inside a
subtransaction, and the subtransaction was created before any OrioleDB
modification, the transaction is not emitted by logical decoding:

```sql
BEGIN;
SAVEPOINT s1;
INSERT INTO data(data) VALUES ('savepoint');
RELEASE SAVEPOINT s1;
COMMIT;
```

The row is written to the table. No records for the transaction are emitted
to logical slots (test_decoding, pgoutput). No error is raised.

## Root cause

Push and pop of the logical-xid context are both guarded by
`have_retained_undo_location()`, evaluated at different points in time:

1. At `SAVEPOINT`, `undo_subxact_callback(SUBXACT_EVENT_START_SUB)` does not
   call `assign_subtransaction_logical_xid()` because
   `have_retained_undo_location()` is false. Nothing is pushed onto
   `prevLogicalXids`.
2. The first modification inside the subtransaction assigns the logical xid
   lazily in `get_current_oxid()`. `have_retained_undo_location()` becomes
   true.
3. At subtransaction end, `oxid_subxact_callback(SUBXACT_EVENT_COMMIT_SUB)`
   evaluates the guard again (now true), calls `release_logical_xid()`, and
   calls `setup_prev_logical_xid_ctx()`, which pops the empty stack and
   resets `logicalXidContext` to `InvalidTransactionId`.
4. Top-level commit calls
   `wal_commit(oxid, get_current_logical_xid(), ...)` and writes
   `WAL_REC_COMMIT` with `logicalXid = InvalidTransactionId`.
5. In `decode_on_record()`, an invalid logicalXid on `WAL_REC_COMMIT` is
   treated as an autonomous transaction and skipped. The DML records queued
   in the reorder buffer under the xid from step 2 are never committed.

DEBUG4 output from an unpatched build (`main@9ba4d1b`, PG17 patchset
`d48d155c`):

```
DEBUG:  RECEIVE record type 5 (INSERT) oids [ 5 16473 16483 ] oxid 55 logicalXId 96 heapXid 0
DEBUG:  RECEIVE record type 2 (COMMIT) oxid 55 logicalXId 0 heapXid 0
DEBUG:  IGNORED record type 2 (COMMIT) invalid logicalXid for oxid 55
```

The INSERT record carries logicalXId 96; the COMMIT record for the same oxid
carries 0. No `STORE logical xid` / `RESTORE logical xid` lines are present,
consistent with a pop without a preceding push.

The same reset suppresses `wal_joint_commit()` in `XACT_EVENT_PRE_COMMIT`
(`src/transam/undo.c:2460`), which requires a valid
`logicalXidContext.xid`. Mixed heap+OrioleDB transactions with the same
savepoint ordering therefore lose their OrioleDB changes in decoding as
well, when no heap xid exists at subtransaction end (i.e. the heap
modification happens after the subtransaction closes). Verified on
main@9ba4d1b: `BEGIN; SAVEPOINT s1; INSERT INTO oriole_tbl ...; RELEASE
SAVEPOINT s1; INSERT INTO heap_tbl ...; COMMIT;` decodes only the heap
row. If the heap modification happens inside the subtransaction, the
subtransaction-level `wal_joint_commit()` path compensates and decoding
is complete even without this fix.

## Change

- `prevLogicalXids` entries are now `PrevLogicalXidEntry`
  (`LogicalXidCtx` + `SubTransactionId` of the pushing subtransaction).
- `assign_subtransaction_logical_xid()` takes the `SubTransactionId` and
  pushes unconditionally, including an invalid context.
- `setup_prev_logical_xid_ctx()` pops only if the stack top was pushed by
  the ending subtransaction; otherwise it returns without modifying
  `logicalXidContext`.
- The release/restore block in `SUBXACT_EVENT_COMMIT_SUB` runs only when the
  stack top matches the ending subtransaction.
- `release_assigned_logical_xids()` skips entries with invalid xids.

## Validation

Environment: PG17.10, patchset `d48d155c`, extension `main@9ba4d1b`,
`--enable-debug --enable-cassert`.

Unpatched build: repro above inserts the row, `pg_logical_slot_get_changes`
returns no records for the transaction.

Patched build, `pg_logical_slot_get_changes` output verified for:

- savepoint → insert → release → commit: emitted;
- savepoint → insert → commit (no explicit release): emitted;
- savepoint → insert → rollback to savepoint → insert → commit: only the
  second insert emitted;
- insert → savepoint → insert → release → commit (previously working path):
  both inserts emitted, no regression.

Added testgres test `test_savepoint_before_first_change`
(`test/t/logical_test.py`) covering the first three cases. Full
`test.t.logical_test` suite run on the patched build.

## Notes

- Pre-existing, unchanged by this patch: `SUBXACT_EVENT_ABORT_SUB` restores
  the parent context without releasing the aborted subtransaction's logical
  xid slot; the slot is freed at backend transaction end only if still on
  the stack.
- During validation, transaction id `4286578688` (0xFF800000) appeared in
  test_decoding output as the decoded xid. Produced by the wraparound branch
  of `acquire_logical_xid()` (`src/transam/oxid.c:503`). Pre-existing
  behavior; possibly related to #663
  (`could not access status of transaction 4286578688`).

### DIFF
--- a/include/transam/oxid.h
+++ b/include/transam/oxid.h
@@ -209,7 +209,7 @@ extern void oxid_notify(OXid oxid);
 extern void oxid_notify_all(void);
 extern void advance_oxids(OXid new_xid);
 extern OXid get_current_oxid(void);
-extern void assign_subtransaction_logical_xid(void);
+extern void assign_subtransaction_logical_xid(SubTransactionId mySubid);
 extern void set_oxid_csn(OXid oxid, CommitSeqNo csn);
 extern void set_oxid_xlog_ptr(OXid oxid, XLogRecPtr ptr);
 extern void set_current_oxid(OXid oxid);
--- a/src/transam/oxid.c
+++ b/src/transam/oxid.c
@@ -114,16 +114,6 @@ reset_logical_xid_ctx(void)
 	logicalXidContext.useHeap = false;
 }
 
-static inline LogicalXidCtx *
-clone_logical_xid_ctx(void)
-{
-	LogicalXidCtx *clone = (LogicalXidCtx *) palloc(sizeof(LogicalXidCtx));
-
-	Assert(clone);
-	memcpy(clone, &logicalXidContext, sizeof(LogicalXidCtx));
-	return clone;
-}
-
 Datum
 orioledb_get_current_logical_xid(PG_FUNCTION_ARGS)
 {
@@ -136,8 +126,41 @@ orioledb_get_current_heap_xid(PG_FUNCTION_ARGS)
 	PG_RETURN_INT64(GetCurrentTransactionIdIfAny());
 }
 
-static List *prevLogicalXids = NIL; /* remember all xids on subxact's chain
-									 * for correct release */
+/*
+ * Saved logical xid context together with the subtransaction whose start
+ * pushed it.  The subid tag makes push and pop symmetric: a subtransaction
+ * end must only pop the entry it pushed itself.  A subtransaction that
+ * started before the transaction's first OrioleDB modification pushes
+ * nothing (see undo_subxact_callback), so its end must leave the current
+ * context untouched.
+ */
+typedef struct
+{
+	LogicalXidCtx ctx;
+	SubTransactionId subid;
+} PrevLogicalXidEntry;
+
+static List *prevLogicalXids = NIL; /* stack of PrevLogicalXidEntry for all
+									 * xids on subxact's chain, for correct
+									 * restore and release */
+
+/*
+ * Check whether the top of prevLogicalXids was pushed by the subtransaction
+ * identified by subid.
+ */
+static bool
+prev_logical_xid_pushed_by(SubTransactionId subid)
+{
+	PrevLogicalXidEntry *entry;
+
+	if (prevLogicalXids == NIL)
+		return false;
+
+	entry = (PrevLogicalXidEntry *) llast(prevLogicalXids);
+
+	return entry != NULL && entry->subid == subid;
+}
+
 static OXidMapItem *xidBuffer;
 
 /*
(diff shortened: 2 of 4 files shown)


DmitryNFomin:
Ran a second validation round on a clean stack: patched PostgreSQL at `.pgtags` commit `d48d155c`, extension at exactly this branch (`7ab534d`), `--enable-debug --enable-cassert`.

Results:

- `regresscheck`: 48/48 pass.
- `isolationcheck`: 32/32 pass.
- `test.t.logical_test`: 50/50 pass (1 skip: `wal2json` not installed).
- Decode scenarios on the patched build, all correct:
  - savepoint → insert → release → commit: emitted;
  - savepoint → insert → commit (implicit release): emitted;
  - savepoint → insert → rollback to savepoint → insert → commit: only the post-rollback insert emitted;
  - savepoint → oriole insert + heap insert inside subxact → release → commit: both emitted;
  - savepoint → oriole insert → release → heap insert → commit: both emitted.
- The last two scenarios were also run on unpatched `main@9ba4d1b`: with the heap modification inside the subtransaction, decoding is complete even unpatched (subtransaction-level `wal_joint_commit()` compensates); with the heap modification after the subtransaction ends, the OrioleDB change is absent from the decode. The PR description has been updated to state this condition precisely.
- `pgindent` (with regenerated `orioledb.typedefs`) and `yapf --diff` produce no changes on the modified files.

PR #983, opened 2026-07-24
Fix cost model incorrectly penalising index-only scans

pashkinelfe:
- Implement amnblocks, report index stats in amvacuumcleanup. (With nblocks = 0 for index other plans have unjustified preference over IOS)
- Set allvisfrac=1.0 for orioledb tables. Make this in accordance to visibility-aware orioledb indices, this makes IOS even more preferred in many cases
- Disable index-only scan for bridged indexes. In bridged scans every tuple triggers a bridge+primary lookup for visibility, making IOS no cheaper than a regular index scan.
- Report relallvisible = curpages in orioledb_amvacuumcleanup (this in no-op, for correctness only)
- Also fix bug in backward IOS in PG16


pgnickb (approved):
LGTM

pashkinelfe:
Force-pushed to update pgtags after patchset merges

PR #984, opened 2026-07-24
Fix primary index tree ignoring the table's tablespace

### LINKED ISSUE #986: No such file or directory error when pk spills on disk when table has tablespace
This test is currently red:

```python
def test_primary_tree_follows_table_tablespace(self):
		tbsp_dir = tempfile.mkdtemp(prefix='oriole_tbsp_')
		self.addCleanup(shutil.rmtree, tbsp_dir, ignore_errors=True)

		node = self.node
		# Small pool so insert overflows it and forces the PK tree onto disk.
		node.append_conf('postgresql.conf', "orioledb.main_buffers = 8MB\n")
		node.start()

		node.safe_psql('postgres', "CREATE EXTENSION orioledb;")
		node.safe_psql('postgres',
		               f"CREATE TABLESPACE ots LOCATION '{tbsp_dir}';")
		node.safe_psql(
		    'postgres', """
			CREATE TABLE o_ts (k int PRIMARY KEY, v text NOT NULL)
				USING orioledb TABLESPACE ots;
		""")
		node.safe_psql(
		    'postgres', """
			INSERT INTO o_ts
				SELECT i, repeat('x', 2000) FROM generate_series(1, 20000) i;
		""")

		self.assertEqual(
		    node.execute('postgres', "SELECT count(*) FROM o_ts;")[0][0],
		    20000)
```

I created two PRs:
#984 - moves PK to table's tablespace
#987 - reactively ensures default directory exists

deadtrickster:
When writing a repro for pg_checksums crash I hit this strange "no such file or directory" error

```
    FATAL: could not open data file orioledb_data/5/16481: No such file or directory
```

The key tree was resolving not to the table's tablespace, but default, which wasn't
there, IIUC because repro script didn't have anything in the default tablespace yet.

Somewhat large set is needed to spill the key on disk.

Fixed by forcing the primary tree to follow table's tablespace.

I did so many multi-terabyte tests and didn't catch it with tpcc, oh well.

closes #986

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -178,6 +178,7 @@ ISOLATIONCHECKS = bitmap_hist_scan \
 				  concurrent_truncate \
 				  uniq
 TESTGRESCHECKS_PART_1 = test/t/amcheck_test.py \
+						test/t/tablespace_test.py \
 						test/t/checkpointer_test.py \
 						test/t/correlation_test.py \
 						test/t/eviction_bgwriter_test.py \
--- a/src/catalog/o_indices.c
+++ b/src/catalog/o_indices.c
@@ -374,7 +374,11 @@ make_primary_o_index(OTable *table, OIndexVersionMode ixVerMode)
 	else
 		result->compress = table->primary_compress;
 	result->fillfactor = table->fillfactor;
-	result->tablespace = tableIndex->tablespace;
+
+	/*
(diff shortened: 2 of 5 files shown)


homper:
Decided to go this way: #987

PR #985, opened 2026-07-24
Add sparse files tests for relations and undo buffers

pashkinelfe:
test_sparse_files: extend with data correctness checks, DELETE/INSERT cycle, restart survival, and post-mutation st_blocks verification.

test_sparse_files_undo: verify data correctness after undo hole punching. Two REPEATABLE READ snapshots pin different undo ranges; after releasing the older one and checkpointing, the surviving snapshot must still read correct data through undo files that now contain holes. Also checks current-session values, orioledb_tbl_check, and restart survival.


pgnickb (approved):
LGTM.

Took me a bit to understand all the intentions behind the code. Would be nice to have more comments or a complete summary either in the test or in the commit message.

PR #988, opened 2026-07-24
Replace blanket streaming disable with per-transaction gating

DmitryNFomin:
Part of #663. Requires the paired orioledb/postgres change
(https://github.com/orioledb/postgres/pull/79) and a `.pgtags` bump after it merges; CI on this PR
will fail the patchset check until then.

## Problem

`orioledb_decode()` sets `ctx->streaming = false` for every OrioleDB WAL
container (`src/recovery/logical.c`, from commit e7feb24448ad
"Temporarily disable streaming replication"). The flag is never
restored, so one decoded OrioleDB record disables streaming for the rest
of the decoding session:

- Heap transactions stop streaming as well. Measured with test_decoding,
  `logical_decoding_work_mem=64kB`, fresh slots: a 5000-row heap
  transaction alone streams (18 blocks); the same transaction preceded
  by a single-row OrioleDB transaction does not stream (0 blocks).
- Consumers requesting `streaming=on/parallel` silently degrade to
  spill-to-disk decoding on the publisher.

## Change

- Every OrioleDB transaction is marked `RBTXN_NO_STREAMING` at its first
  decoded DML record. OrioleDB transactions keep the current spill-path
  behavior; heap transactions stream again.
- A per-decoding-session logicalXid→oxid map (fed from `WAL_REC_XID`
  records, pruned at finish records) backs the fork-side
  `decoding_xid_status_hook`: streamed decoding resolves OrioleDB
  logical-xid status from oxid state (`oxid_get_csn()`) instead of clog,
  removing the `could not access status of transaction` failure mode for
  any streaming that does occur.
- `ctx->streaming = false` is removed.

## Why OrioleDB transactions stay on the spill path

Streaming them is unsafe until the logical-xid graph is linked eagerly.
The graph (top logical xid, per-subtransaction logical xids, heap xid
for mixed transactions) is assembled only when finish records are
decoded (SWITCH_LOGICAL_XID / JOINT_COMMIT / savepoint records +
`ReorderBufferCommitChild`). A transaction that starts streaming and is
later re-parented by that machinery loses its already-streamed chunks:
they are never stream-committed. Observed directly (experimental build
with streaming enabled): a transaction inserting 2000 rows, then
creating a savepoint, then inserting 3000 more streams the first 2000
rows in 8 blocks with no `committing streamed transaction` marker — a
consumer applies only 3000 of 5000 rows. Whether a transaction will
later create a savepoint or acquire a heap xid is unknowable at the
moment streaming would start, so the gate must cover all OrioleDB
transactions. Enabling real streaming requires emitting the parent/child
links at record-emission time — flagged for discussion in #663.

## Validation

Environment: patched PG 17.10 + this extension, `--enable-debug
--enable-cassert`, test_decoding, `logical_decoding_work_mem=64kB`.

- Heap 5000-row transaction: streams (18 blocks + stream commit), all
  rows present.
- OrioleDB 5000-row transaction: 0 stream markers, 5000/5000 rows via
  spill + commit replay.
- OrioleDB 2000+savepoint+3000 transaction: 0 stream markers, 5000/5000
  rows (the data-loss scenario above is gone).
- Heap 5000-row transaction after OrioleDB records in the same decode
  pass: streams (18 blocks) — restores the behavior lost today.
- Suites on this stack: `test.t.logical_test` 50/50 (1 skip: wal2json),
  `regresscheck` 48/48, `isolationcheck` 32/32.
- New regression test `test_streaming_gate` covers the four scenarios.

### DIFF
--- a/include/recovery/logical.h
+++ b/include/recovery/logical.h
@@ -17,10 +17,12 @@
 #include "btree/btree.h"
 #include "recovery/internal.h"
 
+#include "access/xact.h"
 #include "replication/decode.h"
 #include "replication/logical.h"
 
 extern void orioledb_decode(LogicalDecodingContext *ctx,
 							XLogRecordBuffer *buf);
+extern DecodingXidStatus orioledb_logical_xid_status(TransactionId xid);
 
 #endif							/* __LOGICAL_H__ */
--- a/src/orioledb.c
+++ b/src/orioledb.c
@@ -1281,6 +1281,7 @@ _PG_init(void)
 	RegisterXactCallback(undo_xact_callback, NULL);
 	RegisterSubXactCallback(undo_subxact_callback, NULL);
 	get_xidless_commit_lsn_hook = orioledb_get_xidless_commit_lsn;
+	decoding_xid_status_hook = orioledb_logical_xid_status;
 	CacheRegisterUsercacheCallback(orioledb_usercache_hook, PointerGetDatum(NULL));
 	CheckPoint_hook = o_perform_checkpoint;
 	after_checkpoint_cleanup_hook = o_after_checkpoint_cleanup_hook;
--- a/src/recovery/logical.c
+++ b/src/recovery/logical.c
@@ -25,6 +25,9 @@
 #include "tableam/descr.h"
 #include "tuple/slot.h"
 
+#include "transam/oxid.h"
+
+#include "access/transam.h"
 #include "catalog/pg_tablespace.h"
 #include "replication/origin.h"
 #include "replication/reorderbuffer.h"
@@ -141,6 +144,27 @@ get_reorder_buffer_txn(ReorderBuffer *rb, TransactionId xid)
 		return ent->txn;
 	else
 		return NULL;
+}
+
+/*
+ * Mark the top-level transaction of the given xid as not streamable.
+ *
+ * OrioleDB's logical-xid graph (top logical xid, per-subtransaction logical
+ * xids, heap xid for mixed transactions) is fully assembled only when the
+ * finish records are decoded.  Streaming a partially-assembled graph sends
+ * pieces of one transaction as several independent streamed transactions,
+ * which the subscriber applies concurrently: parallel apply workers deadlock
+ * on each other and transaction atomicity is lost.  Until the graph is
+ * linked eagerly at record-emission time, transactions with subtransactions
+ * or heap involvement must take the spill path.
+ */
+static void
+disable_streaming_for_txn(ReorderBuffer *rb, TransactionId xid)
+{
+	ReorderBufferTXN *txn = get_reorder_buffer_txn(rb, xid);
+
+	if (txn != NULL)
+		rbtxn_get_toptxn(txn)->txn_flags |= RBTXN_NO_STREAMING;
 }
 
 /*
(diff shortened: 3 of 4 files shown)


pgnickb:
This was done intentionally. When the transaction starts we don't know if it will be a heap-only transaction. At any moment after the streaming is started it can modify an OrioleDB table. 

Are you testing for this case? How's your (claude's?) code handling it?

I haven't looked at the code, but I'm happy to discuss the decision.

DmitryNFomin:
The scenario was not covered by the PR's original validation. Tested now; results below. No code changes were needed — only test coverage (a6b875f).

## Handling

The gate does not predict at transaction start. It reacts to WAL. On the first OrioleDB modification of a transaction with an assigned heap xid, `acquire_logical_xid_wrapper()` emits `WAL_REC_SWITCH_LOGICAL_XID (heapXid -> logicalXid)` before the OrioleDB DML records in the same container. On decode of that record (`decode_on_record`, `WAL_REC_SWITCH_LOGICAL_XID` case):

1. `ReorderBufferAssignChild(heapXid, logicalXid)` (pre-existing);
2. `RBTXN_NO_STREAMING` is set on the top-level transaction.

No further streaming rounds select the transaction; subsequent changes spill. The already-streamed part is the standard partially-streamed-then-serialized state. At commit, `ReorderBufferStreamCommit` streams the remaining changes, including the OrioleDB child's. Child changes processed while the transaction is in progress hit `SetupCheckXidLive`/`HandleConcurrentAbort` on the fabricated child xid; `decoding_xid_status_hook` resolves it from the decode-session logicalXid→oxid map (fed by `WAL_REC_XID`) instead of the clog lookup that failed with `could not access status of transaction N` before e7feb24448ad.

## Results

PG 17.10 at this PR's branch pair, `--enable-cassert`, test_decoding, `logical_decoding_work_mem = 64kB`, heap part sized to stream before the OrioleDB part begins.

Wire-level (`pg_logical_slot_get_changes`, `stream-changes`):

- heap 5000 rows, then OrioleDB 200, one transaction: 18 streamed blocks, 1 stream commit, 5200/5200 changes, no errors.
- OrioleDB 200 first, then heap 5000: 1 stream commit, 0 plain BEGIN/COMMIT, 5200/5200 changes.
- heap 5000 + `SAVEPOINT; 200 OrioleDB rows; ROLLBACK TO SAVEPOINT; 1 OrioleDB row; COMMIT`: 5001/5001 changes, rolled-back rows absent, 1 stream commit.

End-to-end, two nodes, `CREATE SUBSCRIPTION ... WITH (streaming = parallel)`, same three transactions: row counts identical on both sides (15000 heap / 401 OrioleDB), post-rollback row replicated, `stream_txns = 3`, no errors or deadlocks in the subscriber log.

Note: in these transactions the OrioleDB changes do travel over the stream protocol, in the stream-commit tail. Same wire behavior as `ctx->streaming = false` when streaming has already started (the window in the e7feb24448ad message); the difference is that the fabricated xids exposed in that window are now resolvable, where today they can still produce the clog failure.

Would be glad to discuss the longer-term direction whenever convenient: either keep OrioleDB-touching transactions out of streaming (as this PR does), or emit the logical-xid parent/child links eagerly so they can be streamed as well. I wrote up the constraints I ran into with the eager approach in the #663 comment — your view on which way is worth pursuing would help a lot.

pgnickb:
I took a quick look at the code. What's this `decoding_xid_status_hook` that you set in `_PG_init`?

pashkinelfe:
I think we'd better work on enabling streaming on Oriole rather than these decoding complications to enable relatively low-value streaming for heap. I'm also not convinced that enabling streaming only for heap in a logical decoding session is safe for data integrity.

PR #989, opened 2026-07-27
Fix concurrency bugs found stress-hunting 027_stream_regress under churn

akorotkov:
## Summary

A batch of fixes for concurrency bugs surfaced by stress-hunting the
`027_stream_regress` workload under churn (concurrent `createdb`/`dropdb`
plus aggressive checkpoints, `checkpoint_timeout=30s`, on both a primary and a
streaming standby). Each commit fixes one distinct, independently-reproduced
bug (or an invalid debug assert). The stress-hunt harness and the churn-only CI
tuning (the `pg_tests`/`pg_tests_asan` matrix, `-DCHECK_PAGE_STRUCT`, the raised
replica sync timeout) are intentionally **not** part of this PR.

The fixes were validated on the churn matrix (64 `pg_tests` + 8 `pg_tests_asan`
jobs) and build cleanly against the patched PG18.

The sys-tree fuzzy-checkpoint recovery redesign that previously lived here has been split into its own PR (branch `fuzzy-sys-tree-checkpoint`).

### Crashes / invalid asserts under concurrency
- **vacuum: drop invalid page-count assert in the bridge-index second pass.**
  The `vacuumed_pages == lpdead_item_pages` cross-check is invalid for a
  concurrently-mutable B-tree (the two passes count dead-item pages across
  splits/merges), firing as a flaky SIGABRT (~12/64 jobs). The vacuuming is
  correct; only the structural cross-check was wrong. Item-count invariant kept.
- **btree scan: stop the bitmap-jump path at the rightmost internal page.**
  The bitmap "jump" branch copied the current internal page's hikey with no
  rightmost check; the rightmost page has no hikey → assert (`page_get_hikey`)
  / over-read. Made the jump branch terminate like the non-jump branch does.
- **btree iterator: scope the monotonicity assert to a single `col = ANY` element.**
  The cassert monotonicity check compared tuples *across* array elements, which
  need not be ordered relative to each other → rare SELECT crash. Reset
  `prevTuple` in `o_btree_iterator_advance()` so the check stays within one
  element (where ordering does hold).
- **btree io: tolerate a concurrently-removed secondary relnode file at fsync.**
  At PreCommit, `iterate_relnode_files` fsyncs segment/`.map`/`.evt` sidecars;
  the checkpointer's superseded-map cleanup can unlink a listed `.map` before
  the callback runs → `fsync` ERROR → PANIC (seen on CREATE TABLE AS). Rather
  than racily pre-checking with `access()` (which only narrows the window),
  `fsync_callback` now opens the file itself and returns quietly on `ENOENT`,
  closing the race; a genuine I/O error on a present file still ERRORs.
- **checkpoint: return the re-fetched descriptor from `checkpoint_ix`.**
  `checkpoint_ix()` re-fetches the index descriptor after its writeback relock
  (a concurrent DROP/TRUNCATE invalidation can free+rebuild it), but returned
  only `bool`, so `checkpoint_tables_callback()` kept using the stale descriptor
  in `sort_checkpoint_map_file`/`add_index_id_item` → later MAXALIGN trap in
  `FreeTupleDesc`. Return the current descriptor (NULL on failure) and adopt it;
  concurrent-deletion detection is preserved (no refcnt pin, so no `valid_doff`).
- **ddl: don't store a NULL missing value when adding a column.**
  For `ALTER TABLE ... ADD COLUMN`, `o_table_fill_constr()` marked the column's
  fast-default "missing value" present even when the evaluated default was NULL
  (the `!old_field` guard short-circuited the `missingIsNull` check), then
  `datumCopy()`'d a NULL varlena → SIGSEGV (or a garbage-`VARSIZE` hang). Gate
  `hasmissing` on `!missingIsNull`. Surfaced adding a domain-over-array column
  with no default to a populated table.
- **descr: preserve refcnt and free old contents when refilling an index descr.**
  `get_index_descr()` re-filled an existing `oIndexDescrHash` entry in place on a
  metadata-version mismatch via `o_index_fill_descr()`, which `memset`s the descr
  and resets refcnt to 0 — unlike `recreate_index_descr()` it neither freed the
  old tupdescs/`index_mctx` nor preserved refcnt. A descriptor still referenced
  by an `OTableDescr` thus had its refcnt clobbered to 0 and could be freed by a
  concurrent checkpoint invalidation while still in use (use-after-free → later
  `FreeTupleDesc` MAXALIGN trap). Mirror `recreate_index_descr()`: save refcnt,
  free the old contents, refill, restore refcnt.

### Sanitizer (ASAN/UBSAN) fixes
- **btree: decode rowids alignment-safely (no allocation).** Rowid addendum
  structs were decoded by casting a pointer at a MAXALIGN offset from a bytea
  that can be only 4-byte-aligned in place; the structs need 8-byte alignment
  (`CommitSeqNo`) → misalignment UB (UBSAN). Each decode/encode goes through an
  aligned fixed-size stack copy — no heap allocation, nothing to leak.
- **ddl: unpoison the drop arg to silence an ASAN stack-use-after-scope FP.**
  A legitimate `((ObjectAccessDrop *) arg)->dropflags` read tripped a
  false-positive from PG's `sigsetjmp`/`longjmp` stack-shadow staleness.
- **indexam: unpoison `amcostestimate` out-params (same FP class).** Writing
  `cost_index()`'s stack locals tripped the same `sigsetjmp`/`longjmp` false
  stack-use-after-scope. Both ASAN fixes are no-ops off ASAN.

### CI
- **ci: filter the OID-variant of the `regress_user11` role-collision race.**
  Under churn the known cross-test role race also surfaces as
  `could not find tuple for role <oid>`; added that variant to the existing
  scoped `skip_hunk_errors` so it no longer leaks into the filtered residual.

### DIFF
--- a/src/checkpoint/checkpoint.c
+++ b/src/checkpoint/checkpoint.c
@@ -185,7 +185,7 @@ static void sort_checkpoint_tmp_file(BTreeDescr *descr, int cur_chkp_index);
 static inline void checkpoint_ix_init_state(CheckpointState *state, BTreeDescr *descr);
 static void checkpoint_init_new_seq_bufs(BTreeDescr *descr, int chkpNum);
 static bool checkpoint_temporary_tree(int flags, BTreeDescr *descr);
-static bool checkpoint_ix(int flags, BTreeDescr *descr);
+static BTreeDescr *checkpoint_ix(int flags, BTreeDescr *descr);
 static uint64 checkpoint_btree(BTreeDescr **descrPtr, CheckpointState *state,
 							   CheckpointWriteBack *writeback);
 static Jsonb *prepare_checkpoint_step_params(BTreeDescr *descr,
@@ -1127,7 +1127,7 @@ checkpoint_sys_trees(int flags, uint32 cur_chkp_num,
 		if (desc->storageType == BTreeStoragePersistence ||
 			desc->storageType == BTreeStorageUnlogged)
 		{
-			success = checkpoint_ix(flags, desc);
+			success = checkpoint_ix(flags, desc) != NULL;
 			/* System trees can't be concurrently deleted */
 			Assert(success);
 			if (!orioledb_s3_mode)
@@ -1160,7 +1160,7 @@ checkpoint_chkp_nums(int flags, uint32 cur_chkp_num,
 	checkpoint_ix_init_state(checkpoint_state, desc);
 	checkpoint_init_new_seq_bufs(desc, cur_chkp_num);
 
-	success = checkpoint_ix(flags, desc);
+	success = checkpoint_ix(flags, desc) != NULL;
 
 	/* System trees can't be concurrently deleted */
 	Assert(success);
@@ -2711,12 +2711,16 @@ backend_set_autonomous_level(CheckpointState *state, uint32 level)
 /*
  * Make checkpoint of an index (persistent or unlogged).
  *
- * The caller must hold the OrioleDB checkpointer table lock.  On success
- * (true), the lock is still held and the caller must release it.  On
- * failure (false), the lock has already been released by
- * perform_writeback_and_relock() inside checkpoint_btree().
+ * The caller must hold the OrioleDB checkpointer table lock.  On success the
+ * lock is still held and the (possibly re-fetched) descriptor is returned; the
+ * caller MUST switch to it, because checkpoint_btree() /
+ * perform_writeback_and_relock() re-fetch the descriptor after each
+ * unlock/relock cycle and the descriptor passed in may have been freed by a
+ * concurrent invalidation in the meantime.  On failure NULL is returned and
+ * the lock has already been released by perform_writeback_and_relock() inside
+ * checkpoint_btree().
  */
-static bool
+static BTreeDescr *
 checkpoint_ix(int flags, BTreeDescr *descr)
 {
 	FileExtentsArray *free_extents = NULL;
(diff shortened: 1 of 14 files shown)


src/recovery/recovery.c
```diff
@@ -3641,6 +3665,155 @@ recovery_send_init(int worker_num)
...
+		{
+			if (s->o_tables_meta_locked &&
+				!XLogRecPtrIsInvalid(s->meta_window_lsn) &&
+				(XLogRecPtrIsInvalid(oldest) || s->meta_window_lsn < oldest))
+				oldest = s->meta_window_lsn;
+		}
+	}
+
+	pg_atomic_write_u64(&checkpoint_state->oldestOpenMetaWindow,
+						(uint64) oldest);
```
pgnickb: Nit pick:
add `pg_write_barrier()` here after the write?

src/checkpoint/checkpoint.c
```diff
@@ -1416,29 +1421,59 @@ o_perform_checkpoint(XLogRecPtr redo_pos, int flags)
...
+	 * CommitTransaction()'s PreCommit_on_commit_actions() via o_drop_table()
+	 * -> o_tables_meta_lock() (LW_SHARED), *after* orioledb stamped the
+	 * commit position at XACT_EVENT_PRE_COMMIT.  If we held these locks
+	 * EXCLUSIVE and spun waiting for that commit to drain, it could never
+	 * take the lock and we would deadlock.  So whenever an unfinished commit
+	 * is found, release both locks, sleep to give it (and any other blocked
+	 * committer) a chance to grab them and finish, then retry from a fresh
+	 * replayStartPtr.
 	 */
+	for (;;)
```
pgnickb: Not a regression but perhaps still worth performing a `CHECK_FOR_INTERRUPTS` here.

I think the check in `acquire_chkp_lock_drain` won't be reached from here most of the time because the lock is released before the sleep

pgnickb (commented):
Some small comments.

pgnickb (approved):
With the META_LOCK/UNLOCK moved out, this is good to go.

PR #990, opened 2026-07-27
Add CHECK_FOR_INTERRUPTS to workers_synchronize

pgnickb:
Previously the busy loop in the workers_synchronize function didn't check for interrupts which meant it woudld ignore signals such as SIGTERM and won't exit cleanly.

### DIFF
--- a/include/workers/interrupt.h
+++ b/include/workers/interrupt.h
@@ -14,30 +14,11 @@
 #ifndef __WORKERS_INTERRUPT_H__
 #define __WORKERS_INTERRUPT_H__
 
-#include "miscadmin.h"
-#include "postmaster/startup.h"
-
-extern void o_worker_handle_interrupts(void);
-
-#if PG_VERSION_NUM >= 180000
-#define O_STARTUP_PROC_INTERRUPTS() ProcessStartupProcInterrupts()
-#else
-#define O_STARTUP_PROC_INTERRUPTS() HandleStartupProcInterrupts()
-#endif
-
 /*
  * Like CHECK_FOR_INTERRUPTS(), but backend-aware: the startup process and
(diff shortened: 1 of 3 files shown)


pashkinelfe (approved):

src/recovery/recovery.c
```diff
@@ -5044,6 +5044,8 @@ workers_synchronize(XLogRecPtr ptr, bool send_synchronize)
 			BgwHandleStatus status;
 			pid_t		pid;
 
+			CHECK_FOR_INTERRUPTS();
```
za-arthur: Need to check if `CHECK_FOR_INTERRUPTS` suitable here. It might not be:
https://github.com/orioledb/postgres/blob/patches17_20/src/backend/access/transam/xlogrecovery.c#L1784

za-arthur (changes requested):

pgnickb:
After further internal discussion we agreed to make a proper macro that would prevent this kind of mistake in the future. Will work on it and publish soon

include/workers/interrupt.h
```diff
@@ -14,6 +14,30 @@
...
+#else
+#define O_STARTUP_PROC_INTERRUPTS() HandleStartupProcInterrupts()
+#endif
+
+/*
+ * Like CHECK_FOR_INTERRUPTS(), but backend-aware: the startup process and
+ * orioledb's own background workers (recovery/S3/bgwriter workers) redefine
+ * signal handling and don't respond to a plain CHECK_FOR_INTERRUPTS().
+ */
+#define O_CHECK_FOR_INTERRUPTS() \
```
za-arthur: You can re-use existing function `o_worker_handle_interrupts()` and add this logic into it instead of adding new macros.

za-arthur (approved):
LGTM

PR #991, opened 2026-07-27
Add comment for lossy IOS filter in gist test

pashkinelfe:


akorotkov:
Yes, we don't support IOS because we don't have a visibility map.
We could potentially scan only the external index + bridge index instead of the external index + bridge index + primary key, and call it an "index-only scan". But that's not implemented.

pashkinelfe:
> Yes, we don't support IOS because we don't have a visibility map. We could potentially scan only the external index + bridge index instead of the external index + bridge index + primary key, and call it an "index-only scan". But that's not implemented.

I agree. So that we can't currently have mentioned IOS upstream anomaly

pgnickb (approved):
LGTM

PR #995, opened 2026-07-29
recovery,checkpoint: fuzzy sys-tree checkpoint with a front consistency stage

akorotkov:
## Summary

Make the OrioleDB system trees a **fuzzy checkpoint image** reconciled by a
**front sys-tree-consistency recovery stage**, instead of holding
`oTablesMetaLock` across replay / across the checkpoint's commit drain. This:

- Fixes the `027_stream_regress` standby self-deadlock (leader no longer holds
  `oTablesMetaLock` across replay, so an interleaved `dbase_redo` can absorb its
  `ProcSignalBarrier`).
- Fixes the checkpoint `ON COMMIT DROP` AB-BA deadlock (no meta locks held across
  the commit wait).
- Fixes an SK/PK crash-recovery divergence exposed by the fuzzy checkpoint:
  `test_recovery_sk_cross_row_rotation` under `dm_log_writes` OS-buffer-loss
  recovery left one spurious secondary-index entry. Root cause was recording
  only sys-tree undo during the sys-tree stage, which dropped an aborting
  transaction's data-log undo chain from the xids file. Now all undo logs are
  recorded and the descriptor-before-sys-trees hazard is handled by a
  recovery-side deferral (`deferred_checkpoint_undo`, drained at the
  `replayStartPtr` crossing).

Split out of #989 as an independent change. Full rationale in the commit
message.

**Validation:** `dm_log_writes` x64 harness = **64/64 pass (0 failures)** for
`test_recovery_sk_cross_row_rotation` (the folded redesign was 6-7/64). Plus
`recovery_test`, `replication_test`, `meta_lock_deadlock_test`,
`systree_stage_test`, `systree_xids_gap_test`, `systree_wal_retention_gap_test`.

## Details

recovery,checkpoint: fuzzy sys-tree checkpoint with a front consistency stage

Fix the 027_stream_regress standby self-deadlock: the recovery leader no longer
holds oTablesMetaLock across replay, so an interleaved dbase_redo can absorb its
ProcSignalBarrier.  Sys trees become a fuzzy checkpoint image reconciled by a
front sys-tree-consistency recovery stage.

Checkpoint (o_perform_checkpoint):
- Capture the sys trees as a fuzzy per-page image WITHOUT the EXCLUSIVE
  oTablesMetaLock/oSysTreesLock freeze.  Pin sysTreesStartPtr =
  get_checkpoint_xlog_ptr() right before the sys-tree walk and replayStartPtr
  after it.
- Right after pinning sysTreesStartPtr, wait_finish_active_commits() waits for
  every commit stamped at or before it to finish, so no such commit is still in
  flight when the in-progress xids are snapshotted later.  Otherwise a
  committing transaction could be snapshotted as in-progress while its commit
  WAL record sits before the replay window (which starts at sysTreesStartPtr)
  and is never replayed, and recovery -- seeing its undo stack in the xids file
  with no commit -- would wrongly abort it.  We hold no meta locks across the
  wait, so the ON COMMIT DROP AB-BA deadlock does not arise.
- No clamp on sysTreesStartPtr is needed.  A sys-tree WAL record already durable
  at the pin is already reflected on its page (a buffer is modified before its
  XLogInsert() returns); anything from there on falls inside the forward-replay
  window or is caught by the checkpoint's xids file -- the same two mechanisms
  the PK/data walk already relies on.  (An earlier oldest-open-DDL-window clamp
  was unenforceable -- Postgres's WAL retention and its single redo pass both
  key off checkPointRedo, never a clamped-back pointer -- as well as
  unnecessary, so it is not introduced.)
- Arm the xids-queue flush (before_writing_xids_file/start_write_xids) BEFORE
  checkpoint_sys_trees(), mirroring the PK/data walk, so a transaction that
  finishes while the fuzzy walk runs is recorded in the xids file and reconciled
  at recovery instead of being baked into the image.
- Add a checkpoint_before_replay_start stop event between the two pins so tests
  can land WAL in the [sysTreesStartPtr, replayStartPtr) stage-0 window.

xids file records all undo logs.  before_writing_xids_file/start_write_xids
guarantee an in-flight transaction's undo chains land in the xids file so
recovery can revert them.  Recording only sys-tree undo during the sys-tree
stage would drop an aborting transaction's data-log undo, whose page the fuzzy
walk may capture before the abort's (time-unbounded) undo reverts it, leaving
that change unrevertible on recovery.  The descriptor-before-sys-trees hazard
that full recording would otherwise create is handled by the recovery-side
deferral below, not by restricting what is queued.

Recovery:
- Apply sys-tree records in place from sysTreesStartPtr (the overwrite recovery
  callbacks are idempotent against the fuzzy image); apply data (PK/toast/SK)
  records only from replayStartPtr.  Keep o_tables_meta_locked tracking -- it
  guards handle_o_tables_meta_unlock()'s reconciliation.
- Drop the two stale exact-cut Assert(o_tables_meta_locked) checks: with a fuzzy
  checkpoint recovery can start past a window's META_LOCK, replaying a UNLOCK /
  O_TABLES modify with the flag unset; handle_o_tables_meta_unlock returns early
  and the modify applies idempotently.
- Defer stage-0 data-log undo.  A checkpoint xid (read from the xids file into
  checkpoint_undo_stacks) can finish -- commit, abort, or roll back to a
  savepoint -- inside stage 0, but applying its captured UndoLogRegular /
  UndoLogRegularPageLevel undo there would fetch a data descriptor before the
  sys trees are consistent.  walk_checkpoint_stacks() parks such an entry on
  deferred_checkpoint_undo and orioledb_redo() drains it at the replayStartPtr
  crossing (UndoLogSystem undo still applies in place).  The current oxid's own
  rebuilt data undo stack stays empty in stage 0 (its data modifies <
  replayStartPtr are skipped), so only the checkpoint-captured stack needs it.

Tests (TESTGRESCHECKS): meta_lock_deadlock_test (the 027_stream_regress standby
self-deadlock), systree_stage_test (park at checkpoint_before_replay_start,
abort a data txn in the stage-0 window, then crash+recover), systree_xids_gap_test
(DDL rollback racing the xids-arm gap) and systree_wal_retention_gap_test (DDL
window spanning an ordinary checkpoint, recovering with no clamp).

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -196,6 +196,7 @@ TESTGRESCHECKS_PART_1 = test/t/amcheck_test.py \
 						test/t/recovery_opclass_test.py \
 						test/t/recovery_worker_test.py \
 						test/t/replication_test.py \
+						test/t/fuzzy_checkpoint_test.py \
 						test/t/types_test.py \
 						test/t/undo_eviction_test.py \
 						test/t/undo_ring_keep_checkpoint_test.py \
--- a/include/orioledb.h
+++ b/include/orioledb.h
@@ -251,6 +251,7 @@ typedef struct
 {
 	UndoRetainSharedLocations undoRetainLocations[(int) UndoLogsCount];
 	pg_atomic_uint64 commitInProgressXlogLocation;
+
 	int			autonomousNestingLevel;
 	LWLock		undoStackLocationsFlushLock;
 	bool		flushUndoLocations;
--- a/src/catalog/o_tables.c
+++ b/src/catalog/o_tables.c
@@ -31,6 +31,7 @@
 #include "tuple/toast.h"
 #include "utils/elog.h"
 #include "utils/planner.h"
+#include "utils/stopevent.h"
 
 #include "access/hash.h"
 #include "access/heapam.h"
@@ -2445,6 +2446,18 @@ o_tables_meta_unlock(ORelOids oids, Oid oldRelnode)
 {
 	if (!is_recovery_process())
 	{
+		/*
+		 * Test hook: park here holding oTablesMetaLock, after the
+		 * WAL_REC_O_TABLES_META_LOCK (+ the sys-tree modify) container has
+		 * already been flushed by systrees_modify_end but before
+		 * WAL_REC_O_TABLES_META_UNLOCK is written.  A concurrent DROP
+		 * DATABASE issued while we are parked then lands its dbase_redo WAL
+		 * record between META_LOCK and META_UNLOCK -- the interleaving that
+		 * makes the standby recovery leader replay dbase_redo while holding
+		 * the meta lock.
+		 */
+		STOPEVENT(STOPEVENT_BEFORE_O_TABLES_META_UNLOCK, NULL);
+
 		add_o_tables_meta_unlock_wal_record(oids, oldRelnode);
 		(void) flush_local_wal(false, false);
 
--- a/src/checkpoint/checkpoint.c
+++ b/src/checkpoint/checkpoint.c
@@ -670,7 +670,7 @@ add_index_id_item(List *list, BTreeDescr *desc)
  * `redo_pos`.
  *
  * Ensures every transaction, which WAL record is written before redo_pos,
- * passed to write_to_xids_queue(),
+ * don't go to write_to_xids_queue(),
  */
 static inline void
 wait_finish_active_commits(XLogRecPtr redo_pos)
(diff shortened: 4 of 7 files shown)


pgnickb (commented):
This iteration looks good.

I'd like to later go over some of these comments and rewrite them into more human-friendly ones. Also perhaps worth updating/adding docs, as this might be relevant.

pgnickb (approved):

PR #998, opened 2026-08-03
Fix DROP INDEX CONCURRENT fail by timeout

bwyltcaxe:
This patch moves locks for trees to-be-dropped to PRE COMMIT stage: this lets DROP INDEX CONCURRENT which holds ShareUpdateExclusiveLock to be canceled by request or timeout.  During COMMIT phase interrupt are switched off on PG core's side, so backend would wait forever.

### DIFF
--- a/src/btree/undo.c
+++ b/src/btree/undo.c
@@ -938,6 +938,33 @@ btree_relnode_undo_callback(UndoLogType undoType, UndoLocation location,
 			for (i = 0; i < numTrees; i++)
 				fsync_btree_files(trees[i]);
 		}
+
+		/*
+		 * Acquire here locks the commit stage below needs to drop the old
+		 * trees.
+		 *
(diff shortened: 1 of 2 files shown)


pashkinelfe:
Ci tests fail on this PR @bwyltcaxe can you update?

bwyltcaxe:
> Ci tests fail on this PR @bwyltcaxe can you update?

Yeah, sure. Sorry for silence: I do remember about this PR, just didn't have enough time to look again yet :(

bwyltcaxe:
Looks like I fixed it (excluding Bad Gateway -- but it looks like I can do nothing about it). Removed `o_tables_rel_lock_extended_no_inval` -- for 1500 tables the locks become exhausted.

I am not sure this is the best fix, I'd like to prefer to make non-blocking-deferrable approach, but it looks like overkill too, so I have no better idea here

homper (approved):

PR #1011, opened 2026-08-05
Parallel index scan support (rebased onto rewind-removal)

akorotkov:
Re-creation of #863 (author @fatih-akatlar). #1010 is merged, so this now sits directly on `main`.

**History reworked for review.** The original 22-commit history is preserved unchanged at `parallel-index-scan-on-rewind-orig-20260817`; the tree it produced was verified byte-identical to the reworked one before the rebase (same tree hash `f10e0015`), and the diff against `main` is the same 30 files / +4559 / -890.

Five commits now, in review order:

1. **Fix index-only scan cost estimation by setting allvisfrac to 1** — independent of everything else, but it moves plans, which is why so many expected outputs change.
2. **Fix backward array (SAOP) index scan dropping all but the smallest element** — an independent bug fix that was buried mid-branch.
3. **Add parallel index scan support** — the core, with the range-filtering and `fullyInRange` refinements folded in.
4. **Parallel ordered index scan** — forward and backward, lokey derivation at arbitrary depth, cost-based path selection.
5. **Regenerate expected outputs** — PG16/PG17 variants, no code.

Three sequences of pure churn were collapsed rather than replayed, since a reviewer should not have to read a change and then its undo:

- `_bt_preprocess_keys()` in `o_rescan_custom_scan()` was added by the parallel-scan commit and removed ten commits later; it is now simply never added, and only the comment explaining why it must not be called remains.
- `orioledb.debug_parallel_ordered_scan` was added and then removed; net zero in `include/orioledb.h`.
- The serial ordered scan engine and its debug SRF were added and then removed; net zero in `src/tableam/func.c` and `sql/orioledb--1.8--1.9_dev.sql`.

### DIFF
--- a/src/btree/scan.c
+++ b/src/btree/scan.c
@@ -3316,6 +3316,24 @@ btree_seq_scan_set_range_filter(BTreeSeqScan *scan,
 }
 
 /*
+ * Attach scan callbacks after construction.
+ *
+ * make_btree_seq_scan_cb() takes them up front, but a parallel scan is built
+ * by make_btree_seq_scan() and only then learns what it has to filter.  Only
+ * meaningful before the first getnext, and it deliberately does not turn on
+ * the partial-leaf (FETCH) walk: that is reserved for the non-parallel
+ * key-directed scan the constructor recognises.
+ */
+void
+btree_seq_scan_set_callbacks(BTreeSeqScan *scan,
+							 BTreeSeqScanCallbacks *cb, void *arg)
+{
+	Assert(!scan->initialized);
+	scan->cb = cb;
+	scan->arg = arg;
+}
+
+/*
  * Enable ordered (key-order) scanning.  Must be called before the first
  * getnext.  In ordered mode on-disk downlinks are read inline in key order
  * (see BTreeSeqScan.ordered).  Both forward and backward directions are
--- a/src/btree/scan.c
+++ b/src/btree/scan.c
@@ -1538,9 +1538,26 @@ check_downlink_in_scan_range(BTreeSeqScan *scan, OTuple keyRangeLow,
 }
 
 /*
+ * Does this side of the scan's qualification range constrain anything?
+ *
+ * The first unbounded key makes the whole bound infinite:
+ * o_idx_cmp_key_bound_to_tuple() answers on it without looking at the rest.
+ */
+static inline bool
+scan_range_side_is_open(OBTreeKeyBound *bound)
+{
+	return bound->nkeys == 0 ||
+		(bound->keys[0].flags & O_VALUE_BOUND_UNBOUNDED) != 0;
+}
+
+/*
  * Checks if a page's entire key range is within the scan's qualification
  * range.  When this returns true, every tuple on the page is guaranteed to
  * be in range, so per-tuple checks can be skipped entirely.
+ *
+ * A page unbounded on one side (the leftmost or the rightmost page of the
+ * tree) is still fully in range when the scan is unbounded on that side too,
+ * which is the common case for a one-sided qualification such as "id > 100".
  */
 static bool
 page_fully_in_scan_range(BTreeSeqScan *scan, OTuple keyRangeLow,
@@ -1561,8 +1578,8 @@ page_fully_in_scan_range(BTreeSeqScan *scan, OTuple keyRangeLow,
 		if (cmp < 0)
 			return false;
 	}
-	else
-		return false;			/* page starts at -infinity */
+	else if (!scan_range_side_is_open(&scanRange->low))
+		return false;			/* page starts at -infinity, the scan does not */
 
 	if (!O_TUPLE_IS_NULL(keyRangeHigh))
 	{
@@ -1574,8 +1591,8 @@ page_fully_in_scan_range(BTreeSeqScan *scan, OTuple keyRangeLow,
 		if (cmp > 0)
 			return false;
 	}
-	else
-		return false;			/* page ends at +infinity */
+	else if (!scan_range_side_is_open(&scanRange->high))
+		return false;			/* page ends at +infinity, the scan does not */
 
 	return true;
 }
--- a/src/btree/scan.c
+++ b/src/btree/scan.c
@@ -227,6 +227,27 @@ struct BTreeSeqScan
 	OBTreeKeyRange *scanRange;
 
 	/*
+	 * Ordered (key-order) scan.  When set, on-disk downlinks are read inline
+	 * during the internal-page walk (in strict key order) instead of being
+	 * deferred to a physically-sorted disk phase, so every tuple this scan
+	 * emits is in scanDir order.  This lets a parallel scan feed Gather Merge
+	 * (each worker's stream is individually sorted) at the cost of losing the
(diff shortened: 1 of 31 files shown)


src/btree/scan.c
```diff
@@ -1438,6 +1492,491 @@ check_in_memory_leaf_page(BTreeSeqScan *scan, OTuple keyRangeLow, OTuple keyRang
...
+		int			cmp;
+
+		cmp = o_btree_cmp(scan->desc,
+						  &keyRangeLow, BTreeKeyNonLeafKey,
+						  &scanRange->low, BTreeKeyBound);
+		if (cmp < 0)
+			return false;
+	}
+	else
+		return false;			/* page starts at -infinity */
```
pgnickb: Nit: it could be useful to actually handle the unbounded case as well instead of skipping the edge pages completely. 

I'm not sure if effects of this can ever be significant.

src/tableam/scan.c
```diff
@@ -376,14 +474,118 @@ orioledb_set_rel_pathlist_hook(PlannerInfo *root, RelOptInfo *rel,
...
+								ordered->parallel_aware = true;
+								ordered->parallel_workers = parallel_workers;
+
+								/*
+								 * Penalise the inline on-disk reads: roughly
+								 * one random access per selected index page
+								 * instead of the sequential disk phase,
+								 * spread over the workers.
+								 */
+								pages = ix_path->indexinfo->pages *
```
pgnickb: Nit: We do the same computation in `o_estimate_parallel_workers`. Shall this be generalized somehow to make it more future proof?

src/tableam/index_scan.c
```diff
@@ -1027,6 +1042,170 @@ o_index_scan_getnext(OTableDescr *descr, OScanState *ostate,
...
+
+		if (ostate->ixNum == PrimaryIndexNumber || ostate->onlyCurIx)
+		{
+			tts_orioledb_store_tuple(ss->ss_ScanTupleSlot, tuple,
+									 descr, tupleCsn, ostate->ixNum,
+									 true, &hint);
+		}
+		else
+		{
+			OBTreeKeyBound bound;
```
pgnickb, line 1348: This is more or less exact copy of similar logic in `o_index_scan_getnext`. I'd add a comment mentioning that and that both should be kept in sync.

akorotkov:
## SAOP and skip scan: how PostgreSQL does it, and where we stood

### PostgreSQL

nbtree keeps the primitive-scan structure in parallel mode. `BTParallelScanDescData`
carries `btps_arrElems[]` (the `BTArrayKeyInfo.cur_elem` offset of every SAOP array)
and, after PG18, a flattened datum per skip array. A worker that exhausts the current
primitive scan calls `_bt_parallel_primscan_schedule()` to publish the next array
state; the next worker to seize the scan restores it with
`_bt_parallel_restore_arrays()` and does a fresh descent. So a parallel `IN (1, 10^6)`
descends twice, exactly like the serial scan, and skip arrays behave the same way.

### What our parallel scan did

`o_exec_parallel_idx_scan_load_keyrange()` passes `numPrefixExactKeys = 0`, which makes
`o_key_data_to_key_range()` take the union branch for every array: `low = elem_values[0]`,
`high = elem_values[num_elems - 1]`. One range, one scan, membership checked per tuple.
The reviewer's reading is correct, and the cost model does not see it -- `transform_path()`
copies core's `IndexPath` costs verbatim, and both `o_estimate_parallel_workers()` and the
ordered-path penalty use `indexselectivity`, the selectivity of the array membership, not
the width of the union that is actually scanned.

Measured on a 2M-row table, `k = id % 100`, index on `k`, `WHERE k IN (0, 99)`:

| | plan | time |
|---|---|---|
| `max_parallel_workers_per_gather = 0` | Index Only Scan | 50.8 ms |
| `= 4` | Parallel Custom Scan (o_scan) | 107.9 ms |

The planner picked the parallel plan and it was 2.1x slower with five processes.

### Fix pushed (2a67b969)

We cannot do per-element descents without a shared array-key state, but we do not have to
read the leaves either: a leaf whose key range holds no array element cannot hold a
matching row. The scan already has an `isRangeValid` downlink callback, consulted before
both the in-memory and the on-disk branch, so the filter is ~150 lines and no coordination
change. Only the level-1 internal chain -- a small fraction of the tree -- is still walked
in full.

Restricted to an array on the **leading** index column: an array on a later column is
interleaved with every value of the leading one, so page granularity buys nothing there.
A leading skip array does not enumerate its elements and is left alone.

A/B/A/B, full rebuild per phase, best of 3, same query:

| build | serial | 4 workers |
|---|---|---|
| without filter | 50.5 ms | 187.8 ms |
| **with filter** | 53.0 ms | **17.7 ms** |
| without filter | 50.5 ms | 173.1 ms |
| **with filter** | 48.4 ms | **18.4 ms** |

From 3.5x slower than serial to 2.7x faster; every phase returns the same 40000 rows.
`parallel_scan.sql` gains a sparse-list case on an ASC and on a DESC index (the element
order flips, so the page test has to compare in index order), each checked against the
serial result.

### Skip scan: the problem is not the parallel path

The reviewer's premise was that the serial path has skip-scan logic that the parallel path
lacks. The logic is there, and it is armed -- an instrumented build shows `so->skipScan = 1`,
`numArrayKeys = 1`, `skipScanProbePending = 1` -- but it never narrows. On PG18, index on
`(a, b)`, `WHERE b = <val>`, serial, `enable_seqscan/bitmapscan = off`:

| leading-column cardinality | heap | orioledb |
|---|---|---|
| 3 distinct `a` | 0.88 ms, `Index Searches: 5` | 638.5 ms, 1 range switch |
| 100 distinct `a` | 3.27 ms, `Index Searches: 101` | 63.8 ms, 1 range switch |

`switch_to_next_range()` is entered exactly twice for the whole scan instead of once per
distinct leading value, and the index is read end to end. So the serial scan is 20x-725x
behind PostgreSQL here, and a parallel guard would only have preserved that. I did write
one and then dropped it: with no skip-scan benefit to protect, it changed no plan I could
construct. Worth its own issue rather than a knob in this PR.

### Two suspected correctness bugs that did not reproduce

Contradictory quals (`k > 90 AND k < 5`) return 0 rows in the parallel plan, and a union
range whose high bound sits below the index maximum (`k IN (0, 50)`, disk phase after
CHECKPOINT + restart) returns exactly the serial row count. Neither the missing `qual_ok`
check nor early termination in the physically-ordered disk phase produced a wrong answer.

### Review comments (5c5068b0)

* `page_fully_in_scan_range()` gave up on a page unbounded on either side, so the leftmost
  and rightmost page of every range-filtered scan paid per-tuple checks even when the scan
  was unbounded on that side too. An open page edge now counts as in range when the scan's
  bound on that side is open.
* The expected index-page count is now named (`o_index_path_pages()`) instead of spelled
  out in both the worker estimate and the ordered-path penalty.
* Cross-references added between the secondary-to-primary lookup in
  `o_exec_parallel_idx_scan()` and the one in `o_index_scan_getnext()`.

Local on this branch: `regresscheck` 49/49, `isolationcheck` 37/37,
`testgrescheck_part_1` 44/44.

pgnickb (approved):
After further discussion offline and the changes made I think this is ready for a merge.

PR #1032, opened 2026-08-13
Don't build an index descriptor from a torn OIndex/OTable pair

akorotkov:
`Assert("tupdesc->natts == all_attrs")`, `o_tables.c:1009`, in the **standby checkpointer**:

```
CreateRestartPoint -> o_perform_checkpoint -> o_indices_foreach_oids
  -> checkpoint_tables_callback -> o_fetch_index_descr
  -> o_index_fill_descr -> o_tupdesc_load_constr
```

Twice on `arm64 / PG18 / pg_tests` in three days — runs
[31498780558](https://github.com/orioledb/orioledb/actions/runs/31498780558) (`all_attrs=2`, `fields_start=1`) and
[31616384031](https://github.com/orioledb/orioledb/actions/runs/31616384031) (`all_attrs=3`, `fields_start=0`) — both
`postgres: standby_1: checkpointer`. This is the assert fabd3eac added on the
suspicion that the two records could diverge. They do.

### What is torn

A primary index descriptor is assembled from two sys-tree rows:

* the leaf tuple descriptor from the **OIndex** record, and
* `missing[]` plus `primary_init_nfields` from the **OTable** record.

A DDL writes them one after the other — `o_indices_update()` first,
`o_tables_update()` second — and a caller that names the table by OIDs alone
re-reads both here under `o_non_deleted_snapshot`, which shows uncommitted
rows. The read can therefore land between the two writes and pair a new OIndex
with the old OTable.

### Why only the standby

On the primary the window is closed by `oTablesMetaLock`: the DDL holds it
SHARED across both writes, and `o_perform_checkpoint()` takes it EXCLUSIVE
around the whole `o_indices_foreach_oids()` walk. Recovery does not
participate — since 4e0aa762 `WAL_REC_O_TABLES_META_LOCK` deliberately does not
take the lock, because the enclosed modifies are applied in place against the
fuzzy sys-tree image — so on a standby a restartpoint walks straight into the
window.

### Why it matters beyond the assert

A torn pair is not merely inaccurate:

* `o_tupdesc_load_constr()` sizes `missing[]` from the OTable while
  `FreeTupleDesc()` frees it entry by entry according to the descriptor — the
  `pfree()` of a bogus pointer the assert's comment predicted, and a very
  plausible root for the rare checkpointer `FreeTupleDesc` MAXALIGN crash;
* `fillFixedFormatSpec()` reads `primary_init_nfields` attributes out of a
  descriptor that may be shorter — a plain out-of-bounds read.

---

## The fix

The pair carries its own answer. `make_primary_o_index()` and
`make_ctid_o_index()` stamp the new OIndex and the OTable's
`primary_ixversion` with the same value, and every `o_indices_update()` call
site writes the OTable immediately after. Equal versions mean one incarnation;
unequal ones mean exactly the window above. So the check is a single
comparison of `OIndexDescr.version` against `OTable.primary_ixversion`.

That needs the version to actually reach the descriptor, which it did not:

### 1. The incarnation version never reached the descriptor (f116eb09)

`deserialize_o_index()` took the version from the key the caller *asked* with.
OID-only callers ask with `O_TABLE_INVALID_VERSION`, so `OIndexDescr.version`
was version-unknown on every such path — nothing to compare. `*found_key` is
the key of the chunk actually read and carries the real version, exactly as
`o_tables_get_extended()` already uses for OTable. Only the version is taken
from it: its OIDs are not necessarily the requested ones (see the note below),
and callers identify a descriptor by the OIDs they asked for.

This also sharpens `get_index_descr()`'s cache check: until now a descriptor
built by an OID-only fetch stored version-unknown, so a later fetch naming a
version never matched it and rebuilt the descriptor every time.

### 2. The version check itself (ea5b5fe2)

Replaces the structural comparison (`natts`, per-field `byval`/`typlen`) the
first commit used. Structural equality is enough for memory safety but is not
what "the same incarnation" means — an `ALTER COLUMN TYPE` between two
by-value 4-byte types produces a torn pair it accepts. The structural
predicate stays as the assertion inside `o_tupdesc_load_constr()`, now static
and assert-only, since equal versions imply it.

**Over the regression suite the version check refuses five pairs the structural
one passed — every one an OIndex exactly one incarnation ahead of its OTable.**

---

## Dropped from this PR: making reloid part of the OIndex key

An earlier revision of this branch also made `o_index_chunk_cmp()` compare
`reloid`, because `(datoid, spcoid, relnode)` is not a relation identity —
relfilenumbers are reused, and a lookup for an index with no record of its own
silently resolves to a neighbour's record on the same relnode, relabelled with
the requested OIDs. `o_tables_oids_indexes()` leans on exactly that when ALTER
TABLE reuses an index tree under a new `pg_class` entry (its `reuse_relnode`
branch skips both the delete and the add), so the change came with an
`o_indices_rekey()` to move the record to the new reloid.

**That does not work, and the reason is worth recording.** A churn run of this
branch ([31720227564](https://github.com/orioledb/orioledb/actions/runs/31720227564))
came back 32/32 `pg_tests` and 4/4 `dm_log_writes` green and **8/8
`check_page` red** — deterministic, not a flake:
`test_alter_column_type` (`indices_build_test.py:552`) reads the table empty
after `ALTER COLUMN TYPE` + `stop -m immediate` + restart.

Reduced to a two-statement repro, the tell is decisive: with a `CHECKPOINT`
between the ALTER and the crash the rows survive; without one they are gone.
So the loss is in **WAL replay**. Records written before the ALTER name the
tree by its old reloid, and once the OIndex record moves and the comparator
starts separating reloids, recovery can no longer resolve them and drops their
data silently. Recovery *depends* on the reloid-blind lookup.

Making reloid part of the tree identity therefore needs recovery to resolve a
tree without it (or an old→new mapping), which is well outside this PR. Filed
separately; the commit is dropped here.

### Testing

regress 48/48, isolation 35/35, `indices_build_test` 10/10, and the
deterministic `restartpoint_ddl_window_test` added here, which parks the
recovery leader inside the DDL's sys-tree window and drives a restartpoint over
it: it kills the standby without the fix and passes with it.

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -230,6 +230,7 @@ TESTGRESCHECKS_PART_2 = test/t/checkpoint_concurrent_test.py \
 						test/t/recovery_item_rollback_test.py \
 						test/t/recovery_row_lock_test.py \
 						test/t/reindex_test.py \
+						test/t/restartpoint_ddl_window_test.py \
 						test/t/s3_test.py \
 						test/t/schema_test.py \
 						test/t/seq_scan_resowner_test.py \
--- a/src/catalog/o_indices.c
+++ b/src/catalog/o_indices.c
@@ -1379,9 +1379,53 @@ o_index_fill_descr(OIndexDescr *descr, OIndex *oIndex, void *o_table_source, OTa
 
 		if (oTable)
 		{
-			o_tupdesc_load_constr(descr->leafTupdesc, oTable, descr);
-			primary_init_nfields = palloc(sizeof(*primary_init_nfields));
-			*primary_init_nfields = oTable->primary_init_nfields;
+			/*
+			 * The leaf tuple descriptor above came from the OIndex record;
+			 * the constraints and primary_init_nfields come from the OTable
+			 * record.  Those are two separate sys-tree rows, and a DDL
+			 * updates them one after the other -- o_indices_update() first,
+			 * o_tables_update() second.
+			 *
+			 * A caller that handed us the OTable it already holds always sees
+			 * the pair whole.  A caller that named the table by OIDs alone
+			 * re-reads it here under o_non_deleted_snapshot, which shows
+			 * uncommitted rows, so it can catch a DDL between those two
+			 * writes and pair the new OIndex with the old OTable.  On the
+			 * primary oTablesMetaLock keeps the checkpointer out of that
+			 * window; recovery applies the enclosed modifies without taking
+			 * it, so on a standby a restartpoint walks right into it.
+			 *
+			 * A torn pair is not merely inaccurate.  o_tupdesc_load_constr()
+			 * would size missing[] from the OTable while FreeTupleDesc()
+			 * frees it by the descriptor, and fillFixedFormatSpec() would
+			 * read primary_init_nfields attributes out of a shorter
+			 * descriptor.  Everything those callers actually need comes from
+			 * the OIndex alone, so build that much and mark the descriptor
+			 * invalid: nobody gets handed it again, and the next fetch
+			 * re-reads a settled pair.
+			 *
+			 * The index incarnation version says whether the pair is whole.
+			 * make_primary_o_index()/make_ctid_o_index() stamp the new OIndex
+			 * and the OTable's primary_ixversion with the same value, and
+			 * every o_indices_update() call site writes the OTable right
+			 * after, so the two rows carry equal versions once both writes
+			 * have landed and unequal ones for exactly the window in between.
+			 * A caller that supplies the OTable it holds fetched the OIndex
+			 * by that same version (see o_table_descr_fill_indices()) and
+			 * oIndicesFetchCallback() returns only an exact version match, so
+			 * the equality holds there by construction.
+			 */
+			if (descr->version == oTable->primary_ixversion)
+			{
+				o_tupdesc_load_constr(descr->leafTupdesc, oTable, descr);
+				primary_init_nfields = palloc(sizeof(*primary_init_nfields));
+				*primary_init_nfields = oTable->primary_init_nfields;
+			}
+			else
+			{
+				Assert(source == oTableSourceContext);
+				descr->valid = false;
+			}
 			if (free_oTable)
 				o_table_free(oTable);
 		}
--- a/src/catalog/o_indices.c
+++ b/src/catalog/o_indices.c
@@ -1696,6 +1696,7 @@ o_indices_get_extended(ORelOids oids, OIndexType type,
 	{
 		OIndexChunkBoundKey boundKey;
 		OIndexChunkBoundKey *found_key = NULL;
+		OIndexChunkKey deserKey;
 		Size		dataLength;
 		Pointer		result;
 		OIndex	   *oIndex;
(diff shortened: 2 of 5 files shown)


akorotkov:
Added a deterministic test (`test/t/restartpoint_ddl_window_test.py`) and, with it, confirmation of the diagnosis.

It parks the recovery leader on the record that switches to the O_TABLES sys
tree — so an `ALTER TABLE ... ADD COLUMN`'s O_INDICES modifies are applied and
its O_TABLES ones are not — and then makes the standby take a restartpoint.
Nine seconds, deterministic: **without this PR's fix the standby aborts on
`Assert("tupdesc->natts == all_attrs")` on every run; with it, the test passes.**

Two details worth flagging:

* the primary `CHECKPOINT` is deliberately *before* the table exists. A standby
  cannot take a restartpoint until it has replayed a checkpoint record, and the
  restartpoint that record triggers must not be the one that caches this
  table's descriptor — `get_index_descr()` would then hand the cached copy back
  instead of reading the two sys-tree rows again.
* the `replay_on_record` stop event now pushes the record's type and, for a
  `WAL_REC_RELATION` naming a sys tree, that tree's number, so a test can say
  *where* to stop. Its existing user arms it match-all and is unaffected.

Independently, the churn harness reproduced the original crash twice in ~35
minutes on `main` (standby `rep_pg.log`, `CreateRestartPoint` →
`o_perform_checkpoint` → ... → `o_tupdesc_load_constr`), and has been running
on this branch since without a hit.

akorotkov:
Scope, after grepping every failed job of the last three days: this assert is
the dominant CI flake, not a PG18 curiosity. **Four hits across four different
branches, two majors and two check types:**

| run | branch | cell |
|---|---|---|
| [31477764124](https://github.com/orioledb/orioledb/actions/runs/31477764124) | `some_postgres_patches_improvements` | arm64, **17**, gcc, pg_tests |
| [31498780558](https://github.com/orioledb/orioledb/actions/runs/31498780558) | `main` | arm64, 18, clang, pg_tests |
| [31593135253](https://github.com/orioledb/orioledb/actions/runs/31593135253) | `seqscan-descr-pin-dropped-twice` | arm64, 18, clang, **sanitize** |
| [31616384031](https://github.com/orioledb/orioledb/actions/runs/31616384031) | `main` | arm64, 18, gcc, pg_tests |

Every one of them is `postgres: standby_1: checkpointer` in
`CreateRestartPoint`, which is exactly what the fix and the new test address.

akorotkov:
Churn validation so far: **24 completed iterations of the `pg_tests` workload
(primary + standby + create/drop-database churn + PG regress/isolation/
subscription) across 3 parallel instances, 76 minutes, zero hits.**

The baseline for comparison is the same harness on `main`, which produced the
assert **twice in ~35 minutes** (both `rep_pg.log`, both `CreateRestartPoint`).

The branch under churn is `main` + this PR + #1033, so the run is also evidence
that neither change regresses that workload. Leaving it running.

akorotkov:
The one red cell (`arm64/17/gcc valgrind_2`) is the `recovery_heap_verdict_test`
flake from `main` that #1031 fixes — same `'committed oxid' not found`, nothing
to do with this branch.

Worth recording from the same logs: **the new standby test passes under
valgrind in every valgrind_2 cell**, ~17 s each, including inside the job that
went red on the flake above. So the park-replay + restartpoint choreography
holds up at valgrind timings, not just locally.

PR #1033, opened 2026-08-13
Start a split's historical page walk at the reading page's own low bound

akorotkov:
`merge_test.test_split_diff_old_snapshot_read` failed **`5023 != 5000`** on
`valgrind_1` (amd64/18/gcc, [run 31606843432](https://github.com/orioledb/orioledb/actions/runs/31606843432)):
an old REPEATABLE READ snapshot counted 23 rows too many. `count(DISTINCT id)`
was 5000, so nothing invisible leaked in — **a sequential scan handed out 23
keys twice**.

### What goes wrong

A scan in flight keeps a split from taking the differential shortcut, so every
split under it leaves a *full* page image in undo, and both halves point at it.
Reading a half under an older snapshot reconstructs that whole pre-split page,
so the walk has to start at the half's own low boundary — the rest belongs to a
sibling the scan visits separately. `get_page_from_undo()` reports that
boundary through `page_lokey`, deciding which side the caller is on by
comparing the caller's hikey with the **pre-split page's hikey** — a test that
only tells the two halves of that one split apart.

One INSERT splits a page repeatedly. A leaf that came from splitting the right
half *again* ends before the pre-split hikey, so it read as the left half, took
no low boundary, and walked the reconstructed page from its very first item —
re-emitting every key that had gone to its left sibling, which the scan had
already returned from that sibling's own downlink.

Instrumenting the scan showed it directly — three leaves rebuilding the same
image with hikey 1770, and the middle one starting at 1666:

```
SDPAGE blkno=2074 low=1666 high=1712 livehi=1712 hist=1 histhi=1770 histfirst=1666
SDPAGE blkno=2187 low=1712 high=1741 livehi=1741 hist=1 histhi=1770 histfirst=1666  <-- 
SDPAGE blkno=2188 low=1741 high=1770 livehi=1770 hist=1 histhi=1770 histfirst=1742
```

### The fix

A page descends from exactly one half of any split in its history, so its hikey
tells the two apart on its own: at or before the split key is the left side,
after it the right (rightmost meaning unbounded, hence right). Use that in both
the full-image path and in `reconstruct_split_diff()`, which tested for a hikey
merely *different* from the split key — the same mistake mirrored, for a left
half that was split again.

The differential path already reported the boundary, which is why the failure
needed a full image and so only showed up with a scan running.

### Test

`test_split_diff_scan_during_split` loops a scan on an old snapshot while 5000
interleaved keys split the leaves under it, and checks no key comes back twice.
It fails 3 runs out of 3 before the fix; after it, 120 iterations of the same
scenario are clean. regress (48), isolation (34) and merge_test (12) pass.

### DIFF
--- a/src/btree/undo.c
+++ b/src/btree/undo.c
@@ -1346,9 +1346,17 @@ reconstruct_split_diff(BTreeDescr *desc, UndoLocation undoLocation,
 	splitKey.tuple.data = splitKey.fixedData;
 
 	/*
-	 * Determine which half `dest` is from its own hikey: the left half's
-	 * hikey is exactly the split key; the right half's hikey is the original
-	 * page's (larger) hikey, or it is rightmost.
+	 * Determine which side of the split `dest` is on from its own hikey.  Its
+	 * key range lies wholly within one half -- a page is a descendant of
+	 * exactly one of them -- so everything left of the split ends at or
+	 * before the split key and everything right of it ends after.  Being
+	 * rightmost means an unbounded hikey, hence the right side.
+	 *
+	 * Testing for a hikey merely *different* from the split key would only
+	 * tell the two halves of this very split apart: once the left half is
+	 * itself split, its left part ends before the split key and would read as
+	 * the right half, taking the split key as a low boundary that lies past
+	 * its whole range.
 	 */
 	if (O_PAGE_IS(dest, RIGHTMOST))
 		rightHalf = true;
@@ -1358,7 +1366,7 @@ reconstruct_split_diff(BTreeDescr *desc, UndoLocation undoLocation,
 
 		BTREE_PAGE_GET_HIKEY(destHikey, dest);
 		rightHalf = (o_btree_cmp(desc, &destHikey, BTreeKeyNonLeafKey,
-								 &splitKey.tuple, BTreeKeyNonLeafKey) != 0);
+								 &splitKey.tuple, BTreeKeyNonLeafKey) > 0);
 	}
 
 	if (is_left != NULL)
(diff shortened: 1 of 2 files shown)


akorotkov:
One adjacent thing this PR deliberately does **not** change, noted so it is not
lost: in the full-image Split branch `get_page_from_undo()` sets only
`page_lokey`, never `lokey`. `lokey` is what a **backward** iterator asks for
(`undo_it_find_internal()` passes it for `BackwardScanDirection` only), so a
backward index scan reading a half through a full split image gets no low
boundary at all — the same shape of over-read this PR fixes for the seq scan,
one path over.

I have no reproducer for it and the failing test does not cover backward index
scans through a split, so changing it here would be a guess. Flagging it as the
next place to look if an ordered scan is ever seen returning a row twice.

akorotkov:
`check (arm64, 16, gcc, valgrind_2)` failed, and it is **not** this branch:
`test.t.recovery_heap_verdict_test` again, with the same signature as before —

```
01:15:33 checkpoint starting: time
01:15:43 checkpoint complete: ... redo lsn=0/1600DC0
01:15:50 redo starts at 0/1600DC0        <-- past the orphan oxid
```

a timed checkpoint moved the redo point past the records the test needs
replayed, because its 200 updates are still 200 psql processes at ~1.4 s each
under valgrind. That is exactly what #1031 fixes, and #1031 is not merged, so
`main` still carries it and every PR's valgrind_2 can draw it.

Nothing to do here; re-running this cell after #1031 lands is enough.

akorotkov:
The other two valgrind_2 failures (arm64/16/clang, arm64/18/gcc) are the same
`recovery_heap_verdict_test` flake as the first — three cells of this run, none
of them this branch's code. #1031 fixes it and is 120/120 green.

akorotkov:
Correcting my note above: I chased that `lokey` gap and **the theory is wrong**.

I instrumented every consumer of `UndoIterator.lokey` to warn when it reads a
value no branch had written, and reproduced a backward-scan anomaly with **zero
warnings** — so the full-image Split branch not setting `lokey` is not what
bites. (`undo_it_find_internal()` marks such an image `leftmost`, which makes
`undo_it_next_page()` stop before it would use the unset key.) The staleness
still looks worth closing on its own, but it is not a live defect on this path.

What the hunt did turn up is a real and **pre-existing** bug, filed as #1036: a
backward ordered index scan on an old snapshot silently loses a contiguous
range of rows while the leaves under it split, with forward index scans and
sequential scans on the same snapshot at the same instant returning everything.
Plain `main` reproduces it, so it is neither this PR's doing nor this PR's to
fix.

PR #1034, opened 2026-08-13
ci: don't fail the run when the coveralls upload does

akorotkov:
`Failed to download coveralls binary or checksum (Linux).` turned the whole
check run red on `main` ([run 31612397180](https://github.com/orioledb/orioledb/actions/runs/31612397180))
after every test job had already passed. The action fetches its own binary at
run time, so its network hiccups say nothing about the code under test.

`fail-on-error: false` keeps the coverage upload best-effort.

### DIFF
--- a/.github/workflows/check.yml
+++ b/.github/workflows/check.yml
@@ -200,6 +200,11 @@ jobs:
         uses: coverallsapp/github-action@v2
         with:
           files: ./orioledb/coverage.info
+          # The action downloads its own binary at run time; when that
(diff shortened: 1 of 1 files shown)


akorotkov:
The two valgrind_2 failures are `recovery_heap_verdict_test`, the `main`-branch
flake that #1031 fixes — worth noting that this PR changes one line of a
workflow YAML and nothing else, so there is no plausible connection. #1031 is
120/120 green.

PR #1035, opened 2026-08-13
Only declare an S3 part loaded when the transition is ours to make

akorotkov:
`TRAP: failed Assert("status == S3PartStatusLoading")`, `src/s3/headers.c:786`,
in **`orioledb s3 worker 2`** on `valgrind_2`
([run 31606843432](https://github.com/orioledb/orioledb/actions/runs/31606843432)) —
one line after that worker logged that it had started. It killed the second
`node.start()` of `s3_test.test_s3_data_eviction`, and PG18's postmaster
followed with `pmchild freelist for backend type 6 is corrupt`.

`s3_header_mark_part_loaded()` moves a part from Loading to Loaded and asserts
it finds Loading — i.e. that its caller is the one that claimed the part. Two
callers never check.

**`s3process_task()`, for `S3TaskTypeReadFilePart`.** An s3 worker is registered
with `bgw_restart_time = 0` and records the task it is about to run in
`workers_locations[]`, so a worker that errors is restarted at once and replays
it — which is what a crash immediately after `started` looks like. The first
attempt only has to have died between marking the part loaded and clearing its
slot for the replay to find the part `Loaded`, abort, and abort again on the
same task.

**`perform_page_io_build()`**, where an allocation running into a part the file
did not have yet marks it present without a download. After a crash restart the
datafile length rewinds into a part the previous run had already uploaded and
evicted, so a reader can have a download of exactly that part in flight;
forcing it to Loaded from there leaves the s3 worker to hit the same assertion
when the download finishes, and hands eviction a part still being written into.
While here: the extra `s3_headers_increase_loaded_parts(1)` double-counted,
because `s3_header_mark_part_loading()` already counts the part on the
transition it makes.

Both callers now go through `s3_header_mark_part_loading()` and act on what it
found. The assertion also says which part and which status now, since the
process that trips it is not the one that changed the status.

**Not reproduced.** 10 local runs of the failing test are clean and the
collision never showed up under an instrumented build, so this is the two
illegal transitions the code allows, matching the crash, rather than a caught
one. The whole `s3_test` suite (17 tests) plus regress (48) and isolation (34)
pass with the change.

### DIFF
--- a/src/btree/io.c
+++ b/src/btree/io.c
@@ -2284,9 +2284,24 @@ perform_page_io_build(BTreeDescr *desc, Page img,
 			tag.checkpointNum = chkpNum;
 			tag.segNum = offset / ORIOLEDB_SEGMENT_SIZE;
 			index = (offset % ORIOLEDB_SEGMENT_SIZE) / ORIOLEDB_S3_PART_SIZE;
-			s3_header_mark_part_loading(tag, index);
-			s3_header_mark_part_loaded(tag, index);
-			s3_headers_increase_loaded_parts(1);
+
+			/*
+			 * The allocation ran into a part the file did not have yet, so
+			 * its contents are here and nothing has to be fetched -- but only
+			 * if we are the ones who claimed it.  After a crash restart the
+			 * datafile length rewinds into a part the previous run had
+			 * already uploaded and evicted, so a reader can have a download
+			 * of exactly this part in flight; forcing it to Loaded from here
+			 * leaves the s3 worker to fail Assert(status ==
+			 * S3PartStatusLoading) when that download finishes, and hands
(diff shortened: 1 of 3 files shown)


akorotkov:
CI ran the test this came from: **`test_s3_data_eviction` passes in both cells
that execute `s3_test`** — 14.5 s on one, 363 s on the valgrind one. That is
not evidence the race is gone (it is rare enough that I could not provoke it in
10 local runs either), but it does confirm the change does not disturb the path
it touches. The full 17-test `s3_test` suite is green locally as well.

The one red cell on this PR (`arm64/17/gcc valgrind_2`) is the
`recovery_heap_verdict_test` flake from `main`, which #1031 fixes.

PR #1044, opened 2026-08-15
Fix the crash-recovery livelock: register a transaction by its own joint commit, and keep it above replayStartPtr

akorotkov:
Crash recovery livelocks: the startup process spins at 100% CPU inside `o_btree_modify_handle_conflicts()` on one WAL record, and the cluster never opens. Found by a jepsen-style append workload with a SIGKILL nemesis; the wedged data directory is what the WAL below is read off.

Rebased onto `main` now that 49f9ed38 is in — that fix is what makes a transaction reliably write a joint commit at PRE_COMMIT, which this one relies on.

## How it works now, and why it breaks

A transaction that owns a heap xid writes **no OrioleDB finish record** — `XACT_EVENT_COMMIT` takes the `set_oxid_xlog_ptr()` branch. Its verdict comes from the builtin commit record, which `o_xact_redo_hook()` can only apply to an oxid that `WAL_REC_JOINT_COMMIT` put on `joint_commit_list`.

`wal_joint_commit()` is called from two places: `XACT_EVENT_PRE_COMMIT` (the transaction's own) and `oxid_subxact_callback()` at **RELEASE SAVEPOINT**. The record does not say which, so replay registers the transaction on either — including the subtransaction one, written arbitrarily long before the transaction ends.

Nothing stops a checkpoint from pinning `replayStartPtr` in the middle of that span: it is pinned with no wait, unlike `sysTreesStartPtr`, which waits on `commitInProgressXlogLocation` one screen above. So the registration lands below the pin while the transaction's later changes land above it:

```
0/1C248B28  oxid 1162773  JOINT_COMMIT(xid=311481)   <- below replayStartPtr, never replayed
0/1C249B70  replayStartPtr
0/1C249E28  oxid 1162773  UPDATE [5,57380,57391]     <- replayed, leaves an in-progress version
0/1C249EA8  builtin XACT_COMMIT xid=311481           <- the hook finds nothing on the list
0/1C24ADE0  oxid 1162811  UPDATE, same tree          <- spins forever
```

Replay applies changes whose verdict it can never learn. The oxid stays `COMMITSEQNO_INPROGRESS` for the whole replay — leftovers are only aborted in `recovery_finish()`, i.e. after replay — and the next record touching one of its rows retries without end: the conflict is real, but the writer's process is gone, so `wait_for_oxid()` returns without blocking. A backend can be cancelled; the startup process cannot.

## The fix

**Only the transaction's own joint commit registers it.** `WAL_REC_JOINT_COMMIT` now carries whether it was written for a subtransaction, and replay ignores those. The one written at PRE_COMMIT comes after everything the transaction wrote, so registration can no longer precede the changes. Reading the flag is gated on the WAL version (`ORIOLEDB_WAL_VERSION` 18 → 19), so older WAL keeps its previous meaning.

**And the registration can no longer fall below the pin.** The checkpoint now waits for the commits at or below `replayStartPtr` to finish, exactly as it already does for `sysTreesStartPtr`. `commitInProgressXlogLocation` is published by `flush_local_wal()` for that joint commit and cleared by `wal_after_commit()`, so after the wait every transaction whose registration is below the pin has finished and has nothing left to write above it.

Together: replay cannot see a transaction's changes without also seeing the record that says whose they are.

The first commit is independent hardening for the same area — a double settle that a second `dlist_push_tail()` turns into a cycle in `finished_list`, whose drain then walks an entry it already freed (`detected double pfree`, or the `pairingheap_remove()` assert under `--enable-cassert`). On the abort path the builtin record precedes ours, so both settle paths fire; pre-existing, and parallel-recovery only.

## Testing

* `test/t/recovery_livelock_test.py` builds that WAL deterministically: it hangs for 43 s without the fix and passes in 2 s with it.
* A second shape reaches commit with **no** joint commit at all — an oxid acquired inside `PreCommit_on_commit_actions()`, which runs after `XACT_EVENT_PRE_COMMIT`. `ON COMMIT DROP` of an OrioleDB temp table does write to the system trees, so it is not simply invisible to replay; a probe on the branch of `recovery_finish()` that guesses a verdict shows nothing is left unsettled there. `test_recovery_after_on_commit_drop` pins it.
* `regresscheck` 49/49, `isolationcheck` 35/35, `testgrescheck_part_1` (299 tests), `recovery_test.py` 57/57.

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -232,6 +232,7 @@ TESTGRESCHECKS_PART_2 = test/t/checkpoint_concurrent_test.py \
 						test/t/recovery_heap_verdict_test.py \
 						test/t/recovery_heap_xid_binding_test.py \
 						test/t/recovery_item_rollback_test.py \
+						test/t/recovery_livelock_test.py \
 						test/t/recovery_row_lock_test.py \
 						test/t/reindex_test.py \
 						test/t/restartpoint_ddl_window_test.py \
--- a/include/recovery/wal.h
+++ b/include/recovery/wal.h
@@ -28,7 +28,7 @@
  * ORIOLEDB_COMPRESS_VERSION (see big comment on versioning
  * in include/orioledb.h)
  */
-#define ORIOLEDB_WAL_VERSION (18)
+#define ORIOLEDB_WAL_VERSION (19)
 
 /*
  * Value has been fixed at the moment of introducing WAL versioning.
@@ -58,6 +58,16 @@
  * We should never change this value.
  */
 #define ORIOLEDB_REL_TABLESPACE_WAL_VERSION (18)
+
+/*
+ * WAL_REC_JOINT_COMMIT says whether it was written for a subtransaction.
+ * Replay ignores the subtransaction ones: they are emitted at RELEASE
+ * SAVEPOINT, long before the transaction ends, and treating them as the
+ * transaction's registration leaves it registered across a window in which it
+ * keeps writing -- and in which a checkpoint can pin replayStartPtr past the
+ * record replay needed to see.
+ */
+#define ORIOLEDB_SUBXACT_JOINT_COMMIT_WAL_VERSION (19)
 
 /* Constants for commitInProgressXlogLocation */
 #define OWalTmpCommitPos			(0)
@@ -164,6 +174,8 @@ typedef struct
 	uint8		xid[sizeof(TransactionId)];
 	uint8		xmin[sizeof(OXid)];
 	uint8		csn[sizeof(CommitSeqNo)];
+	/* Since ORIOLEDB_SUBXACT_JOINT_COMMIT_WAL_VERSION */
+	uint8		subTransaction;
 } WALRecJointCommit;
 
 typedef struct
--- a/include/recovery/wal_reader.h
+++ b/include/recovery/wal_reader.h
@@ -52,6 +52,7 @@ typedef struct WalRecord
 			TransactionId xid;
 			OXid		xmin;
 			CommitSeqNo csn;
+			bool		subTransaction;
 		}			joint_commit;
 		struct
 		{
--- a/src/btree/modify.c
+++ b/src/btree/modify.c
@@ -37,6 +37,14 @@
 /*
  * Context for o_btree_modify_internal()
  */
+/*
+ * How many fruitless conflict retries before the modify loop reports itself.
+ * High enough that ordinary contention never trips it -- a real conflict is
+ * resolved within a handful of retries -- and low enough that a loop which
+ * will never converge says so within milliseconds.
+ */
+#define MODIFY_CONFLICT_RETRY_REPORT 10000
+
 typedef struct
 {
 	OBTreeFindPageContext *pageFindContext;
@@ -63,6 +71,7 @@ typedef struct
 	BTreeKeyType keyType;
 	UndoLocation savepointUndoLocation;
 	BTreeModifyCallbackInfo *callbackInfo;
+	int			conflictRetries;
 } BTreeModifyInternalContext;
 
 typedef enum ConflictResolution
@@ -145,6 +154,7 @@ o_btree_modify_internal(OBTreeFindPageContext *pageFindContext,
 	context.savepointUndoLocation = get_subxact_undo_location(desc->undoType);
 	context.pageReserveKind = pageReserveKind;
 	context.callbackInfo = callbackInfo;
+	context.conflictRetries = 0;
 
 	Assert(callbackInfo);
 	Assert((action != BTreeOperationInsert) || (tupleType == BTreeKeyLeafTuple));
@@ -632,6 +642,27 @@ o_btree_modify_handle_conflicts(BTreeModifyInternalContext *context)
 				{
 					Assert(cbAction == OBTreeCallbackActionXidNoWait);
 				}
+
+				/*
+				 * This retry can fail to converge.  If the conflicting
+				 * transaction has no owning process -- which is what every
+				 * transaction that was in flight at crash time looks like --
+				 * wait_for_tuple() above returns without blocking, the
+				 * conflict is still there when we look again, and we come
+				 * straight back here.  A backend can at least be cancelled;
+				 * the startup process cannot, so replay spins on one record
+				 * at 100% CPU and the cluster never opens.
+				 *
+				 * Make that state observable: the loop is otherwise
+				 * completely silent.  The page lock is already dropped here,
+				 * so waiting is safe.
+				 */
+				if (++context->conflictRetries == MODIFY_CONFLICT_RETRY_REPORT)
+					elog(LOG, "orioledb: modify conflict retry is not converging: "
+						 "oxid " UINT64_FORMAT ", conflicting oxid " UINT64_FORMAT
+						 ", %d retries, recovery %d",
+						 (uint64) context->opOxid, (uint64) oxid,
+						 context->conflictRetries, is_recovery_process() ? 1 : 0);
 
 				result = refind_page(pageFindContext,
 									 context->key,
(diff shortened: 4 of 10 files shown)


akorotkov:
Note on the overlap with #1045, since it changes what commit 2 is worth.

#1045 fixes the subtransaction logical-xid handoff, which is what made `wal_joint_commit()` skip the binding in the first place. With that in, the logical xid survives `RELEASE SAVEPOINT`, the joint commit record fires again and carries the (oxid, heap xid) pair — and I could not construct any SQL shape that still reaches the bare-binding path. I verified this on a scratch merge of both branches: `recovery_heap_xid_binding_test.py` passes there even with commit 2's call removed.

So, concretely:

* **applied alone, commit 2 fixes real data loss** — the test fails with `0 != 555` without it;
* **applied after #1045, it is a guard rather than a live fix**, and its test then only asserts the property (a committed transaction's OrioleDB changes survive crash recovery) rather than exercising the new path. I have said so in the test's docstring rather than leave a test that quietly proves nothing.

I would keep it: it enforces the invariant replay actually depends on — every oxid whose changes replay applies must be settleable — directly, instead of by way of the logical xid lifecycle, which is a different subsystem's concern. But it is a reasonable thing to want dropped if you would rather #1045 carry it alone, and the two commits here are independent.

Commit 1 is unaffected by any of this: the livelock it fixes needs only a checkpoint pinning `replayStartPtr` past a subtransaction's joint commit.

akorotkov:
Converting to draft: **commit 1 has a regression** and must not be merged as it stands.

CI is red across 16/17/18 — `test/t/recovery_test.py::test_recovery_partiton` — and it reproduces locally on every run: parallel crash recovery stalls with the startup process and two recovery workers spinning, and in CI one cell caught the same state as a segfault:

```
#0  table_descr_inc_refcnt (descr=0x0) at src/tableam/descr.c:2454
#1  recovery_queue_process (queue=..., id=2) at src/recovery/worker.c:440
#2  recovery_worker_main
```

i.e. a worker was told to apply changes for a relation whose descriptor it cannot fetch.

What I have established so far:

* it is the `joint_commit_list` registration in commit 1, not the `break` removal — restoring the single-match `break` still hangs;
* it is parallel-recovery-specific: the preserved production data directory replays fine single-process, and the deterministic livelock test passes;
* the likely mechanism is that the hook now settles many more oxids, and settling at the builtin commit record makes the leader do `workers_send_oxid_finish()` / `workers_synchronize()` at points in the stream where it previously never did, which perturbs the leader-to-worker ordering that the workers' descriptor fetches depend on.

The diagnosis of the livelock itself (the WAL evidence at the top of this PR) stands; it is the *remedy* that is wrong — settling out of band relative to the worker dispatch stream is more intrusive than I took it for. I am reworking it toward something local: leave the dispatch stream alone and resolve the stuck oxid where the conflict is actually detected, from the heap xid the recovery xid state already carries.

Commit 2 (`wal_bind_heap_xid`) is independent of this and unaffected; #1045 is a separate branch and is green.

akorotkov:
Follow-up, and this one changes what the fix has to look like.

I bisected my own change by hunk. Only the `joint_commit_list` registration in commit 1 causes the regression — with just that hunk disabled, `recovery_test.py` passes 57/57 twice over, and the other test that failed once (`test_recovery_sk_modify_pending_replica_concurrent`) is flaky on `main` terms, not mine. The rest of commit 1 (unlinking on settle, skipping an already settled oxid, the non-converging LOG) is clean.

But dropping the registration then breaks commit 2's test, and the reason is the useful part. With a probe on the abort path in `recovery_finish()`:

```
LOG:  ABORTPROBE: worker 1 aborts oxid 52 xid 0 at recovery_finish
LOG:  ABORTPROBE: worker 2 aborts oxid 52 xid 0 at recovery_finish
```

**The workers never learn the heap xid.** `state->xid` is only ever set on the leader, from `rec->heapXid` while parsing WAL; a worker's state is created by `recovery_switch_to_oxid(oxid, id)` from a message that carries no xid. So at end of replay every worker sees an in-progress oxid with `xid == 0`, decides it cannot have committed, and applies undo — rolling back a committed transaction's changes, in parallel recovery, whatever the leader knows.

That reframes both problems as one missing piece: **the (oxid, heap xid) binding has to reach the workers**, not just the leader. Settling at the builtin commit record was my attempt to paper over that by broadcasting the verdict, and it is too intrusive — it makes the leader do `workers_send_oxid_finish()` / `workers_synchronize()` at points in the stream where the leader-to-worker ordering does not expect it, and workers wedge inside `recovery_queue_process()` (one CI cell caught it as a NULL descriptor there).

The direction I would take next is to carry the heap xid in the leader-to-worker protocol — either on `RECOVERY_MODIFY_OXID`, or as a small explicit bind message — so each worker's recovery xid state can answer the same question the leader's can. That adds information without changing *when* anything happens, which is what went wrong here.

Leaving this in draft. The diagnosis at the top of the PR (the WAL evidence for the livelock) and #1045 are unaffected.

akorotkov:
Tried the direction I proposed above — carry the binding to the workers — and it needs to be narrower than I first wrote. Recording what each variant does so the next attempt starts from measurements rather than from my guesses.

**Variant: broadcast a bind message to all workers when the leader learns the (oxid, heap xid) pair.** Added `RecoveryMsgTypeBindHeapXid`; the worker calls `recovery_switch_to_oxid()` and stores the xid.

* `recovery_heap_xid_binding_test` passes, so the mechanism does fix the data loss;
* but `recovery_test.py` breaks in a spread of tests (`test_missingattr_recovery`, `test_recovery_deep_sql_function_*`, ...).

Two distinct reasons, one fixed and one not:

1. Every worker then reaches end-of-replay with a valid xid, and the heap-verdict path in `recovery_finish()` has no `worker_id` guard — so *each* worker allocated a csn and published it for the same oxid, which is the multiple-settle hazard the comment there already warns about. Guarding the publication with `worker_id < 0` (workers still run the undo bookkeeping, they just do not publish) is correct on its own and I would keep it.
2. That alone does not fix the spread, because `recovery_switch_to_oxid()` in a worker *creates recovery xid state* — with retain-undo locations and an xmin entry — for transactions that worker never participates in. Broadcasting therefore perturbs every worker's retain/xmin bookkeeping, which is the same class of mistake as settling out of band, just in a different place.

So the constraint is sharper than "get the xid to the workers": **only workers that actually apply a change of the transaction may learn about it, and only through the path that already gives them state for it.** Concretely that means carrying the xid on `RECOVERY_MODIFY_OXID` (the message that already makes a worker adopt an oxid), plus a top-up to `used_by[]` workers for the case where the heap xid is acquired after the modifies were dispatched. That is a change to the hot-path wire format, so I am not landing it unverified at the end of a session.

Still open, unchanged: the livelock itself. Control run confirms it is real and not mine — on a build carrying only the binding-record commit (no registration), the jepsen-style workload wedges the cluster inside one 10-minute run; artifact kept at `/data/jrun/WEDGE2_200123`.

akorotkov:
The missing piece is now its own PR: #1046 carries the (oxid, heap xid) binding to the workers that actually apply the transaction's changes, and confines publishing the verdict to the leader. That fixes the parallel-recovery data loss on its own — a committed transaction's OrioleDB half no longer gets undone — and it is the piece this PR was really missing.

Commit 2 here (`wal_bind_heap_xid`) moved to #1046 as well, since without the worker side it does not help in parallel recovery.

What is left in this PR is the livelock, and it is still unfixed: I would rather leave it in draft with a correct diagnosis than merge a remedy I have shown breaks parallel recovery. Whoever picks it up now starts from a much better position — with #1046 the workers know the heap xid, which is the prerequisite for resolving a stuck conflict locally instead of settling out of band.

The rest of commit 1 is worth keeping in some form and is independent of the registration: `recovery_finish_current_oxid()` unlinking on settle, the already-settled skip in the `WAL_REC_COMMIT`/`WAL_REC_ROLLBACK` path (that one fixes a real pre-existing double-settle: the builtin abort record precedes ours, and settling twice closes `finished_list` into a cycle), and the non-converging-retry LOG in `o_btree_modify_handle_conflicts()`. Say the word and I will split those out.

PR #1046, opened 2026-08-15
Stop crash recovery from rolling back half of a committed transaction

akorotkov:
In **parallel** crash recovery a committed transaction could come back with its OrioleDB changes rolled back and its heap changes intact. Two things are missing, one on the wire and one between the leader and its workers.

A transaction that owns a heap xid writes no OrioleDB finish record — `XACT_EVENT_COMMIT` takes the `set_oxid_xlog_ptr()` branch — so `recovery_finish()` decides its fate from the heap xid.

## 1. The workers never learn the heap xid

`state->xid` is only ever set on the leader, from `rec->heapXid` while parsing WAL. A worker's recovery xid state is created by `recovery_switch_to_oxid()` from a message that carries no xid, so at the end of replay every worker sees an in-progress oxid with no xid, concludes it cannot have committed, and applies undo. Measured with a probe on that path:

```
ABORTPROBE: worker 1 aborts oxid 52 xid 0 at recovery_finish
ABORTPROBE: worker 2 aborts oxid 52 xid 0 at recovery_finish
```

The binding now travels on the message that makes a worker adopt the oxid in the first place (`RECOVERY_MODIFY_HEAP_XID`, next to the oxid), so only workers that actually apply one of the transaction's changes hear about it. When the heap xid is acquired after some changes were already dispatched, exactly the workers in `used_by[]` get an explicit bind message.

Publishing the verdict stays the leader's alone: workers now reach the heap-verdict path in `recovery_finish()` too, and every process settling the same oxid would settle it repeatedly, which spins `set_oxid_xlog_ptr()` forever — the hazard the comment there already describes. A worker runs the undo bookkeeping and stops.

Two alternatives were tried and **measured to break parallel recovery**, so they are called out in the commit message rather than left for someone to rediscover:

* settling the oxid at the builtin commit record — it makes the leader do `workers_send_oxid_finish()` / `workers_synchronize()` at points the leader-to-worker ordering does not expect, and workers wedge in `recovery_queue_process()` (see #1044);
* broadcasting the binding to all workers — `recovery_switch_to_oxid()` in a worker *creates* state, with retain-undo and xmin entries, for oxids it never touches.

## 2. Sometimes no record carries the heap xid at all

Both records that normally carry it can be missing at once:

* `add_xid_wal_record()` samples `GetTopTransactionIdIfAny()` when a container's first record is buffered, so a transaction that acquires its heap xid **after** its last OrioleDB change has every `WAL_REC_XID` carrying `InvalidTransactionId`;
* `wal_joint_commit()` is gated on a logical xid, which `RELEASE SAVEPOINT` hands back to a parent that has none if the transaction's first OrioleDB write happened inside the subtransaction (that gate is #1045).

```
XID(oxid=53 lxid=192 heapXid=0)  UPDATE o_t
XID(oxid=53 lxid=0   heapXid=0)  UPDATE o_t
XACT_COMMIT xl_xid=755

value before crash: 555
value after crash recovery: 0      <- OrioleDB half rolled back
heap row after crash recovery: 1   <- heap half survived
```

A bare `WAL_REC_XID` is emitted at `PRE_COMMIT` in that case, after flushing the buffered changes so the heap xid is sampled once it exists. One shape stays out of reach and is documented: `XACT_EVENT_PRE_COMMIT` runs before `PreCommit_on_commit_actions()`.

## Testing

* `test/t/recovery_heap_xid_binding_test.py` — fails with `0 != 555` without either half of this PR.
* `recovery_test.py` 57/57, `regresscheck` 49/49, `isolationcheck` 35/35, `testgrescheck_part_1` (302 tests) all green.

Related: #1045 (the logical-xid gate above, independent and green) and #1044 (the crash-recovery livelock, still draft — its diagnosis holds, its remedy does not, and this PR is the piece it was missing).

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -229,7 +229,9 @@ TESTGRESCHECKS_PART_2 = test/t/checkpoint_concurrent_test.py \
 						test/t/pg_dump_restore_test.py \
 						test/t/parallel_test.py \
 						test/t/recovery_heap_verdict_test.py \
+						test/t/recovery_heap_xid_binding_test.py \
 						test/t/recovery_item_rollback_test.py \
+						test/t/recovery_livelock_test.py \
 						test/t/recovery_row_lock_test.py \
 						test/t/reindex_test.py \
 						test/t/restartpoint_ddl_window_test.py \
--- a/Makefile
+++ b/Makefile
@@ -231,7 +231,6 @@ TESTGRESCHECKS_PART_2 = test/t/checkpoint_concurrent_test.py \
 						test/t/recovery_heap_verdict_test.py \
 						test/t/recovery_heap_xid_binding_test.py \
 						test/t/recovery_item_rollback_test.py \
-						test/t/recovery_livelock_test.py \
 						test/t/recovery_row_lock_test.py \
 						test/t/reindex_test.py \
 						test/t/restartpoint_ddl_window_test.py \
--- a/include/recovery/internal.h
+++ b/include/recovery/internal.h
@@ -61,11 +61,14 @@ typedef enum
 	RecoveryMsgTypeLeaderParallelIndexBuild,
 	RecoveryMsgTypeWorkerParallelIndexBuild,
 	RecoveryMsgTypeInit,
-	RecoveryMsgTypeReinsert
+	RecoveryMsgTypeReinsert,
+	RecoveryMsgTypeBindHeapXid
 } RecoveryMsgType;
 
 #define RECOVERY_MODIFY_OXID (0x0100)
 #define RECOVERY_MODIFY_OIDS (0x0200)
+/* the oxid above rides on a heap xid, which follows it */
+#define RECOVERY_MODIFY_HEAP_XID (0x0400)
 
 #define RECOVERY_QUEUE_BUF_SIZE (8 * 1024)
 
@@ -88,6 +91,20 @@ typedef struct
 	RecoveryMsgHeader header;
 	XLogRecPtr	ptr;
 } RecoveryMsgPtr;
+
+/*
+ * Tells a worker which heap xid an oxid it already works on rides on.
+ *
+ * Sent only when the leader learns the binding after that worker has already
+ * been given modifies for the oxid; when it is known beforehand the binding
+ * travels on the modify message itself (RECOVERY_MODIFY_HEAP_XID).
+ */
+typedef struct
+{
+	RecoveryMsgHeader header;
+	OXid		oxid;
+	TransactionId xid;
+} RecoveryMsgBindXid;
 
 typedef struct
 {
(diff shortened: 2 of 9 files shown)


akorotkov:
Converting to draft. Answering the review question — does this fix anything on its own — with measurements rather than argument, and the answer is essentially no.

`test/t/recovery_heap_xid_binding_test.py` run against three builds:

| build | result |
|---|---|
| `main` | **FAILS** — `0 != 555`, the committed transaction's OrioleDB half is rolled back |
| `main` + #1045 | passes |
| `main` + this PR | passes |

So the data loss is real on `main` today, but **#1045 already fixes it**: keeping the transaction's logical xid restores `wal_joint_commit()` at PRE_COMMIT, which is what puts the (oxid, heap xid) binding on the wire. Everything downstream then works through the existing path — and, as you point out, `o_xact_redo_hook()` already calls `workers_send_oxid_finish()`, so the workers do get told the verdict for every transaction the hook can settle.

Which means my framing in the PR body was wrong. "The workers never learn the heap xid" is only true for transactions the hook *cannot* settle — those with no joint commit record at all — and with #1045 in place I cannot construct such a transaction any more. On top of #1045 the extra message here is dead code.

What is left of the two commits:

* the wire change and the bind message: no independent value once #1045 lands; they only matter for a transaction that reaches the end of replay unsettled, which is the livelock case in #1044;
* `wal_bind_heap_xid()`: same — belt and braces for a logical xid being invalid at PRE_COMMIT for some reason other than the one #1045 fixes, and I could not produce one;
* the `worker_id < 0` guard on the heap-verdict csn publication in `recovery_finish()`: that one is correct on its own terms — a worker reaching that path would settle an oxid the leader also settles — but it is currently unreachable, so it is a hardening, not a fix.

I would close this and keep #1045 as the fix, unless you want the wire change kept as groundwork for #1044 — the livelock does need a worker to be able to settle a transaction the hook never sees. Say which and I will act.

The test itself is worth keeping somewhere regardless: it fails on `main` and is the only regression test for that data loss. It could move to #1045.

akorotkov:
Closing, per the measurement above: on top of #1045 this adds nothing observable.

The valuable parts have moved:

* `test/t/recovery_heap_xid_binding_test.py` — the regression test for the committed-transaction data loss, which fails on `main` — is now in #1045, next to the fix that actually addresses it;
* the double-settle hardening and the non-converging-retry report are in #1044, which also keeps `recovery_livelock_test.py` as a skipped reproducer.

Not carried over, deliberately: the `RECOVERY_MODIFY_HEAP_XID` wire field, the bind message and `wal_bind_heap_xid()`. All three only matter for a transaction that reaches the end of replay unsettled, which is the livelock in #1044 — and, as pointed out in review, `o_xact_redo_hook()` already tells the workers the verdict for every transaction it can settle. If the livelock fix ends up needing a worker to settle a transaction the hook never sees, this branch is the starting point; the reasoning and the measurements are in the comments here.

Issue #982, opened 2026-07-24
Repeatable Read Isolation Anomalies, Single Node


pmbauer:
I'm observing the following anomalies under [Postgres Repeatable Read](https://www.postgresql.org/docs/current/transaction-iso.html) (snapshot isolation).  The same workloads run against stock Postgres do not exhibit these anomalies.

- read-own-writes
    - not an RR anomaly per se, but intra-transaction inconsistency
- phantom reads
    - less common than read-own-writes failures, but observed intermittently 
- uniqueness violations
    - so far, only under Antithesis.
    - this may warrant its own ticket, but further investigation needed

These observations are under the following conditions with the caveat these conditions may not be exclusive.  I've yet to run a full matrix varying postgres versions, OrioleDB versions, and lesser isolation levels (e.g. read committed).

- Repeatable Read Isolation
- x86_64, under MacOS/Rosetta emulation
- x86_64, under Antithesis' hypervisor
- OrioleDB latest main
- Jepsen "append" workload (see Appendix for details)

## Reproduction

In the **Anomaly Examples** section I reference some executions on Antithesis' hypervisor.  But, I've managed to trigger the first two anomaly categories locally.

### Local

Local reproduction only requires a Docker VM that can execute x86_64 binaries, native or Rosetta on MacOS.

```bash
git clone git@orioledb/orioledb.git
cd orioledb
git checkout pmbauer/add-initial-antithesis-workload
```

Execute one of the following, depending on your host.  
The initial build will take 10+ minutes depending on your machine and WAN bandwidth.  
By default, it builds the latest OrioleDB on PG17 and a Jepsen client workload.  
For details on customizing the OrioleDB, PG major version, or [Jepsen parameters](https://github.com/orioledb/orioledb/blob/d04582306c49e82c8675027e38b62ca0b39fb7b9/test/antithesis/config/workload/jepsen-repeatable-read/compose.yaml#L12-L16), see [the workload README](https://github.com/orioledb/orioledb/tree/pmbauer/add-initial-antithesis-workload/test/antithesis).  
You can customize any parameter by overriding the env var values.

```bash
# If you are on a Linux/x86_64 host, run this
JEPSEN_TIME_LIMIT=300 make up

# If you are on MacOS with Rosetta
# This runs an x86_64 OrioleDB
JEPSEN_TIME_LIMIT=300 JEPSEN_PLATFORM=linux/arm64 make up
```

You can wait 5 minutes or tail the `jepsen-client` logs to see results

[...]

PR #1047, opened 2026-08-16
Resolve the lock-only chain before the seq scan asks whose version it is (ORI-229)

akorotkov:
A sequential scan could stop seeing the scanning transaction's own write, while an index scan on the same row still found it — the shape reported as ORI-229 / #982.

## The defect

`btree_seq_scan_getnext_internal()` merges the live leaf image with a historical one. Where both hold the same key it decided which to emit from the **raw** leaf header:

```c
if (XACT_INFO_OXID_IS_CURRENT(tuphdr->xactInfo))
```

A concurrent transaction may hold a `FOR KEY SHARE` lock on top of our own version — `KEY SHARE` and `NO KEY UPDATE` do not conflict — and then that header describes the locker, not us. The test answers "not mine", the scan takes the historical branch, and the transaction is handed a version of its own row from before it wrote.

The iterator makes the same decision in `fetch_our_tuple_from_page()` ("the live version overrides whatever we could have from the undo log page image") and resolves the chain first, which **74f2acc2** taught it to do for exactly this reason:

> Look for non-lock-only undo record for that tuple, as there might be FOR KEY SHARE locks placed by concurrent transactions.

`scan.c`'s copy of that decision has not been touched since af569861, the initial public release, so that fix never reached it.

Every other site asking the same question is already careful:

| site | header resolved? |
|---|---|
| `iterator.c:594`, `fetch_our_tuple_from_page()` | yes — `find_non_lock_only_undo_record()` plus `!XACT_INFO_IS_LOCK_ONLY` (74f2acc2) |
| `iterator.c:785`, version walk | yes — resolved on entry at `iterator.c:722` |
| `modify.c:1206`, `page_unique_check()` | yes — resolved one line above |
| `operations.c` (7 sites), `indexam/handler.c:374` | n/a — modify callbacks, handed an already resolved `xactInfo` |
| **`scan.c:2196`** | **no — fixed here** |

The extra work only happens where both images hold the same key, so a scan with no historical image — the ordinary case — pays nothing.

## The test

`test/t/seq_scan_own_write_test.py` builds the state deterministically, and the ordering is the fiddly part:

1. a live sequential scan fixes an RR snapshot — a page image is only kept if someone needs it;
2. another session **splits the page** and commits — a page-level image is only produced by a structural change, not by a row update;
3. we write our row — **after** the split, so the image predates the write and emitting from it actually loses it;
4. a third session takes `FOR KEY SHARE` on that row — the lock-only record on top;
5. the scan has to still see our write.

Without the fix it reads `(1, 0)` where the transaction wrote `42`. With the split *before* our write the wrong branch is still taken but the result happens to be right, which is what made two earlier constructions look clean.

Instrumented at the decision point (probe kept on branch `debug/ori229-probe`, not for merging):

```
ORI229PROBE: cmp==0 lockOnly=1 raw=0 resolved=1
```

— the row is ours, and only the raw header says otherwise.

## What this does and does not explain

Only the scanning transaction's own writes are affected: for other transactions' versions, taking the historical branch is correct, and cross-transaction visibility goes through `o_find_tuple_version()` untouched. So this can account for read-own-writes misses (ORI-229), for lost updates inside a read-modify-write statement such as `UPDATE ... SET val = CONCAT(val, ',', e)` over an unindexed column — the recomputation starts from the stale value and the transaction's own earlier element disappears — and for spurious upsert conflicts. It does not touch G1c/G2 (ORI-242).

One caveat I have not closed: the reproducer takes an explicit `FOR KEY SHARE`, while the jepsen append workload takes no row locks and has no foreign keys, and `RowLockKeyShare` only arises from `LockTupleKeyShare` in `tuple_lock_mode_to_row_lock_mode()`. So this being *the* cause of ORI-229 as that workload reproduces it is not established — it may be one of several holes. The decisive check is to run `ci/rr_append_hunt.py` on a build with this fix and see whether read-own-writes still turns up.

## Testing

`regresscheck` 49/49, `isolationcheck` 35/35, `testgrescheck_part_1` (301 tests) — all green with the fix; the new test fails without it.

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -235,6 +235,7 @@ TESTGRESCHECKS_PART_2 = test/t/checkpoint_concurrent_test.py \
 						test/t/restartpoint_ddl_window_test.py \
 						test/t/s3_test.py \
 						test/t/schema_test.py \
+						test/t/seq_scan_own_write_test.py \
 						test/t/seq_scan_resowner_test.py \
 						test/t/seq_scan_undo_test.py \
 						test/t/temp_local_pool_test.py \
--- a/src/btree/scan.c
+++ b/src/btree/scan.c
@@ -2193,7 +2193,29 @@ btree_seq_scan_getnext_internal(BTreeSeqScan *scan, MemoryContext mctx,
 
 				if (cmp == 0)
 				{
-					if (XACT_INFO_OXID_IS_CURRENT(tuphdr->xactInfo))
+					BTreeLeafTuphdr nonLockTuphdr = *tuphdr;
+
+					/*
+					 * The live version overrides whatever the historical page
+					 * image holds for this key -- but only if it is ours, and
+					 * the header on the page does not answer that on its own.
+					 * A concurrent transaction may have placed a FOR KEY
+					 * SHARE lock on top of our own version, which the two
+					 * modes allow, and then the header describes the locker.
+					 * Deciding on it emits the historical version instead,
+					 * and the transaction stops seeing its own write.
+					 *
+					 * Resolve the chain first, exactly as
+					 * fetch_our_tuple_from_page() does for the iterator. This
+					 * only runs where both images hold the same key, so it
+					 * costs nothing on a scan with no historical image at
+					 * all.
+					 */
+					(void) find_non_lock_only_undo_record(scan->desc->undoType,
+														  &nonLockTuphdr);
+
+					if (!XACT_INFO_IS_LOCK_ONLY(nonLockTuphdr.xactInfo) &&
+						XACT_INFO_OXID_IS_CURRENT(nonLockTuphdr.xactInfo))
 					{
 						BTREE_PAGE_LOCATOR_NEXT(scan->histImg, &scan->histLoc);
 						break;
--- a/test/t/seq_scan_own_write_test.py
+++ b/test/t/seq_scan_own_write_test.py
@@ -1,8 +1,6 @@
 #!/usr/bin/env python3
 # coding: utf-8
 
-import unittest
-
 from .base_test import BaseTest
 
 FAT = "x" * 180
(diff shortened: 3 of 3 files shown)


akorotkov:
> Разве у iterator не то же самое?

Было — и ты это уже починил. `git log -S find_non_lock_only_undo_record -- src/btree/iterator.c` даёт **74f2acc2** (30 Jan 2026, «Fixes for `o_btree_find_tuple_by_key_cb()` with combined result»), и в его сообщении ровно этот случай:

> * Look for non-lock-only undo record for that tuple, as there might be FOR KEY SHARE locks placed by concurrent transactions.

Тот же вопрос в `scan.c` не трогали с af569861, то есть с первого публичного релиза — фикс 74f2acc2 просто не доехал до последовательного скана.

Прошёлся по всем местам, где спрашивается `XACT_INFO_OXID_IS_CURRENT`:

| место | заголовок разрешается? |
|---|---|
| `iterator.c:594`, `fetch_our_tuple_from_page()` | да — `find_non_lock_only_undo_record()` + проверка `!XACT_INFO_IS_LOCK_ONLY` (74f2acc2) |
| `iterator.c:785`, обход версий | да — `tupHdr` уже разрешён на входе (iterator.c:722), плюс `!LOCK_ONLY` в самом условии |
| `modify.c:1206`, `page_unique_check()` | да — `find_non_lock_only_undo_record()` строкой выше |
| `operations.c` (7 мест), `indexam/handler.c:374` | не применимо: это modify-колбэки, им `xactInfo` приходит уже разрешённым из `o_btree_modify_handle_conflicts()` |
| **`scan.c:2196`, слияние live/hist** | **нет — сырой заголовок страницы** |

Так что единственное оставшееся место — последовательный скан, и решение там ровно то же по смыслу, что в `fetch_our_tuple_from_page()`: «живая версия перекрывает то, что мы могли бы взять из образа undo-страницы». Добавил это в описание PR.

Зонд, которым это ловилось, лежит отдельной веткой `debug/ori229-probe` (не для мержа).

PR #1051, opened 2026-08-17
Make the parent-locator invariant an enforced one

### LINKED ISSUE #1050: Assert(loc.chunk inside parentImg) in find_right_page during a serial index scan
Found by the SQLSmith job (`sqlsmith (18, check_page)`), on a branch that touches neither `src/btree/find.c` nor `src/btree/iterator.c`, so this looks pre-existing.

```
TRAP: failed Assert("loc.chunk == NULL || ((Pointer) loc.chunk >= context->parentImg && (Pointer) loc.chunk < context->parentImg + ORIOLEDB_BLCKSZ)")
      File: src/btree/find.c, Line: 1574
client backend was terminated by signal 6: Aborted
```

Backtrace, trimmed to the interesting frames:

```
#6  find_right_page (context=..., hikey=...) at src/btree/find.c:1574
#7  btree_iterator_check_load_next_page at src/btree/iterator.c:1905
#8  o_btree_iterator_fetch_internal at src/btree/iterator.c:1663
#9  o_btree_iterator_fetch at src/btree/iterator.c:1302
#10 o_iterate_index at src/tableam/index_scan.c:812
#11 o_index_scan_getnext (scan_primary=true) at src/tableam/index_scan.c:1001
#12 o_exec_fetch at src/tableam/index_scan.c:1231
#13 o_exec_custom_scan at src/tableam/scan.c:1069
#15 ExecAppend at nodeAppend.c:366
```

A plain serial index scan over a partitioned table; no parallel worker in the stack.

The locator's chunk pointer is outside the parent page image, which puts this in the same family as the FETCH-mode partial-image problems: a locator addressing a region of the image that no `partial_load_chunk()` filled. `check_page` validates page structure on every access, which is presumably why this cell caught it and the plain assert cells did not.

Reproduce with: `SQLSMITH_SEED=1786925906 bash ci/sqlsmith.sh`, though the window looks probabilistic — the same commit passed the other three matrix cells.

Job: https://github.com/orioledb/orioledb/actions/runs/31981541303/job/95249073464

akorotkov:
Fixes #1050, and this time at the producer rather than by guarding the consumers.

`find_right_page()` and `find_left_page()` navigate the immediate parent's locator to reach the sibling downlink, and both assert that its chunk pointer lives in `context->parentImg`. SQLSmith tripped that assertion in `find_right_page()` during an index scan.

**The invariant was never enforced anywhere.** `ASSERT_PARENT_LOCATOR_LOCAL` exists for exactly this — its comment says the check belongs *"at every place we record or advance such a locator, so a stray shared-page/img pointer is caught at its producer rather than only when the sibling step consumes it"* — but the macro had no users.

So assert it where the context is handed over, at `find_page()`'s success return, with the two exemptions the design actually allows:

- the copy into `parentImg` can be deliberately deferred, which `find_right_page()` completes itself;
- a MODIFY descent keeps parents in `context->img` on purpose, which is safe because both sibling steps assert `!MODIFY`.

**That assertion fires in `indices_build`, and names its producer:** `btree_find_context_from_modify_to_read()`. It clears MODIFY and re-reads only the target page, so the parent locator inherited from the MODIFY descent still points into `context->img` — which the re-read has just overwritten — while `parentImg` was never filled at all. Its caller, `slowpath_unique_check()`, then steps right through precisely that locator. The fix is to re-descend with `find_page()` when the context keeps a parent; it rebuilds every level including `parentImg`, and the failure path of that same function already did exactly this.

`regresscheck` 49/49 and `isolationcheck` 35/35 with the assertion in place.

**Correcting an earlier claim of mine in this PR.** The first version guarded the two consumers and reported the producer as unfixed, based on a probe placed at the item-recording site inside the descent loop. That probe was simply too early: `refresh_parent_img_chunk()` runs later in the same iteration, so what it saw was an intermediate state, not what `find_page()` returns. The real violation is the one above, and it is now fixed rather than worked around.

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -145,6 +145,7 @@ REGRESSCHECKS = btree_sys_check \
 				trigger \
 				truncate \
 				types \
+				unique_slowpath \
 				database_template
 ISOLATIONCHECKS = bitmap_hist_scan \
 				  bitmap_for_update_epq \
--- a/src/btree/find.c
+++ b/src/btree/find.c
@@ -75,6 +75,28 @@ static bool convert_fastpath_parent_to_img(OBTreeFindPageContext *context,
 		   ((Pointer) (loc).chunk >= (context)->parentImg && \
 			(Pointer) (loc).chunk < (context)->parentImg + ORIOLEDB_BLCKSZ))
 
+#ifdef USE_ASSERT_CHECKING
+/*
+ * Is a parent locator safe to navigate, i.e. does its chunk pointer live in
+ * context->parentImg?
+ *
+ * The descent binds it to whatever image the parent was read from, and the
+ * branches that read the parent through the locked shared page rebind it only
+ * for the cases they know about.  What they miss would reach find_right_page()
+ * and find_left_page() bound to a shared page the descent has since unlocked,
+ * which may have been evicted and reused by then.  Only assertions ask this,
+ * so it does not exist without them.
+ */
+static inline bool
+parent_locator_is_local(OBTreeFindPageContext *context,
+						BTreePageItemLocator loc)
+{
+	return loc.chunk == NULL ||
+		((Pointer) loc.chunk >= context->parentImg &&
+		 (Pointer) loc.chunk < context->parentImg + ORIOLEDB_BLCKSZ);
+}
+#endif
+
 /*
  * Initialize B-tree page find context.
  */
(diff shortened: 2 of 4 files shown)


akorotkov:
## Correction: there is no producer bug, and the deterministic driver is what shows it

I claimed above that `btree_find_context_from_modify_to_read()` leaves a
non-navigable parent locator and that `slowpath_unique_check()` steps right
through it. **That was wrong, and I have now measured it.**

`slowpath_unique_check()` is reachable deterministically, with no concurrency:
the unique check descends to the key's *lower* bound, so when a value's index
entry begins a leaf page the descent lands on the page that **ends** at it and
must step right. Re-inserting every value of a unique secondary index whose
entries are wide enough to span many pages hits that path once per page.

Instrumented at the `find_right_page()` call inside `slowpath_unique_check()`,
1000 rows with a 400-byte unique key give:

```
slowpath entries: 62   step-right calls: 62
62 USLOWSTEP index=2 chunk_null=0 in_parentimg=1 in_img=0 valid=1 deferred=0
```

The parent locator is bound inside `parentImg` and valid, 62 times out of 62.
The reason is in `find_page()`:

```c
if ((imageFlag || keepParentFlag) && level == targetLevel + 1)
    refresh_parent_img_chunk(&intCxt);
```

`imageFlag`, not just `keepParentFlag` — and `o_btree_insert_unique()` builds its
context with `BTREE_PAGE_FIND_IMAGE`. So the parent of the target is copied into
`parentImg` and the locator rebound, whatever `KEEP_PARENT` says. My "fix" was
gated on `KEEP_PARENT`, which that caller never sets, so it was dead code
carrying a comment that described a bug that isn't there.

What the branch does now:

- drops the dead re-descent and asserts `!KEEP_PARENT` in
  `btree_find_context_from_modify_to_read()`, with a comment saying why the
  locator is navigable rather than claiming it is not;
- **tightens** the `find_page()` exit assertion: the exemption is now
  `MODIFY && !IMAGE` instead of `MODIFY`, so the unique-check context is
  checked rather than waved through;
- adds `unique_slowpath`, the first regression coverage of
  `slowpath_unique_check()` at all.

The assertions themselves stay, and they remain unproven in one direction: they
have not fired in regress, isolation, 4 × 2100 s of SQLSmith, or this targeted
driver. So #1050's original report is still unexplained by anything on this
branch — this PR makes the invariant checkable, not fixed.

PR #1063, opened 2026-08-20
Let OrioleDB run ordered secondary index scans itself

akorotkov:
Addresses the secondary-index part of #450.

## What the benchmarks say now

Re-ran the microbenchmarks from #450 on current `main` (PG18, release build, real sysbench from `mdcallag/sysbench` with his lua and his test args, 10M-row sbtest1, heap and OrioleDB in one cluster). The two catastrophic cases in the issue are **gone** — `range-covered-si` 0.00 → 0.74, `range-notcovered-si` 0.01 → 0.57 — but a general per-row cost in range scans remains, and one part of it has a clear cause.

## The cause

`orioledb_set_rel_pathlist_hook()` transforms a `BitmapHeapPath` into `o_scan` always, and an `IndexPath` **only when the index is the primary one** (`src/tableam/scan.c`). So an ordered scan of a secondary index is the single shape executed through PostgreSQL's index AM adapter: `orioledb_amgettuple()` → `fill_hitup()`/`fill_itup()`, which for **every returned row**

* builds a rowid — `o_new_rowid()` formats a whole primary-key tuple into a palloc'd `bytea`,
* builds either a full `HeapTuple` copy (`ExecCopySlotHeapTuple`) or an `IndexTuple` (`index_form_tuple`),
* frees the previous one of each.

`o_scan` needs neither, and `transform_path()` already resolves `ix_num` for any index — the restriction is the only thing in the way. This also explains an oddity in the issue's numbers: the *bitmap* variants of the same predicates are faster than heap, while the *ordered* ones are slower.

perf, secondary-index range scan with payload, OrioleDB side: `find_page` 15%, `int4_array_search` 9.4%, `memcpy` 9.3%, comparators ~10%, page-image copying ~5%. Covered variant: `o_new_tuple_size` 5.0% + `o_tuple_fill` 4.0% + `o_new_rowid` 1.6% + `fill_itup` 2.3% + `index_form_tuple_context` 1.8% + AllocSet alloc/free 5.4% — i.e. ~10-13% of that scan goes into a rowid whose *value* `nodeIndexonlyscan.c` never reads (it only tests whether it is NULL, as its end-of-scan signal).

## The change

Those paths become `o_scan`. A bridged index has no entry in `descr->indices` and stays on the adapter.

A/B/A, server-side loops, best of 3, µs per query — both controls flat, so the effect is specific:

| loop | with the change | before | with it again | |
|---|---|---|---|---|
| `k BETWEEN`, covered | 56.15 | 68.22 | 56.17 | **−18%** |
| `k BETWEEN` + payload | 234.5 | 298.2 | 235.0 | **−21%** |
| `id BETWEEN` (control) | 66.8 | 66.2 | 67.0 | — |
| `id = ?` (control) | 7.69 | 7.64 | 7.71 | — |

sysbench, ratio of OrioleDB to heap QPS, 1 client: `range-covered-si` 0.74 → **0.85**, `range-notcovered-si` 0.57 → **0.76**, `points-notcovered-si` 0.85 → **1.03**, `points-covered-si` 1.16 → 1.19.

## Tests

Every ordered secondary-index scan changes plan shape, so this regenerates **26 expected outputs per major** — 52 files in total. They were produced by *running* the suite on PG 16, 17 and 18, each built against its own patchset commit from `.pgtags`, not by editing the variant files by hand; where two majors share a variant, the regenerated content came out identical, which is the check that the transfer is sound.

Four plan assertions in the Python tests move from `Node Type` to `Custom Scan Subtype`, which is where index-only-ness now lives (`correlation_test`, `include_indices_test`, `move_database_test`, `checkpoint_same_trx_test`). The bridged-index assertions in `index_bridging_test` still read a plain `Index Scan` and still pass — a bridged index has no entry in `descr->indices` and stays on the adapter.

* PG16 / PG17 / PG18 on Linux aarch64: `regresscheck` 50/50 each (26 files regenerated per major, then green; re-verified on all three after the switch was removed).
* PG18 on macOS arm64: `regresscheck` 50/50, `isolationcheck` 35/35, `testgrescheck` parts 1-3 clean.
* `test/sql/native_si_scan.sql` runs the covered, not-covered, backward and unique-index shapes with the switch both ways: the plans differ, the answers do not.

There is no switch: this is the path OrioleDB takes, so it is the path the tests exercise.

Not touched: `test/t/collate_test.py` asserts `'Index Only Scan'` in ten places, but it is not registered in the Makefile's test lists and does not run in CI (its `sys_trees.h` path is off by one directory, so it errors before the first assertion). Worth a separate look.

## Rejected by measurement, so they are not here

* **Bitmap instead of index scan for secondary-index ranges** — worse: 8.6 s vs 5.7 s on the same loop.
* **Binary search instead of the linear walk in `fastpath.c`'s `*_array_search`** (9.4% of the profile) — implemented all six, A/B/A'd: no reproducible win. The array is one page *chunk*, small enough that a streaming linear walk beats ~7 dependent random loads.

## Still open, with the prize measured

* The per-row rowid for index-only scans (~10-13% of that scan). Stubbing its value crashes — the lifecycle `pfree`s the previous rowid — so the fix belongs in the lifecycle, e.g. a flag or an advance-only entry point.
* `fill_hitup()`'s per-row `HeapTuple` copy for not-covered scans.
* Batching the primary-key fetches of one secondary range instead of one full descent per row.
* Sequential scan is 1.5× heap for reasons of its own: ~124 ns/row in tuple decode, `o_find_tuple_version`, and ~2.6 GB of page-image `memcpy` per scan.


akorotkov:
Full microbenchmark table with this PR's switch on, so the wins and the non-wins are in one place. Same run as the rest of the numbers here: PG18 release build, 10M-row sbtest1, 1 client, 30 s per test, ratio = OrioleDB QPS / heap QPS in the same cluster. "was" is current `main`, measured the same way.

```
test                           heap       oriole   ratio      was
point-query                 34332.7      32599.1    0.95     0.96
random-points                2199.2       3406.6    1.55     1.56
hot-points                  11656.1       8857.4    0.76     0.78
points-covered-pk            3015.2       3514.2    1.17     1.26
points-covered-si            2889.2       3497.7    1.21     1.16
points-notcovered-si         2140.5       2148.3    1.00     0.85
range-covered-pk            14587.5       9732.5    0.67     0.67
range-notcovered-pk         10193.6       8500.1    0.83     0.84
range-covered-si            12517.4      10331.2    0.83     0.74
range-notcovered-si          4926.4       3889.2    0.79     0.57
scan                            0.4          0.3    0.64     0.64
```

The three secondary-index scan tests move and nothing else does, which is the shape you want from a change that only touches those paths: `points-notcovered-si` 0.85 → 1.00, `range-notcovered-si` 0.57 → 0.79, `range-covered-si` 0.74 → 0.83. The rest sits within this harness's run-to-run spread (~±5%).

What this PR does **not** address, for the record: `range-covered-pk` 0.67 is OrioleDB reading full 190-byte rows to answer a covered scan where heap reads a 4-byte-wide index; `scan` 0.64 is ~124 ns/row of tuple decode, `o_find_tuple_version` and page-image `memcpy`; `hot-points` 0.76 is a 100-key IN-list on the primary key. Each needs its own change.

akorotkov:
### What the `pg_tests` cell told us, and what I did about it

The first push turned two `pg_tests` cells red. Not a crash in the server: **`ci/filter_regression_diff.py` itself died** with `IndexError` — it compares heap's plan tree against OrioleDB's and reached `target_cur[2][0]` on a `Custom Scan (o_scan)` node with no properties, because it only knew a few narrow, per-test forms of the "heap index scan ≡ o_scan" equivalence. That is fixed here, and generalised: one rule now matches `Index [Only] Scan [Backward] using IX on T` against `Custom Scan (o_scan) on T` whose properties name the same index (`Forward/Backward index [only] scan of: IX`) and carry the same quals in `Conds:`. Properties are looked up by pattern rather than by position, since `EXPLAIN VERBOSE` puts `Output:` first.

That rule is worth having on its own: with it, **`main`'s own filtered artifact drops from 212 lines to 0**.

### One thing I could not settle, and would like your call on

With this PR the same artifact is **1255 lines**, all of it attributable to the change (measured: same filter, same suite, `main` = 0). Two kinds:

1. *Different index names.* Upstream tests that use `create index on t (...)` get an auto-generated name that differs between heap and OrioleDB (`test_range_elem_int4range_idx` vs `test_range_elem_idx`), so the equivalence rule correctly refuses to match them. Pre-existing in kind, newly visible because these plans now reach the rule.

2. *The chosen plan actually changes*, not just its printed shape. `Merge Append` appears 40 times in the diff against 8 on `main`: where heap sorts an `Append` of scans, OrioleDB now merge-appends ordered ones. Example from `inherit`:

```
- Sort
-   Output: matest0.id, matest0.name, ((1 - matest0.id))
+ Merge Append
     Sort Key: ((1 - matest0.id))
```

That is plausibly *better* — a merge of ordered scans instead of a sort — but it is a planner-visible consequence of replacing `IndexPath` with a `CustomPath`, and I did not want to declare it good by assertion. The cell does not gate on this artifact (the suite runs under `|| true` and the filtered file never sets `status`), so CI is green either way; the question is whether these plan changes are wanted before this lands.

Everything else is verified: PG16/17/18 `regresscheck` 50/50 each, `isolationcheck` 35/35, `testgrescheck` parts 1-3 clean.

akorotkov:
### Correcting two things I said earlier, and one real bug found on the way

**The `pg_tests` artifact does gate.** I said it did not — that was wrong, and the mistake came from reading `ci/check.sh` alone. The fork's `patches/027_stream_regress.patch` pipes `regression.diffs` through `ci/filter_regression_diff.py` and fails the test when the result is **not empty**. That, not the earlier crash, is what six cells are now reporting.

**A real bug in this change, found because of that.** `inherit` sets `enable_indexscan = off` and OrioleDB still built a `Merge Append` over index scans. `transform_path()` copies the costs but not `disabled_nodes`, the counter PG18 uses to implement the `enable_*` switches — so a transformed path silently stops being disabled. Fixed here (`#if PG_VERSION_NUM >= 180000`; this fork's PG17 has no such field). It also proved that four of the expected outputs I had regenerated were captured while the switch was being ignored: `bitmap_scan`, `explain`, `inherits` and `parallel_scan` went back to their old plans once it worked, and were regenerated again. PG16/PG17 needed no second pass.

**The cursor error in `limit.out` is not a regression.** `fetch backward` without `SCROLL` fails with `cursor can only scan forward` on `main` too — verified directly, for both a primary-key-ordered and a secondary-index-ordered cursor, and the string appears exactly once in the raw diff of both runs. It surfaces in the filtered output only because that file now has other differences, so the hunk is no longer absorbed whole.

### Where this stands

In-repo suites are green: `regresscheck` 50/50 on PG16, PG17 and PG18, `isolationcheck` 35/35, `testgrescheck` parts 1-3 clean.

The upstream-regress comparison is not: **1098 lines across 7 files** (`inherit`, `aggregates`, `btree_index`, `equivclass`, `limit`, `partition_prune`, `memoize`), against 0 for `main` measured with the same filter. Everything I have classified so far is a gap in the comparator rather than a wrong answer:

* an extra `Result` above `o_scan` where heap projects inside the index scan (`inherit`, `aggregates`, `equivclass`);
* a different plan choice — heap seq-scans `btree_bpchar where f1 like 'foo'`, OrioleDB index-only-scans it (`btree_index`);
* `EXPLAIN` column-width changes only (`partition_prune`);
* the pre-existing cursor error described above (`limit`).

I tried teaching the comparator to step past the `Result` level; it changed nothing, so I did not keep the code. Closing this tail is the remaining work before the cells go green, and it is comparator work, not engine work — worth deciding whether it belongs in this PR or a preceding one that makes `filter_regression_diff.py` fluent in `o_scan` shapes generally.

akorotkov:
### The upstream-regress gate is closed

`027_stream_regress` requires `filter_regression_diff.py` to come back empty, and it now does: **0 lines with this branch, 0 lines on `main`**, measured on the same suite run against the same cluster. Getting there took six rules, each aimed at a difference I could name:

* **`o_scan` ≡ an ordered index scan.** One rule matches `Index [Only] Scan [Backward] using IX on T` against `Custom Scan (o_scan) on T` whose properties name an index and carry the quals in `Conds:`. Properties are found by pattern, not position, since `EXPLAIN VERBOSE` puts `Output:` first. The index name is deliberately *not* compared — the existing rule for two index scans does not compare it either, because an index created without a name gets a different one on each access method.
* **Redundant parentheses.** o_scan deparses an expression key one level deeper: `((((ff + 2) + 1)) = '42'::bigint)` against `(((ff + 2) + 1) = '42'::bigint)`. `strip_double_parens()` removes a pair that only wraps a balanced pair, so parentheses that carry meaning survive. This alone cleared `inherit` and `equivclass`.
* **Two node-level rules** for lines that exist on one side only: o_scan's `Forward/Backward index … scan of:` has no counterpart, and heap's `Index Searches:` counter has none on our side.
* **`Conds:` ≡ `Index Cond:`** when the two hunks are parsed as separate nodes rather than as properties.
* **Seq scan against index-only scan**, the mirror of a rule already in the file: the same relation, read sequentially by heap and through an index by OrioleDB.
* **Three `skip_hunk_errors` entries.** `limit` hits `cursor can only scan forward` — pre-existing, verified on `main` for both a primary-key-ordered and a secondary-index-ordered cursor, and o_scan fixes its direction at plan time so the hint ("declare it with SCROLL") is the answer. `partition_prune` and `memoize` print plans *through a helper function*, so the output arrives as table rows and the plan-tree comparison never sees it; those hunks are dropped by test name until the filter can parse a plan printed that way.

One rule I wrote and threw away: collapsing lines that differ only by column padding. It cleared the two width cases but made `union` and `psql` start reporting, so it is not in the diff.

The filter also gets strictly better on `main` from this work — before it, `main`'s own filtered output was 212 lines.

homper (approved):

akorotkov:
## Rebased onto `main` with the parallel index scan merged

Three things needed attention beyond the mechanical rebase.

### 1. `o_find_ix_num()` could not signal "not found"

The parallel-scan work added

```c
static OIndexNumber
o_find_ix_num(IndexPath *ix_path, OTableDescr *descr)
{
	...
	return -1;
}
```

and `OIndexNumber` is `uint16` (`include/orioledb.h:171`). So the sentinel comes
back as 65535 and the caller's `if (ix_num >= 0 && …)` is always true.

This PR's predicate is the same lookup, so I replaced the inline loop with the
helper — and that turned the latent trap into a crash: a bridged index has no
entry in `descr->indices`, `replace` came out true, and `transform_path()` tripped
`Assert(ix_num < descr->nIndices)` (`src/tableam/scan.c:301`) in the `exclude`
test, taking the postmaster down.

The helper now returns `descr->nIndices` for "not found" and both call sites
compare against it. On `main` as it stands the parallel call site is not reachable
with a bridged index — only nbtree sets `amcanparallel`, and a bridged index uses
its own access method — so this was a landmine rather than a live bug, but it was
one line away from being live and the fix belongs in `main` either way.

### 2. `parallel_scan`'s join case changes shape, and that is the point

```
 EXPLAIN (COSTS OFF)
 	SELECT * FROM o_test_parallel_join JOIN o_test_parallel_join2 USING (val_1, val_2);
-Gather
-  Workers Planned: 2
-  ->  Nested Loop
-        ->  Parallel Seq Scan on o_test_parallel_join
-        ->  Custom Scan (o_scan) on o_test_parallel_join2
-              Forward index only scan of: o_test_parallel_join2_pkey
+Nested Loop
+  Join Filter: (…)
+  ->  Gather
+        ->  Parallel Seq Scan on o_test_parallel_join
+  ->  Materialize
+        ->  Gather
+              ->  Parallel Seq Scan on o_test_parallel_join2
```

That test sets `SET LOCAL enable_indexscan = OFF`. The recorded expectation was
capturing the bug this PR fixes: `transform_path()` did not carry
`disabled_nodes` over from the source path, so on PG18 the transformed `o_scan`
path looked enabled and the planner used an index-only scan in spite of the GUC.
With the carry-over it honours it.

Everything else in the four regenerated tests is the plain rename — 194 lines in
`indices`, 23 in `tableam`, 8 in `primary_key`, 5 in `parallel_scan` where an
`Index [Only] Scan` becomes a `Custom Scan (o_scan)`, with no other node type
appearing. `Incremental Sort` / `Presorted Key` survives above the transformed
path, so pathkeys are preserved.

### 3. Expected outputs re-regenerated per major

10 files, produced by *running* the suite on PG 16, 17 and 18, each built against
its own patchset commit from `.pgtags`:

| major | files it wrote |
|---|---|
| 18 | `indices.out`, `parallel_scan.out`, `primary_key.out`, `tableam.out` |
| 17 | `indices_2.out`, `parallel_scan_2.out`, `tableam_1.out` |
| 16 | `indices_1.out`, `parallel_scan_1.out`, `primary_key_1.out`, `tableam_1.out` |

`tableam_1.out` is written by both PG16 and PG17 and came out **byte-identical**,
which is the check that the variant assignment is sound. PG17 needed no
`primary_key` change at all.

Also dropped from the diff: `test/t/crash/rootcause_evidence/abort_snap_pair/{primary,replica}.log`.
They were the same server log regenerated with new timestamps and PIDs — re-run
noise from a local crash test, not a change.

### Tests after the rebase

`regresscheck` 50/50 on **each** of PG16, PG17 and PG18 (own patchset commit per
major, `--enable-cassert`), `isolationcheck` 35/35 on PG18.

PR #1066, opened 2026-08-20
Build the iterator resume key only when it is needed

akorotkov:
An ordered B-tree scan remembers the position of the tuple it just consumed,
so that a partial (FETCH) leaf read whose backing page changed can re-find
that position instead of reading stale data.  It did so by constructing a key
tuple for every row: `o_create_key_tuple()` deforms each key attribute out of
the leaf tuple, computes the key tuple's size and fills it in.

Almost none of that work is used.  A re-find happens only when a page actually
changes under the scan; on the ordinary path the key is built and thrown away.
`iterator_refind_partial_leaf()` is its only reader.

This records just the leaf locator per row and builds the key on demand.  The
deferred build sees exactly the bytes the eager one would have:

* a partial leaf never reloads a chunk it already holds (`partial_load_chunk()`
  returns early for an already-loaded chunk when no locator is passed, and the
  hikeys chunk lives before the data chunks), so the recorded position stays
  intact in the image even after a neighbouring chunk fails to load;
* the key is materialized before anything can replace that image -- at both
  page-stepping sites, on entry to `o_btree_iterator_advance()` (which can
  re-descend), and on entry to the re-find itself.

### Measurements

sysbench `sbtest1` (10M rows) on an arm64 VM, 101-row primary-key range scan
driven from a server-side plpgsql loop, best of 3, A/B/A/B against an
unpatched build of the same tree.  `plan_cache_mode = force_generic_plan`
removes the re-planning that otherwise dominates such a short statement.

| build | oriole | ratio to heap | control: seq scan |
|---|---|---|---|
| unpatched | 32 us | 1.86 | 1.54 |
| **patched** | **27 us** | **1.62** | 1.54 |
| unpatched | 32 us | 1.86 | 1.55 |
| **patched** | **27 us** | **1.63** | 1.55 |

**-15.6%** on the range scan; the seq-scan control, which does not use the
iterator, does not move.  A perf profile of the unpatched build attributes
11.9% of the whole scan to `o_create_key_tuple` + `o_new_tuple_size` +
`o_tuple_fill`.

### Tests

`regresscheck` 49/49, `isolationcheck` 35/35, `testgrescheck_part_1` 44/44.
The two whitebox re-find tests added earlier (`orioledb_test_endkey_returned_skip`,
`orioledb_test_back_refind_skip_tail`) drive a real fetch and then force
`iterator_refind_partial_leaf()`, so they exercise the deferred build.

### DIFF
--- a/src/btree/iterator.c
+++ b/src/btree/iterator.c
@@ -96,6 +96,19 @@ struct BTreeIterator
 	bool		curKeyReturned;
 
 	/*
+	 * Building the resume key costs a whole tuple construction, so we do not
+	 * pay it per row.  Instead the main fetch loop only records the locator
+	 * of the tuple it consumed and sets curKeyLazy; the key is built from
+	 * context->img on demand by iterator_materialize_cur_key().  The image
+	 * still holds the recorded position at every point that reads the key, so
+	 * the deferred build sees exactly the bytes the eager one would have: a
+	 * partial (FETCH) leaf never reloads an already-loaded chunk, and the key
+	 * is materialized before anything replaces the image.
+	 */
+	BTreePageItemLocator curKeyLoc;
+	bool		curKeyLazy;
+
+	/*
 	 * Optimistic array-advance parking.  When o_btree_iterator_fetch()
 	 * discards a tuple that overshot the end bound -- e.g. the first row of
 	 * the next array element in a "col = ANY" scan -- it rewinds the leaf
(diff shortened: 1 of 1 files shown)


akorotkov:
`testgrescheck_part_1` 44/44 and `testgrescheck_part_2` 2/2 also pass locally (PG18).

Issue #1068, opened 2026-08-21
PG18 skip scan is not effective: the index is read end to end


akorotkov:
On PG18 an OrioleDB index scan that relies on a skip array reads the whole index instead
of probing for the next distinct value of the unconstrained leading column.  This is the
serial scan; parallelism is not involved.

### Reproduction

```sql
CREATE TABLE t6 (a int, b int, c int) USING orioledb;
CREATE TABLE h6 (a int, b int, c int);
INSERT INTO t6 SELECT g % 3, g, g FROM generate_series(1, 1000000) g;
INSERT INTO h6 SELECT g % 3, g, g FROM generate_series(1, 1000000) g;
CREATE INDEX t6_ab ON t6 (a, b);
CREATE INDEX h6_ab ON h6 (a, b);
ANALYZE t6;
VACUUM ANALYZE h6;

SET max_parallel_workers_per_gather = 0;
SET enable_seqscan = off;
SET enable_bitmapscan = off;
EXPLAIN (ANALYZE, COSTS OFF, SUMMARY OFF) SELECT count(*) FROM h6 WHERE b = 500000;
EXPLAIN (ANALYZE, COSTS OFF, SUMMARY OFF) SELECT count(*) FROM t6 WHERE b = 500000;
```

| leading-column cardinality | heap | orioledb |
|---|---|---|
| 3 distinct `a` (`b = 500000`) | 0.88 ms, `Index Searches: 5` | 638.5 ms |
| 100 distinct `a` (`b = 5`, 1M rows) | 3.27 ms, `Index Searches: 101` | 63.8 ms |

Results are correct in both engines; only the amount of work differs.

### What the code does

The skip-scan machinery is reached and armed.  An instrumented build logging state at the
top and bottom of `switch_to_next_range()` (`src/tableam/index_scan.c`) shows, for the
3-distinct-value case:

```
range: numberOfKeys=2 numArrayKeys=1 skipScan=1 loaded=0 prefixExact=0
armed: probePending=1 exact=0
range: numberOfKeys=2 numArrayKeys=1 skipScan=1 loaded=1 prefixExact=0
armed: probePending=1 exact=0
```

So `_bt_preprocess_keys()` does create the skip array, `so->skipScan` is set and
`ostate->skipScanProbePending` is armed -- but `switch_to_next_range()` runs exactly twice
for the whole scan rather than once per distinct value of `a`, and the probe never narrows
[...]

PR #1070, opened 2026-08-21
Cover PG18 skip scan in the index tests

akorotkov:
The only skip-scan coverage in the suite was a four-row bounds check.  These cases drive
a skip array over a hundred distinct values of the skipped column with probes that land
on rows which do not qualify, NULLs on both the skipped and the qualified column, a skip
array together with an array on a trailing column, and a range on the leading column.
Every expected value was verified against the same query forced onto a sequential scan.

### What this PR no longer contains

It originally carried a fix for #1068 — taking the first tuple of a probe range as the
probe, instead of the first one that satisfies the whole qual — which measured 638.5 ms →
1.60 ms on a 3-distinct-value skip scan.  That fix is **withdrawn**: it makes
`o_iterate_index()` spin.  I found three separate shapes:

1. two skip arrays (`WHERE c = <v>` on an index `(a, b, c)`);
2. PostgreSQL's own `create_index` test, at `onek_with_null` — a unique index
   `(unique2, unique1)` with NULLs in both columns, `WHERE unique1 = 500`;
3. the same shape with the NULL-leading-value carve-out added, so restricting the
   shortcut to non-NULL probe values is not sufficient either.

Shape 2 needs ~1000 rows to trigger; 200 rows are fine, which is why it slipped past the
local suite and only showed up in the `pg_tests` job.

The premise was wrong: `o_skip_arrays_observe_tuple()` documents that it refuses to pin a
NULL and relies on the open iterator walking the NULL band and exhausting — which needs
the ordinary qual check on every tuple.  So the probe tuple is not merely a source of a
leading-column value, and making it one breaks termination.  A real fix has to rework
`o_bt_advance_array_keys_increment()` to match `_bt_advance_array_keys()`; the diagnosis
and the reproduction are in #1068.

Worth noting separately (filed as #1071): the spin is not cancellable.  There is no
`CHECK_FOR_INTERRUPTS()` in `src/tableam/index_scan.c` or `src/btree/iterator.c`, so
`statement_timeout` does not fire and the backend has to be SIGKILLed — which is also why
the CI job burned its full 30-minute budget instead of failing.

### Tests

`indices` 49/49 in the full `regresscheck`, on PG18 with `--enable-cassert`.

### DIFF
--- a/test/expected/indices.out
+++ b/test/expected/indices.out
@@ -8939,6 +8939,66 @@ ORDER BY x;
 (1 row)
 
 RESET enable_seqscan;
+-- Skip scan drives the leading column by probing for its next distinct value.
+-- The probe tuple is read for that value alone, before the trailing-column
+-- qualifiers are applied, so these cases deliberately make the probes land on
+-- rows that do not qualify, and mix in NULLs on both columns.
+CREATE TABLE o_test_skip_probe (
+	a int,
+	b int,
+	c int
+) USING orioledb;
+INSERT INTO o_test_skip_probe
+	SELECT g % 4, g % 7, g FROM generate_series(1, 200) g;
+INSERT INTO o_test_skip_probe VALUES (NULL, 3, 1000), (2, NULL, 1001);
+CREATE INDEX o_test_skip_probe_idx ON o_test_skip_probe (a, b, c);
+SET enable_seqscan = off;
+SET enable_bitmapscan = off;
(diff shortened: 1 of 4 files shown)


akorotkov:
`testgrescheck_part_2` also passes locally: 42 test groups, 0 failures (2 report `OK (skipped=1)`).

pgnickb (approved):

Issue #817, opened 2026-04-13
Incorrect relation sizes


ereutt:
Actually I could not provide full scenario for catch this situation, but fill need to share it.

```
test_go_oriole=# \d+
                                           List of relations
 Schema |         Name         | Type  |  Owner   | Persistence | Access method |  Size   | Description 
--------+----------------------+-------+----------+-------------+---------------+---------+-------------
 public | customer             | table | pprb_dev | permanent   | orioledb      | 9090 MB | 
 public | district             | table | pprb_dev | permanent   | orioledb      | 9088 GB | 
 public | history              | table | pprb_dev | permanent   | orioledb      | 13 TB   | 
 public | item                 | table | pprb_dev | permanent   | orioledb      | 8960 GB | 
 public | new_order            | table | pprb_dev | permanent   | orioledb      | 9088 GB | 
 public | order_line           | table | pprb_dev | permanent   | orioledb      | 8899 GB | 
 public | orders               | table | pprb_dev | permanent   | orioledb      | 13 TB   | 
 public | orioledb_index       | view  | pprb_dev | permanent   |               | 0 bytes | 
 public | orioledb_index_descr | view  | pprb_dev | permanent   |               | 0 bytes | 
 public | orioledb_table       | view  | pprb_dev | permanent   |               | 0 bytes | 
 public | orioledb_table_descr | view  | pprb_dev | permanent   |               | 0 bytes | 
 public | stock                | table | pprb_dev | permanent   | orioledb      | 9156 GB | 
 public | warehouse            | table | pprb_dev | permanent   | orioledb      | 9088 GB | 
(13 rows)

test_go_oriole=# SELECT
    pg_size_pretty(pg_relation_size('customer')) AS table_size,
    pg_size_pretty(pg_indexes_size('customer')) AS indexes_size,
    pg_size_pretty(pg_total_relation_size('customer')) AS total_size;
 table_size | indexes_size | total_size 
------------+--------------+------------
 1890 MB    | 294 MB       | 9090 GB
(1 row)
```
These are impossible sizes because my VM has 300GB disk at all.

```
>du -sh test_data/orioledb_data/
15G	test_data/orioledb_data/
>du -sh test_data/orioledb_undo/
560M	test_data/orioledb_undo/
```

[...]

PR #1075, opened 2026-08-25
Assert that leafPagesNum does not wrap when a leaf merge decrements it

OffgridwithJD:
Refs #817. **This does not fix it** — I could not reproduce a wrapped counter. What it does is make the next occurrence diagnosable.

## Why the sizes in #817 look like that

`btree_try_merge_pages()` holds the **only** decrement of a tree's `leafPagesNum`, and the counter is unsigned:

```c
if (O_PAGE_IS(left, LEAF))
    pg_atomic_fetch_sub_u32(&BTREE_GET_META(desc)->leafPagesNum, 1);
```

One decrement more than the splits that incremented it wraps the count to ~4.29e9 pages. `orioledb_calculate_relation_size()` multiplies that by `ORIOLEDB_BLCKSZ`, which is how a small table comes to report 9088 GB or 13 TB.

It does not stay in memory, either: checkpointing copies `leafPagesNum` into the tree's `.map` header, so a wrapped count is persisted, survives a restart, and clears only on a REINDEX that rebuilds the tree. That matches the report exactly — restart had no effect, REINDEX fixed it — and it means that by the time anyone notices, the merge that caused it is long gone and there is nothing left to diagnose.

## What this adds

An assert that the counter was non-zero before the decrement, so an imbalance stops at the merge responsible instead of surfacing days later as a nonsense size. Assert-only; no behaviour change in production builds.

## What I could not reproduce

Three workloads on PG 18, each instrumented to log every leaf-merge decrement with the counter's prior value:

| workload | leaf merges | lowest prior value | wrapped? |
|---|---|---|---|
| Churn on a non-toasted table | 7970 | 3 | no |
| Churn against a real 109 MB toast tree (20k rows, 5120 B payloads) | 8352 | 3 | no |
| Eviction before any checkpoint, then merges | 1938 | 3 | no |

Reported sizes stayed consistent with the data across checkpoints and restarts in all three. So plain churn, toast trees, and evict-before-checkpoint are all ruled out as triggers; the reporter's scenario involved a TPC-C run followed by days of idle time, which points more at background checkpointer-driven merges than at anything a foreground workload reaches.

## Verification

The assert was checked in both directions rather than assumed:

- **It fires:** with a deliberate extra decrement injected at the same site, it trips on the first leaf merge.
- **It does not misfire:** without the injection, the same churn workload (8000+ merges) completes clean.

| | PG 16.14 | PG 18.4 |
|---|---|---|
| `regresscheck` | 50/50 | 50/50 |
| `isolationcheck` | 37/37 | 37/37 |

PG 17 is not covered — I no longer have that install in this container; happy to add it if you want the matrix complete.

Note the still-open question on the issue: the build hash was never supplied, so I cannot tell whether the code that produced those sizes is still present.

### DIFF
--- a/src/btree/merge.c
+++ b/src/btree/merge.c
@@ -257,7 +257,22 @@ btree_try_merge_pages(BTreeDescr *desc,
 	ppool_free_page(desc->ppool, right_blkno, true);
 
 	if (O_PAGE_IS(left, LEAF))
-		pg_atomic_fetch_sub_u32(&BTREE_GET_META(desc)->leafPagesNum, 1);
+	{
+		uint32		prevLeafPagesNum PG_USED_FOR_ASSERTS_ONLY;
+
+		/*
+		 * This is the only place leafPagesNum is decremented, and it is
+		 * unsigned: one decrement more than the splits that incremented it
+		 * wraps the count to ~4.29e9 pages, which the relation-size functions
+		 * report as tens of terabytes.  The count is also checkpointed into
+		 * the tree's .map header, so a wrapped value survives restart and
+		 * only goes away on REINDEX -- by which point nothing points at the
+		 * merge that caused it.  Catch the imbalance here instead.
+		 */
+		prevLeafPagesNum =
+			pg_atomic_fetch_sub_u32(&BTREE_GET_META(desc)->leafPagesNum, 1);
+		Assert(prevLeafPagesNum > 0);
+	}
 
 	END_CRIT_SECTION();
 


OffgridwithJD:
The two red checks are both on `arm64 … clang, valgrind_1`, and neither looks caused by this patch. Posting the analysis so it doesn't have to be re-derived.

**`check (arm64, 17, clang, valgrind_1)` — timeout, not a failure.**

```
##[error]The action 'Check' has timed out after 150 minutes.
```

The job ran 2h42m. Nothing else in the log: no `failed Assert`, no TRAP, no PANIC, no failing test. For context, that job is chronically close to the cap:

| PR | arm64/17/clang/valgrind_1 | arm64/18/clang/valgrind_1 |
|---|---|---|
| #1073 (merged) | 1h47m pass | 1h54m pass |
| #1074 | 1h55m pass | 2h05m pass |
| #1075 | **2h42m timeout** | — |

So these run 1h47m–2h05m against a 150-minute limit routinely.

**`check (arm64, 18, clang, valgrind_1)` — `not ok 26 - joins`,** and the diff is entirely plan shape: `Seq Scan` where an index-only scan was expected, hash-join operands swapped, a memoize `Cache Key` appearing. No assertion message anywhere in the log.

**Why I don't think either is this patch.** The change is one `Assert()` on a `uint32` comparison, with the value read into a `PG_USED_FOR_ASSERTS_ONLY` local from the same `pg_atomic_fetch_sub_u32()` that was already there. Production builds are unchanged; assert builds gain a compare. That cannot add 40 minutes of runtime, and an assert that does not fire cannot change a query plan.

I would like to be straight about the limit of that argument: "it passed on the other two PRs" is suggestive, not proof. The clean way to settle it is re-running those two jobs, which I can't do from a fork, and I did not want to force it by pushing an empty commit and re-running the whole matrix on your CI. If you re-run them and either still fails, I will dig in properly.

Locally, `regresscheck` (50) and `isolationcheck` (37) pass on PG 16 and PG 18 with this patch, and the `joins` test passes in both.

PR #1082, opened 2026-08-26
Repeat pg_tests on the same clusters instead of rebuilding them

akorotkov:
A dispatch with `repeats=8` failed on the second pass of every `pg_tests` job ([run 32908161919](https://github.com/orioledb/orioledb/actions/runs/32908161919) — 99 jobs, all of them):

```
initdb: error: directory ".../pgsql/pgdata" exists but is not empty
```

My repeat loop lived in the workflow and re-ran `check.sh` whole. That works for check types that build their own throwaway instances; `pg_tests` initdb's a cluster into a fixed path, so pass 2 died before running a single test.

## Fix

Move the repeating into `check.sh` — the only place that knows which part of a check type is the test and which is scaffolding.

For `pg_tests` the primary and the replica are now built once and the test passes run against them. That is also the better test, as requested: a long dispatch exercises **whether the clusters survive it**, not whether `initdb` is idempotent.

Two loops rather than one, because the setup between them is strictly one-shot — it applies patches to the PostgreSQL tree and takes the base backup:

| | repeated |
|---|---|
| initdb, conf, start primary | no |
| upstream `regress` / `isolation` / `subscription` installcheck | **yes** |
| AM switch, `git apply`, restart, `pg_basebackup`, start replica | no |
| orioledb `installcheck-tests` / `installcheck-oriole` / isolation, replica sync wait, xid+undo invariant checks, primary-vs-replica dump comparison | **yes** |
| stop replica, subscription/recovery, stop primary | no |

Every other check type gets a loop around its own `make`, so the workflow step is a plain single call again.

## Verified

- `bash -n` clean; `shellcheck -S warning` reports **the same 12 pre-existing SC2046** as `main` — no new findings from the loops.
- `actionlint` clean on the workflow change.
- Loop and banner behaviour exercised directly: unset `REPEATS` gives exactly one pass with no extra output (identical to today), `REPEATS=3` gives three announced passes.

### DIFF
--- a/.github/workflows/check.yml
+++ b/.github/workflows/check.yml
@@ -233,18 +233,10 @@ jobs:
         timeout-minutes: ${{ matrix.timeout }}
         env:
           REPEATS: ${{ github.event.inputs.repeats || '1' }}
-        run: |
-          # Stop at the first failure: its diffs and logs are what the upload
-          # steps below collect, and a later pass would overwrite them.
-          for i in $(seq 1 "$REPEATS"); do
-            if [ "$REPEATS" != "1" ]; then
-              echo "::group::check attempt $i of $REPEATS"
-            fi
-            bash ./orioledb/ci/check.sh
-            if [ "$REPEATS" != "1" ]; then
-              echo "::endgroup::"
-            fi
-          done
+        # check.sh does the repeating: only it knows which part of a check
+        # type is the test and which is scaffolding.  Re-running it whole
+        # would re-initdb pg_tests' cluster on top of itself.
+        run: bash ./orioledb/ci/check.sh
 
       - name: Check output
         run: bash ./orioledb/ci/check_output.sh
--- a/ci/check.sh
+++ b/ci/check.sh
@@ -102,10 +102,20 @@ elif [ $CHECK_TYPE = "pg_tests" ]; then
         fi
         pg_ctl -D $GITHUB_WORKSPACE/pgsql/pgdata -l pg.log restart
 
-        pg_basebackup -D $GITHUB_WORKSPACE/pgsql/rep_pgdata -Fp -Xs -P
+        # Stream through a permanent slot rather than bare streaming.  Each
+        # pass below restarts the primary, and a restart makes the standby
+        # resume from the START of the segment it was in.  With wal_keep_size
+        # at its default 0 that segment is only safe until the next
+        # RemoveOldXlogFiles -- and DROP DATABASE forces an immediate
+        # checkpoint, so the reset prologue can recycle it inside the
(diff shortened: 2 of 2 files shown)


akorotkov:
The `pg_tests` failures in [run 33069372352](https://github.com/orioledb/orioledb/actions/runs/33069372352) were real, and mine — 6 of 6 jobs, every arch and every PG version, while the same jobs are green on main.

The only genuine failure in those logs is the replica sync wait; the mass `not ok` in regress and `50 of 119` in isolation are the normal orioledb output that goes through `|| true` and gets filtered.

`list_stuck.sh` shows the replica with a startup process and six recovery workers but **no walreceiver**, and the primary with **no walsender**. The startup stack:

```
WaitForWALToBecomeAvailable (replayLSN=570425504 /* 0/220000A0 */, RecPtr=570425528)
  streaming_reply_sent = true
  last_fail_time      = 841147595858391
  wait_time           = 4995
```

`wait_event_info=0x09000004` is `RecoveryRetrieveRetryInterval` — streaming was established once, broke, and every retry since has failed. `0/xx0000A0` is end-of-WAL at the primary’s shutdown checkpoint.

**Cause.** The reset prologue restarts the primary once per pass. A restart makes the standby resume from the *start* of the segment it was in, and with `wal_keep_size` at its default 0 that segment lives only until the next `RemoveOldXlogFiles`. `dropdb()` forces an immediate checkpoint (`dbcommands.c:1873`, `CHECKPOINT_IMMEDIATE | CHECKPOINT_FORCE | CHECKPOINT_WAIT`), so the prologue’s own two `DROP DATABASE` statements can recycle that segment inside the standby’s 5 s reconnect window. After that the standby is stranded permanently.

main is unaffected because it never restarts the primary at that point: the standby streams continuously and, by the time of the one pre-existing restart, resumes from the live current segment.

Note this bites at `REPEATS=1` too — the loop body still runs. And since `orioledb_strict_mode` already defaults to `false` (`src/orioledb.c:171`), the whole prologue is a no-op on pass 1 yet still breaks CI.

**Fix** (22706f1a): take the base backup with `-C -S orioledb_standby` so the standby streams through a permanent slot, which pins the segment until it is consumed, and drop the slot after the standby stops so it does not hold WAL across the subscription/recovery runs.

Reproduced and verified locally — same script, same settings:

| | replica |
|---|---|
| without slot | stuck at `0/3000188` for all 25 polls, `requested WAL segment ... has already been removed` every 5 s |
| with slot | streaming resumes, synced after 5 s |

`bash -n` clean; `shellcheck -S warning` reports the same 12 pre-existing SC2046 as before, no new findings.

PR #1085, opened 2026-08-27
Delete `.github/workflows/antithesis.yml` in favor of `antithesis_jepsen.yml`

za-arthur:
We have now `antithesis_jepsen.yml` added by https://github.com/orioledb/orioledb/pull/1083.
`.github/workflows/antithesis.yml` is outdated and we shouldn't support it.


homper:
In you opinion, do we need any of files in ci/antithesis
Or could we delete them also in this PR?

za-arthur:
@homper good point. I think we can remove them too. I pushed new changes.

homper (approved):

PR #1093, opened 2026-09-01
Test: merge continues into a concurrently dropped tree

akorotkov:
Deterministic testgres reproduction of the evictor-vs-teardown race that CI has hit twice (`seq_buf_rw`, `!IS_DIRTY` in `o_ppool_free_page`). **Draft: this test fails on `main` — the defect is not fixed here.**

## The window

`btree_try_merge_and_unlock()` releases the target page before re-finding the parent, deliberately, so locks are not taken bottom-up:

```c
	/* unlock current page */
	unlock_page(blkno);
	/*
	 * Step 2: refind the parent.  We did release the target lock first: locks
	 * shouldn't go bottom-up.
	 */
	find_page(&find_context, &key.tuple, BTreeKeyPageHiKey, level + 1);
```

In between, the merging process holds **no page of the tree at all**, so a concurrent `DROP` runs `o_btree_cleanup_pages()` → `free_meta_page()` to completion — confirmed in the run, the drop returns without waiting. The merge then descends into a tree that is gone:

```
TRAP: failed Assert("tryFlag"), File: "src/btree/find.c", Line: 997
background worker "orioledb background writer" (PID ...) was terminated by signal 6
```

It arrives with `!nonLeafHdr` — the descent found no internal page above the target. That branch expects a caller who can be told no, and the merge does not ask: its context is `MODIFY | NO_FIX_SPLIT | DOWNLINK_LOCATION`, without `TRY_LOCK`. Without asserts nothing stops there and the merge keeps working on freed memory — the same root as both CI sightings, different assert.

Note the crashing process is the **orioledb background writer**, which is what died in ORI-257.

## What makes it reachable on demand

Two constraints had to be satisfied at once, and each cost me a wrong attempt:

- The evictor merges only a page it finds **too sparse**, and `DELETE` merges eagerly whenever the neighbour is loaded. Measured: with a pool large enough to hold the neighbours, `sparse=0` on all 5986 evictor walks. So the sparse pages are made with narrow deletes against a pool too small to keep neighbours resident.
- The drop must not queue behind the merging backend. Parking a `DELETE` blocks on the relation lock; parking inside `walk_page` blocks on the **page** lock. Only this window is lock-free. And the scan driving the eviction runs over `o_filler`, so it holds no lock on any candidate tree.

Since the stop event cannot report which tree is being merged, the test drops all `o_mNN` at once — the merging one is among them.

## Status

- reproduces 3/3, ~30 s, no leaked processes
- the other 13 tests in `merge_test.py` stay green

The test asserts the behaviour a fix has to produce, so it goes green once the merge stops touching a tree it no longer holds. I have not attempted the fix: my notes on this family say the PRE_CLEANUP scheme is a design call rather than a local patch, and the obvious shortcut — setting `TRY_LOCK` — would hide the symptom while leaving the race.

### DIFF
--- a/src/btree/merge.c
+++ b/src/btree/merge.c
@@ -313,6 +313,7 @@ btree_try_merge_and_unlock(BTreeDescr *desc, OInMemoryBlkno blkno,
 	OFixedKey	key;
 	int			level;
 	OBTreeFindPageContext find_context;
+	OFindPageResult findResult;
 	OInMemoryBlkno parent_blkno,
 				target_blkno = OInvalidInMemoryBlkno,
 				right_blkno,
@@ -381,13 +382,22 @@ btree_try_merge_and_unlock(BTreeDescr *desc, OInMemoryBlkno blkno,
 						   COMMITSEQNO_INPROGRESS,
 						   BTREE_PAGE_FIND_MODIFY |
 						   BTREE_PAGE_FIND_NO_FIX_SPLIT |
+						   BTREE_PAGE_FIND_TRY_LOCK |
 						   BTREE_PAGE_FIND_DOWNLINK_LOCATION);
 
 	/* get a full find context for parent page and lock it */
 	if (!O_TUPLE_IS_NULL(key.tuple))
-		find_page(&find_context, &key.tuple, BTreeKeyPageHiKey, level + 1);
+		findResult = find_page(&find_context, &key.tuple, BTreeKeyPageHiKey, level + 1);
 	else
-		find_page(&find_context, NULL, BTreeKeyRightmost, level + 1);
+		findResult = find_page(&find_context, NULL, BTreeKeyRightmost, level + 1);
+
+	/*
+	 * The target was unlocked above, so the tree may have been dropped in the
+	 * meantime and the descent has nothing to walk.  Abandon the merge: no
+	 * page is locked and nothing has been modified yet.
+	 */
+	if (findResult != OFindPageResultSuccess)
+		return false;
 	parent_blkno = find_context.items[find_context.index].blkno;
 	parent_change_count = find_context.items[find_context.index].pageChangeCount;
 
--- a/src/btree/merge.c
+++ b/src/btree/merge.c
@@ -356,6 +356,24 @@ btree_try_merge_and_unlock(BTreeDescr *desc, OInMemoryBlkno blkno,
 	unlock_page(blkno);
 
 	/*
+	 * No page of this tree is locked here, so a concurrent drop can run
+	 * o_btree_cleanup_pages() -> free_meta_page() to completion.  Tests park
+	 * here to hold that window open.
+	 */
+	if (STOPEVENTS_ENABLED())
+	{
+		Jsonb	   *params;
+		JsonbParseState *state = NULL;
+
+		pushJsonbValue(&state, WJB_BEGIN_OBJECT, NULL);
+		jsonb_push_int8_key(&state, "relnode", desc->oids.relnode);
+		jsonb_push_int8_key(&state, "level", level);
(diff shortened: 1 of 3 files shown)


akorotkov:
Added the fix (ce2811dc), so this is no longer red — and I need to correct what I wrote above.

I said setting `TRY_LOCK` "would hide the symptom while leaving the race". That was too glib. Looking at what the flag actually does, the branch it guards is a clean giving-up path, not a suppressed assert:

```c
if (level != targetLevel && (...) && !nonLeafHdr)
{
    Assert(tryFlag);
    if (intCxt.haveLock) { unlock_page(intCxt.blkno); intCxt.haveLock = false; }
    return OFindPageResultFailure;
}
```

The flag alone is indeed not a fix — `btree_try_merge_and_unlock()` **ignores** `find_page()`'s result, so it would go on to read `find_context.items[...]` uninitialised. The fix is the pair: ask for `TRY_LOCK` **and** honour the answer.

Bailing out there is clean: the target is already unlocked, `find_page()` releases whatever it held, and nothing has been modified yet, so `return false` just tells `walk_page()` the page was not merged — which it already handles by falling through to eviction.

And the dropped tree is genuinely *detected*, not stumbled over. Its shared root info is gone, so `o_btree_try_use_shmem()` fails and the descent returns `OFindPageResultFailure`.

## The part worth reviewing

`TRY_LOCK` is not free. It makes the whole descent non-blocking:

```c
if (tryFlag) { if (!try_lock_page(intCxt.blkno)) return OFindPageResultFailure; }
if (tryFlag && shmemIsReloaded)                  return OFindPageResultFailure;
if (tryFlag) { if (!o_btree_try_use_shmem(desc)) return OFindPageResultFailure; }
```

So merges become best-effort and will abandon under ordinary contention, not only when the tree is gone. That is safe — a merge not done is always allowed, and the function is already named `try_merge` — but it can leave pages sparse for longer. Whether those lost merges matter under real load is a judgement call the test suite cannot settle; if it does matter, the alternative is a narrower check (validate the tree is still alive after the window) instead of flipping the whole descent to non-blocking.

## Results

| | |
|---|---|
| `test_merge_into_concurrently_dropped_tree` | passes (was 3/3 crash) |
| `merge_test` | 14/14 |
| `regress` | 50/50 |
| `isolation` | 38/38 |

Also dropping the checkpoint question I raised earlier: that window is already closed, the checkpoint number is captured before the unlock.

akorotkov:
This reproduced **on its own** overnight, under plain load with no stop event — a TPC-C plus DDL-churn workload on `main` (`4f921c47`), assert build. Core saved.

```
#4  ExceptionalCondition ("tryFlag", src/btree/find.c:997)
#5  find_page (..., keyType=BTreeKeyPageHiKey, targetLevel=1)      find.c:997
#6  btree_try_merge_and_unlock (blkno=14818, nested=true)          merge.c:370
#7  walk_page (blkno=14818, evict=true)                            io.c:3234
#8  o_ppool_run_maintenance (pool=..., evict=true)                 page_pool.c:445
#9  ppool_run_maintenance
#10 o_ppool_reserve_pages (pool=..., kind=..., count=1)            page_pool.c:201
```

That is the path this PR is about, arrived at without any help: page-pool maintenance evicts, `walk_page` tries to merge a sparse page, `btree_try_merge_and_unlock` unlocks the target and re-descends for the parent, and the tree is gone by then.

Two details worth noting against the earlier description:

- the merging process here is an **ordinary backend** reserving pages (`o_ppool_reserve_pages`), not the background writer. Both reach `o_ppool_run_maintenance`, so either can be the one caught; the requirement is only that it holds no lock on the tree being dropped, which neither does.
- `nested=true`, so this came through the nested merge path.

So the stop event in this PR does not manufacture the window — it only makes an existing one hit on demand instead of after hours.

PR #1094, opened 2026-09-01
Contribute the primary index's own rows to a bridged bitmap

akorotkov:
Fixes the sqlsmith crash in [run 33550566039](https://github.com/orioledb/orioledb/actions/runs/33550566039/job/99998685837).

```
TRAP: failed Assert("false"), src/tableam/bitmap_scan.c:895
o_index_getbitmap <- exec_bitmap_index_state <- o_exec_bitmapqual
                  <- o_exec_bitmapqual <- o_make_bitmap_scan
Core was generated by `postgres: sqlsmith_fuzz postgres [local] DELETE'
```

## Cause

`o_index_getbitmap()` collects a bitmap index scan's rows two ways. Without a bridged index it fills the key bitmap, and it handles **both** index kinds there — a secondary tuple carries the pk as payload, a primary tuple is its own key. With a bridged index it fills a `TIDBitmap` instead, and there only the secondary kind was handled:

```c
else                                        /* tbm_result path */
{
    if (indexDescr->desc.type != oIndexPrimary) { ...fetch pk, tbm_add_tuples... }
    else
        Assert(false);                      /* bitmap_scan.c:895 */
}
```

The nesting in the backtrace is the case that gets there — `o_exec_bitmapqual` recursing means a bitmap qual with more than one arm. The bridged arm is what puts the whole scan on the `TIDBitmap` path, and an arm scanning the **primary** index then has to contribute to that same `TIDBitmap`. A row-array `IN` over the primary key is the ordinary way to produce such an arm, which is why the keybitmap side already carries a comment about exactly that.

## Impact without asserts

No crash — the branch falls through, `nTuples` is counted, and nothing is added to the `TIDBitmap`. The rows the primary arm found are **silently dropped from the result**.

## Fix

The tuple in hand is already the primary tuple, so take its bridge pointer straight from it instead of looking it up by its own key, as the secondary path must.

## Reproduction

```sql
CREATE TABLE t (a int PRIMARY KEY, b int, arr int[]) USING orioledb;
CREATE INDEX ON t USING gin (arr);
SET enable_seqscan = off;
SELECT count(*) FROM t WHERE arr @> ARRAY[3] OR a IN (...);
```

| | |
|---|---|
| `main` | aborts at bitmap_scan.c:895 |
| with the fix | bitmap answer matches the sequential one exactly — 4570 rows, same sum |

Test added to `index_bridging` (both expected files). `regress` 50/50, `isolation` 38/38.

## Noted while here, not fixed

With a **multi-column** primary key the same query instead trips `Assert("ix_descr->nPrimaryFields == 1")` at `bitmap_scan.c:230`, in `seconary_tuple_get_pk_data()`. That is the separate, documented limitation right above it ("bitmap scan works only for first field with int4, int8 or ctid type"), independent of this change.

### DIFF
--- a/src/tableam/bitmap_scan.c
+++ b/src/tableam/bitmap_scan.c
@@ -593,7 +593,6 @@ o_pk_encode_leaf(OTuple tuple, OIndexDescr *id, uint8 *out)
 {
 	bool		isPrimary = (id->desc.type == oIndexPrimary);
 	int			npk = isPrimary ? id->nKeyFields : id->nPrimaryFields;
-	int			pk_from = id->nFields - id->nPrimaryFields;
 	AttrNumber	attnums[INDEX_MAX_KEYS] = {0};
 	Oid			types[INDEX_MAX_KEYS] = {0};
 	int			i;
@@ -606,8 +605,14 @@ o_pk_encode_leaf(OTuple tuple, OIndexDescr *id, uint8 *out)
 			attnums[i] = OIndexKeyAttnumToTupleAttnum(BTreeKeyLeafTuple, id, i + 1);
 		else
 			attnums[i] = id->primaryFieldsAttnums[i];
-		types[i] = TupleDescAttr(id->leafTupdesc,
-								 isPrimary ? attnums[i] - 1 : pk_from + i)->atttypid;
+
+		/*
+		 * Read the type from the very attribute the value is fetched from
+		 * below.  For an ordinary secondary index pk_from + i names the same
+		 * slot, but the bridge index does not lay its pk out that way and the
+		 * two disagree, yielding InvalidOid.
+		 */
+		types[i] = TupleDescAttr(id->leafTupdesc, attnums[i] - 1)->atttypid;
 		len += o_pk_fixed_width(types[i]);
 	}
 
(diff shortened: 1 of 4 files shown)


akorotkov:
Fixed the composite-key limitation too (a07889f4), as asked. It was not one thing — **three** places had to agree before a composite key survived the trip, and I found them by walking the asserts one at a time.

## What the options were

`seconary_tuple_get_pk_data()` densifies the pk into a `uint64`, which can only hold one column. So for a composite key there were three ways out:

1. **Use the fixed-width encoding that already exists.** `o_keybitmap_pk_mode()` returns `O_KEYBITMAP_FIXED` for a composite of small ascending fixed-size columns, and `o_pk_encode_leaf()` / `o_keybitmap_insert_key()` are already the encoders for it. `o_index_getbitmap()` dispatches on this correctly; only the bridge path did not.
2. **Gate the bridged bitmap off** when the pk is not uint64-able. Correct but loses bitmap scans for those tables entirely, and the planner gate `amhasgetbitmap` is per-table, so it would take the non-bridged case down with it.
3. **Widen the densified key.** Not general — it fails again on the next type.

Took 1: the machinery was there, it just was not wired through.

## The three disagreements

- `bridge_next_page()` built its bitmap with `o_keybitmap_create()` unconditionally, while `o_exec_bitmapqual` and the dedup bitmap both choose by mode. Inserting a fixed key into a uint64 bitmap trips `Assert("bm->fixed")`.
- the bridge insert itself went through the uint64 helper unconditionally — `Assert("nPrimaryFields == 1")`.
- `o_pk_encode_leaf()` then rejected the bridge tuple with `unsupported fixed keybitmap type 4294967295`. It fetches each value by `primaryFieldsAttnums[i]` but read that value's **type** from a different slot, `pk_from + i`. Those name the same attribute in an ordinary secondary index, which is why it never showed; the bridge index does not lay its pk out that way and the lookup landed on `InvalidOid`. Now the type comes from the attribute the value comes from.

## Impact without asserts

Same shape as the first commit, worse consequence: the uint64 helper would take the **first pk column alone** as the whole key, so rows sharing that column collapse onto one bitmap entry and rows are silently dropped.

## Verification

| | rows | sum | distinct |
|---|---|---|---|
| bitmap, composite pk | 4569 | 454933836 | 4569 |
| sequential, same predicate | 4569 | 454933836 | 4569 |

Before: aborted. Both the single- and composite-key cases are now in `index_bridging`. `regress` 50/50, `isolation` 38/38.

The caveat I posted earlier is therefore withdrawn — the composite case is no longer a separate limitation, it is fixed here.

PR #1101, opened 2026-09-02
Disable binary table structure diagnostic

homper:
Reject calls with a feature-not-supported error, remove the unsafe traversal implementation, and update the table access method regression expectations.

### DIFF
--- a/src/tableam/func.c
+++ b/src/tableam/func.c
@@ -384,6 +384,7 @@ orioledb_tbl_structure(PG_FUNCTION_ARGS)
 	PG_RETURN_POINTER(result);
 }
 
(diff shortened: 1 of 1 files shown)


pashkinelfe (commented):
I think we can save orioledb_tbl_bin_structure under IS_DEV define and leave existing test expectations. In prod (without IS_DEV) it can output ERROR: no longer supported

homper:
Agree, updated

pashkinelfe (approved):
I'd place IS_DEV check (IS_DEV and non-IS_DEV variants) into orioledb_tbl_bin_structure function. But the current way is also ok

Issue #1113, opened 2026-09-04
S_lock stuck crash in bgwriter


pashkinelfe:
At dd1884809902032c09b98dd0f2f04720cd45a43f (arm64, 18, gcc, debug). Likely crash is pre-existing, not related to the commit mentioned.

https://github.com/orioledb/orioledb/actions/runs/33780949576/job/100734260457

akorotkov:
```
#0  __pthread_kill_implementation (threadid=280877223734432, signo=signo@entry=6, no_tid=no_tid@entry=0) at ./nptl/pthread_kill.c:44

Thread 1 (Thread 0xff74d3256ca0 (LWP 29295)):
#0  __pthread_kill_implementation (threadid=280877223734432, signo=signo@entry=6, no_tid=no_tid@entry=0) at ./nptl/pthread_kill.c:44
        tid = 29295
        ret = 0
        pd = 0xff74d3256ca0
        old_mask = {__val = {281474363775312}}
        ret = <optimized out>
#1  0x0000ff74d2a47540 in __pthread_kill_internal (signo=6, threadid=<optimized out>) at ./nptl/pthread_kill.c:78
No locals.
#2  0x0000ff74d29fcb3c in __GI_raise (sig=sig@entry=6) at ../sysdeps/posix/raise.c:26
        ret = <optimized out>
#3  0x0000ff74d29e7e00 in __GI_abort () at ./stdlib/abort.c:79
        save_stage = 1
        act = {__sigaction_handler = {sa_handler = 0x20, sa_sigaction = 0x20}, sa_mask = {__val = {188337135462532, 281474363775600, 280877215249888, 0, 280877216505856, 280877223734432, 188337139768424, 188337712416656, 280877215243600, 0, 0, 0, 16057815960925364224, 281474363775696, 188337135464144, 0}}, sa_flags = -1471128848, sa_restorer = 0xab4aa84d5318 <syslog_facility>}
#4  0x0000ab4aa80eaed4 in errfinish (filename=filename@entry=0xab4aa81f1978 "s_lock.c", lineno=lineno@entry=89, funcname=funcname@entry=0xab4aa8326ea0 <__func__.0> "s_lock_stuck") at elog.c:600
        edata = 0xab4aa8505c68 <errordata>
        elevel = <optimized out>
        oldcontext = <optimized out>
        econtext = <optimized out>
        __func__ = "errfinish"
[...]

PR #1114, opened 2026-09-04
 Detect stale seq_buf pointers after concurrent tree eviction 

pashkinelfe:
Detect stale seq_buf pointers after concurrent tree eviction
Release seq_buf spinlock before file I/O in seq_buf_finalize

Fix for #1113 

Needs thorough review

### DIFF
--- a/sql/orioledb--1.8--1.9_dev.sql
+++ b/sql/orioledb--1.8--1.9_dev.sql
@@ -22,3 +22,8 @@ CREATE FUNCTION orioledb_check_pending_truncates()
 RETURNS VOID
 AS 'MODULE_PATHNAME'
 VOLATILE LANGUAGE C;
+
+CREATE FUNCTION orioledb_test_free_meta_page(relid oid)
+RETURNS VOID
+AS 'MODULE_PATHNAME'
+VOLATILE LANGUAGE C;
--- a/sql/orioledb--1.8--1.9_dev.sql
+++ b/sql/orioledb--1.8--1.9_dev.sql
@@ -22,8 +22,3 @@ CREATE FUNCTION orioledb_check_pending_truncates()
 RETURNS VOID
 AS 'MODULE_PATHNAME'
 VOLATILE LANGUAGE C;
-
-CREATE FUNCTION orioledb_test_free_meta_page(relid oid)
-RETURNS VOID
-AS 'MODULE_PATHNAME'
-VOLATILE LANGUAGE C;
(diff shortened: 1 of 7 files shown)


pashkinelfe:
@akorotkov I've added test for stale ptr to PR

sql/orioledb--1.8--1.9_dev.sql
homper, line 1: Needs to use 1.9--1.10

pashkinelfe: @homper  Done by https://github.com/orioledb/orioledb/pull/1114/changes/c87bf6e5150194e6340036959c85863f3c6312e4
Also rebased onto fresh main branch

akorotkov:
The reproduction scenario isn't complete.
Concurrent eviction and write shouldn't be possible thanks to page_desc->ionum and the dirty flag.
I'm trying to reproduce this with additional debug instrumentation.

pashkinelfe:
> Concurrent eviction and write shouldn't be possible thanks to page_desc->ionum and the dirty flag. 

Probably condition in the test is still possible on a scenario like this:

```
A1  walk_page_prelock_check → index_oids_get_btree_descr
      resolves desc; desc->nextChkp[i].shared points into meta page memory

A2  reads meta_change_count from meta page

                              B1  get_evict_btree_locks [chkp-namespace AccessExclusiveLock]
                                  — doesn't conflict with A, A never takes this lock

                              B2  evict_btree: locks root page [root page lock],
                                  writes it, ppool_free_page(root) — frees root

                              B3  ppool_free_page(metaPageBlkno) [meta page lock]
                                  — increments meta change count, invalidates oids,
                                    meta page memory is now free for reuse

  A3  try_lock_page(leaf) [leaf page lock]
      — succeeds; leaf was never freed, it's a different page

  A4  walk_page_check_locked:
      — leaf oids valid? yes, leaf wasn't touched
      — IS_DIRTY(leaf)? yes, leaf is still dirty
      — leaf ionum < 0? yes, no IO on the leaf
      — all checks pass; all are leaf-specific, none examine the meta page

  A5  write_page → assign_io_num [leaf ionum]
      — protects against concurrent IO on this leaf, irrelevant for meta

  A6  perform_page_io → free_extent_for_checkpoint
      → seq_buf_write_file_extent(desc->nextChkp[chkp_index])
      → dereferences desc->nextChkp[].shared->lock, ->location, ->pages[]
      — this memory was freed at B3 → use-after-free / crash
```

pashkinelfe:
Fixed in another PR #1172  Closing current one

PR #1115, opened 2026-09-04
Validate serialized predicate terminators

homper:
Not sure about test using stopevents. Maybe it should've added some function like unittesting, but I wanted to test real code


akorotkov:
Reviewed, and pushed a commit closing the two things I found. Shout if you would rather I had left them as comments — happy for you to reshape or drop it.

**The diff as it stands is correct**, and I checked the parts that could have bitten: `o_serialize_string()` has written `strlen + 1` since the first commit, so the new terminator requirement cannot reject legitimate on-disk data; the empty-string case (`str_len == 1`) passes; `o_deserialize_retry_delay()` reproduces the old `Min(1000L << retry, ...)` values exactly for retry 0..7 while fixing real UB, since `O_DESERIALIZE_MAX_RETRIES` is 100 and the shift ran off the end of `long`. I also chased `next[str_len - 1]` for an underflow at `str_len == 0` — guarded by the `if (str_len != 0)`, no problem.

**1. `o_deserialize_node_safe()` still had both defects**, on the same blob a few bytes earlier. It never checked a terminator, and neither of its readers stops at `node_str_len`: `stringToNode()` parses until the string ends and `o_node_str_is_empty()` compares one, so a blob whose last byte is not `\0` is read past the end of the data. The bound check could not catch it either — it added a signed `ptrdiff_t` to an unsigned `node_str_len`, so a corrupted length near `SIZE_MAX` wrapped the sum and passed. Given the PR is specifically about this hazard, leaving the neighbouring function open seemed worth fixing here rather than later.

**2. The test could not reach any of the three length conditions.** The wrapper derives the prefix from the bytea and passes `serialized.len`, so the input is always exactly the right size and only the terminator check ever fires. The wrappers now take a `claimed_length`, which is what makes the truncation paths reachable; each condition gets a case.

The node wrapper takes a version stamp so a test can ask for the branch that does not parse — `stringToNode()` on deliberately malformed input would elog rather than return, while `o_node_str_is_empty()` reads to the terminator just the same and demonstrates an overread just as well.

Verified the other way round, since a test that passes either way proves nothing: with the old node code restored, the no-terminator case returns `true`. Locally `o_tables_test` 11/11 and regresscheck 52/52; pgindent and yapf clean.

One local snag worth knowing if you run this yourself: the `sql/%.sql` rule has no prerequisites, so `sql/orioledb--1.9--1.10.sql` is never regenerated once it exists. A stale one from a non-`IS_DEV` build makes the new function look missing. `rm` it and rebuild.

pashkinelfe (approved):
LGTM

PR #1117, opened 2026-09-04
Tests for builtin collations

homper:
Ok, enabled collation_test and added it inside, but could move it to another place if it will not be possible to fix disabled there test

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -187,6 +187,7 @@ ISOLATIONCHECKS = bitmap_hist_scan \
 				  seqscan_own_writes \
 				  uniq
 TESTGRESCHECKS_PART_1 = test/t/amcheck_test.py \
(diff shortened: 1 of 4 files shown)


homper:
Ok, redundant changes probably, they were working without them, will just reduce it to test then

mmlb:
Would be great to get this merged so we can rely on it for https://github.com/supabase/postgres/pull/2423

homper:
> Would be great to get this merged so we can rely on it for [supabase/postgres#2423](https://github.com/supabase/postgres/pull/2423)

Wait, is it failing for orioledb now on your side? Because this PR is purely tests and doc fixes

mmlb:
Nope wasn't failing or anything, just wanted this is so that we can be sure a regression won't happen.

PR #1122, opened 2026-09-07
Validate undo tuple sizes during historical scans

pashkinelfe:
Tuple-level undo records (BTreeModifyUndoStackItem) store an itemSize that is trusted without bounds checks when reading previous tuple versions. A corrupted itemSize causes an out-of-bounds undo_read into the caller's buffer (e.g. a btree page slot in apply_btree_modify_undo), or a uint16 underflow when itemSize < sizeof(BTreeModifyUndoStackItem).

Add validate_undo_item_size() in btree/undo.h that raises ERRCODE_DATA_CORRUPTED when itemSize is too small (underflow) or the resulting tuple size exceeds O_BTREE_MAX_TUPLE_SIZE. Replace the Assert-only guard in get_prev_leaf_header_and_tuple_from_undo with always-on validation, and add the same check in recovery.c where the identical pattern computes oldTupleSize from undo.

Add orioledb_test_corrupt_row_undo() (IS_DEV) and a regression test that corrupts itemSize to 0xFFFF and verifies the clean error.

Ref: ori-271, ori-272

### DIFF
--- a/sql/orioledb--1.9--1.10_dev.sql
+++ b/sql/orioledb--1.9--1.10_dev.sql
@@ -1,0 +1,9 @@
+/* contrib/orioledb/sql/orioledb--1.9--1.10_dev.sql */
+
+-- complain if script is sourced in psql, rather than via CREATE EXTENSION
+\echo Use "ALTER EXTENSION orioledb UPDATE TO '1.10'" to load this file. \quit
+
+CREATE FUNCTION orioledb_test_corrupt_row_undo(relid oid)
+RETURNS text
+AS 'MODULE_PATHNAME'
+VOLATILE LANGUAGE C;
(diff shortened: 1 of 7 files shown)


sql/orioledb--1.9--1.10_dev.sql
homper: Btw, what do you think, do we need to add all functions to previous dev file?
Or probably not, because it was before beta release?

pashkinelfe, line 1: I think better we bump sql version if _prod OR _dev functions change. Adding _dev functions to existing released sql version because they are invisible for prod users anyway still looks not so good and can lead to confusion IMO.

homper (approved):

PR #1126, opened 2026-09-07
Respect partial index predicates during upsert

homper:


pashkinelfe (approved):
@homper I've added extra tests to the branch

PR #1127, opened 2026-09-08
Fix partial indexes under ON CONFLICT and recovery, and primary key columns inside INCLUDE

JacobZwang:
Four fixes for partial indexes and for indexes whose `INCLUDE` list names a primary key column, found while running an upsert/delete-heavy workload on `beta17` (PostgreSQL 17.11, `patches17_21`). Each commit carries the full explanation; summary below. Draft because I would like a maintainer's view on the descriptor change in the third commit before this is considered final.

## 1. `INSERT … ON CONFLICT` adds non-qualifying rows to partial indexes

The non-conflict path of `INSERT … ON CONFLICT` writes the native secondary indexes itself instead of going through `ExecInsertIndexTuples()`, so nothing evaluated the partial-index predicate: every new row landed in every partial index on the table. Index scans then return rows that fail the query's own `WHERE`, and a partial `UNIQUE` index raises duplicate-key errors against rows it should never have compared. A plain `INSERT` is fine.

Fix: both index loops in `o_tbl_insert_with_arbiter()` skip a partial index whose predicate the new row fails, matching what `ExecCheckIndexConstraints()` does for heap arbiters. Regression cases added to `ioc`.

Minimal reproduction:

```sql
CREATE TABLE pi (id text PRIMARY KEY, frozen_at timestamptz) USING orioledb;
CREATE INDEX pi_open ON pi (id) WHERE frozen_at IS NULL;
INSERT INTO pi VALUES ('x', now()) ON CONFLICT (id) DO NOTHING;
SET enable_seqscan = off;
SELECT id FROM pi WHERE frozen_at IS NULL;   -- returns 'x'
```

## 2. Use-after-free in the index AM error path

`orioledb_amupdate()` and `orioledb_amdelete()` detoast the passed values into private copies, free them right after the tree operation, and then, if it failed, print them into the "unable to remove tuple from secondary index" message. Under load that path segfaulted in `text_to_cstring()` instead of raising the error. Fix: free after the report.

## 3. A primary key column inside `INCLUDE` is dropped from the index key

When a primary key column is already present as an `INCLUDE` column, `add_index_fields()` reuses that position instead of appending a second copy. `OIgnoreColumn()` ignores the whole `INCLUDE` range by position, so every comparator, sort and hash ignored the primary key too, and two rows that agree on the key columns compared equal:

* a non-unique index silently kept one entry for two rows (index scan: one row, seqscan: two),
* deleting or updating one of them found the other, or nothing ("unable to remove tuple from secondary index"),
* if the other row's transaction was still open, the insert waited on it and hashed the key through a NULL hash function (segfault in `wait_for_tuple` → `o_hash_key_from_tuple`).

```sql
CREATE TABLE t (id text PRIMARY KEY, grp text, ts timestamptz, ttl int) USING orioledb;
CREATE INDEX t_ix ON t (grp, ts) INCLUDE (id, ttl);
INSERT INTO t VALUES ('a', 'g', '2026-01-01', 1);
INSERT INTO t VALUES ('b', 'g', '2026-01-01', 1);
SET enable_seqscan = off;
SELECT id FROM t WHERE grp = 'g';   -- one row
```

Fix: the index descriptor flags fields that carry primary key columns (`OIndexField.primary`), `OIgnoreColumn()` never ignores such a field, and the descriptor fill gives it its opclass, comparator and hash function (the DDL side already copies those from the primary key for such `INCLUDE` columns). `hash_combine_mix_field()` also skips ignored columns and NULL hash functions so the hash covers exactly what the comparator compares. The on-disk tuple layout is unchanged; indexes built before this are missing entries and need `REINDEX`. New regression test `include_pk` and isolation spec `include_pk_wait` (the second insert must not block).

Open question for review: this keeps the single-copy layout and changes what counts as an ignored column. The alternative, appending the primary key as a separate trailing field, would change the layout for such indexes and needs a catalog version bump. I went with the smaller change.

## 4. Recovery ignores partial-index predicate flips when the key is unchanged

Both `UPDATE` replay paths (`apply_tbl_update()` in `worker.c` and the pending fix-up path in `recovery.c`) touched a secondary index only when its key changed. A row that leaves or enters a partial index through a non-key column (`deleted = true` under `… WHERE NOT deleted`) has an unchanged key, so recovery neither removed the pre-image entry nor added the post-image one. After any crash or immediate shutdown every such row was back in (or missing from) every partial index; the table and its partial indexes disagreed until `REINDEX`.

Fix: act when the predicate result differs between pre-image and post-image as well as when the key changes. New testgres test `partial_index_recovery_test.py` (four partial indexes, leave/re-enter through non-key columns including `ON CONFLICT DO UPDATE`, immediate stop, index-forced counts must match seqscan before and after).

## Testing

`regresscheck` (51/51) and `isolationcheck` (39/39) pass with the new tests; each new test fails on `beta17` without its fix (the recovery test also fails with only the first three fixes applied). A 250-connection upsert/delete load that segfaulted within seconds on `beta17` ran 120 s clean with all four, with every partial index agreeing with a sequential scan both after the load and after an immediate stop and restart.

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -118,6 +118,7 @@ REGRESSCHECKS = btree_sys_check \
 				generated \
 				getsomeattrs \
 				index_bridging \
+				include_pk \
 				indices \
 				indices_build \
 				inherits \
@@ -158,6 +159,7 @@ ISOLATIONCHECKS = bitmap_hist_scan \
 				  btree_scan \
 				  concurrent_update_delete \
 				  fkeys \
+				  include_pk_wait \
 				  included \
 				  insert_fails \
 				  ioc_deadlock \
--- a/Makefile
+++ b/Makefile
@@ -233,6 +233,7 @@ TESTGRESCHECKS_PART_2 = test/t/checkpoint_concurrent_test.py \
 						test/t/ddl_test.py \
 						test/t/eviction_full_memory_test.py \
 						test/t/include_indices_test.py \
+						test/t/partial_index_recovery_test.py \
 						test/t/indices_build_test.py \
 						test/t/logical_test.py \
 						test/t/logical_xid_leak_test.py \
--- a/include/tableam/descr.h
+++ b/include/tableam/descr.h
@@ -102,11 +102,11 @@ typedef struct
 	OHashFn    *hash_fn;
 
 	/*
-	 * True when this field carries one of the table's primary key columns.
-	 * A primary key column named in an INCLUDE list is stored once, in the
+	 * True when this field carries one of the table's primary key columns. A
+	 * primary key column named in an INCLUDE list is stored once, in the
 	 * INCLUDE position, but it still has to take part in key comparison and
-	 * hashing: without it two rows that agree on the key columns collide
-	 * (see OIgnoreColumn).
+	 * hashing: without it two rows that agree on the key columns collide (see
+	 * OIgnoreColumn).
 	 */
 	bool		primary;
 } OIndexField;
--- a/include/tableam/descr.h
+++ b/include/tableam/descr.h
@@ -100,6 +100,15 @@ typedef struct
 	OComparator *comparator;
 	OExclusionFn *exclusion_fn;
 	OHashFn    *hash_fn;
+
+	/*
+	 * True when this field carries one of the table's primary key columns.
+	 * A primary key column named in an INCLUDE list is stored once, in the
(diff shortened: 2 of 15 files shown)


JacobZwang:
I have not reviewed any of this code. I was benchmarking Oriole for my workload and it kept crashing under load, so I asked Claude to fix it. Maybe this is helpful.

— Actually Jacob

PR #1131, opened 2026-09-08
Include indices reorder

homper:
Store secondary-index fields in [key][PK][INCLUDE] physical order so
INCLUDE columns form a positional tail that comparison, sort, and hash
code can skip via OIgnoreColumn.  Add a persisted IndexAM attnum map
(indexAmAttnums) so INSERT/UPDATE/DELETE and index-only scans translate
between PostgreSQL's declared field order and the new physical layout,
replacing the old duplicate-compaction logic.  Bump
ORIOLEDB_SYS_TREE_VERSION to 2 and guard OIgnoreColumn with an upper
bound so toast/bridge sortstates whose keys extend past nFields are not
falsely collapsed as ignored INCLUDE columns.

Alternative approach to #1127

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -118,6 +118,7 @@ REGRESSCHECKS = btree_sys_check \
 				generated \
 				getsomeattrs \
 				index_bridging \
+				include_pk \
 				indices \
 				indices_build \
 				inherits \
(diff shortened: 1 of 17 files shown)


homper:
Ok, too big of a rewrite, going with #1127 approach

PR #1136, opened 2026-09-09
Bound compressed page metadata in read_page_from_disk

homper:
The downlink extent length and on-disk compress_page_size header field are attacker-controlled metadata. read_page_from_disk() used them to compute read_size and hand a size to o_decompress_page() without bounds checking: a length above ORIOLEDB_BLCKSZ / ORIOLEDB_COMP_BLCKSZ reads len * 512 bytes into the 8 KB stack buffer
OrioleDBChecksummablePage (stack smash), a zero length decompresses uninitialized data, and a crafted compress_page_size reads past the fetched extent.

Validate the extent offset/length before computing read_size and bound compress_page_size against the fetched bytes, raising ERRCODE_DATA_CORRUPTED (new OReadPageResultCorrupted) instead of crashing. Apply the same extent bound in s3_schedule_downlink_load(). Also fix read_disk_leaf_into_img() to check the read result before inspecting the (possibly garbage) page header and walking its undo chain. Add regression tests.

### DIFF
--- a/sql/orioledb--1.8--1.9_dev.sql
+++ b/sql/orioledb--1.8--1.9_dev.sql
@@ -18,6 +18,16 @@ RETURNS text
 AS 'MODULE_PATHNAME'
 VOLATILE LANGUAGE C;
 
+CREATE FUNCTION orioledb_test_corrupt_downlink_len(relid oid, new_len int4)
+RETURNS int8
+AS 'MODULE_PATHNAME'
+VOLATILE LANGUAGE C;
+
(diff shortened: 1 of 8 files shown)


sql/orioledb--1.8--1.9_dev.sql
homper, line 1: Needs to use 1.9--1.10

pashkinelfe:
@homper Provided we have page checksums always-on it looks unlikely forged on-disk page header can appear. The check looks unneeded for non-s3 case. I propose not to add this fix. I did the same in similar on-disk page-tampering reports like ORI-268 (approach can be discussed, anyway)

PR #1139, opened 2026-09-09
Tests and fixes for overlooked "front end" stuff

homper:
Creating this as draft PR, to remember to finish and merge it


akorotkov (commented):
Read through this one. Eight findings, two of them in the production code; the rest are tests that do not test what they say. Also one thing about the commit history worth settling before merge.

## Tests that cannot fail

**`test/t/index_bridging_test.py:434`** — an orphaned `[0]` on its own line:

```python
relnode_after_second = node.execute(...)
[0]
```

leaves `relnode_after_second` as the row tuple, so `assertNotEqual(<int>, <tuple>)` can never fail and the "re-add rewrote the heap" check is dead. (AST scan says it is the only occurrence of the pattern in this PR.)

**`test/t/recovery_test.py:3866`** — in `test_recovery_weird_types_comparator` the `checks` loop is tautological: every predicate matches all rows, and nothing pins the plan to the named secondary index, so both counts come from the PK either way. `orioledb_tbl_check()` does not close that gap — it runs `check_btree()` per tree and never cross-checks SK contents against the PK — so a diverged secondary index passes this test.

**`test/t/recovery_test.py:2958`** — a standalone `node.execute("SET enable_seqscan = off;")` is dropped with its connection, so the "PK index must be usable" assertions that follow run with seq scans enabled.

**`test/t/index_bridging_test.py:691`** — `gist_relnode` is captured and never used; the claimed "the dropped GiST index file is gone" check is a catalog lookup that duplicates `_idx_exists()` two lines above.

## A test that will fail, in one CI job

**`test/t/recovery_test.py:3580`** — `_tblspc_dir()` creates tablespace directories under `node.base_dir` without fsync. This file is the one the `dm_log_writes` job runs, where `replay_to_mark()` can discard the mkdir; `tblspc_redo` then FATALs with `directory "..." does not exist`.

## Production code

**`src/catalog/o_sys_cache.c:2292`** — `typentry->flags |= 0x080000` hardcodes the private `TCFLAGS_CHECKED_DOMAIN_CONSTRAINTS`. This PR already patches the fork's `typcache.h`, so export the macro there instead: if the bits are ever renumbered this sets a different flag with nothing to catch it.

**`src/catalog/o_sys_cache.c:2278`** — the hook is not a faithful stand-in for `load_domaintype_info()`. It skips the "release stale constraint info" block but still sets the CHECKED flag, so an entry that already has `domainData` gets re-marked up to date while keeping its old `DomainConstraintCache`. Narrow today — the hooks only install when `!IsTransactionState()` — but `domainData` should be nulled via `decr_dcc_refcount` before the flag goes on.

## .pgtags

**`.pgtags:1`** pins all three majors to raw SHAs on the unprotected `patchesNN_domain_info_hook` branches rather than to new `patchesNN_<n>` tags. They are children of the currently pinned tags, so nothing is lost as things stand — but if main's `.pgtags` advances first, merging this silently rewinds the fork. Worth a tag, or at least a note in the commit.

The fork hunk itself also places the hook call above the declaration block in `load_domaintype_info()`, which trips `-Wdeclaration-after-statement`.

## History

`4c16bd49b` "Fix partial-index predicate changes lost on UPDATE crash recovery" contains **no code change** — only tests. That fix reached main independently as `cd257f5d9` ("Maintain partial indexes in recovery when only the predicate flips"). As it stands the history claims a fix its diff does not contain; the message wants rewriting to say it is the test for a fix already in.

Minor: `import shutil` and `import tempfile` in `recovery_test.py` are unused.

## Checked and correct

For the record, these looked suspicious and are not: `%%` in the LIKE patterns, the multi-statement `SET ...; SELECT ...` idiom (one connection per `execute()`), the legacy positional `node.execute('postgres', sql)`, `QueryException` from `safe_psql`, the `Page \d+: level = 0` / `Index index_bridge` parsing in `_bridge_tree_page_count`, the `had %lld dead item identifiers removed` regex (one aggregate line in `vacuum.c:1480`), the plan shape `_used_index` assumes for a bridged GiST (plain `Index Scan`, cf. `test/expected/index_bridging.out:1506`), and the arithmetic behind every expected row count. The `tdtypeid = typentry->type_id` fix in `o_load_typcache_tupdesc_hook` is sound: the tupdesc is built fresh per call by `o_class_cache_search_tupdesc()`, the NULL-deref exposure is unchanged, and the assignment is what lets `record_cmp()` resolve the rowtype instead of failing on `RECORDOID`.

PR #1148, opened 2026-09-11
Turn some asserts into simple checks in WAL replay

pgnickb:
Add runtime checks for the cur_recovery_xid_state pointer and the sys tree number range, instead of relying on asserts. This lets non-assert builds detect malformed or out-of-order WAL records

instead of crashing.


src/recovery/recovery.c
```diff
@@ -5030,6 +5059,11 @@ replay_on_record(WalReaderState *r, WalRecord *rec)
 			}
 
 		case WAL_REC_SAVEPOINT:
+			if (cur_recovery_xid_state == NULL)
+			{
+				elog(LOG, "WAL record %s lacks XID state", wal_type_name(rec->type));
+				return WALPARSE_BAD_TYPE;
+			}
```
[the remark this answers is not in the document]
pgnickb: Null is an intentional case here when the primary opts to use fastpath (no WAL) for ROLLBACK

src/recovery/recovery.c
```diff
@@ -4791,7 +4791,12 @@ replay_on_record(WalReaderState *r, WalRecord *rec)
 				commit = (rec->type == WAL_REC_COMMIT);
 
 				Assert(rec->oxid != InvalidOXid);
-				Assert(cur_recovery_xid_state != NULL);
+
+				if (cur_recovery_xid_state == NULL)
+				{
+					elog(LOG, "WAL record %s lacks XID state", wal_type_name(rec->type));
```
za-arthur, line 4797: Why is it `LOG` level and not `WARNING`?

pgnickb: `WARNING` would imply that we'll continue anyway, but we actually willr eturn `WALPARSE_BAD_TYPE` and later an `ERROR` will be thrown. `LOG` makes more sense imo

src/recovery/recovery.c
```diff
@@ -4933,6 +4957,11 @@ replay_on_record(WalReaderState *r, WalRecord *rec)
 			break;
 
 		case WAL_REC_O_TABLES_META_LOCK:
+			if (cur_recovery_xid_state == NULL)
+			{
+				elog(LOG, "WAL record %s lacks XID state", wal_type_name(rec->type));
+				return WALPARSE_BAD_TYPE;
+			}
```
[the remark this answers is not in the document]
pgnickb: Doesn't make sense. This can't happen.

src/recovery/recovery.c
pgnickb: Resolving as per our offline discussion.

PR #1152, opened 2026-09-11
Validate numFreeBlocks against file size in sort_checkpoint_map_file

pashkinelfe:
A corrupt map file could make free_blocks_size exceed the actual file payload, causing OFileRead to return a short read or palloc to over-allocate. Check free_blocks_size against FileSize() before allocating. Add explicit (int) casts to match OFileRead's signature.

### DIFF
--- a/src/checkpoint/checkpoint.c
+++ b/src/checkpoint/checkpoint.c
@@ -2577,13 +2577,21 @@ sort_checkpoint_map_file(BTreeDescr *descr, int cur_chkp_index)
 		free_blocks_size = sizeof(uint32) * header.numFreeBlocks;
 	}
 
+	if (free_blocks_size > FileSize(file) - sizeof(header))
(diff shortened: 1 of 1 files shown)


src/checkpoint/checkpoint.c
```diff
@@ -2577,13 +2577,21 @@ sort_checkpoint_map_file(BTreeDescr *descr, int cur_chkp_index)
...
+							   "numFreeBlocks %lu exceeds file size",
+							   filename,
+							   (unsigned long) header.numFreeBlocks)));
+
 	free_blocks = palloc(free_blocks_size);
 	if (free_blocks_size > 0)
 	{
-		read_size = OFileRead(file, (Pointer) free_blocks, free_blocks_size,
+		read_size = OFileRead(file, (Pointer) free_blocks,
+							  (int) free_blocks_size,
```
homper, line 2591: Ok, was not sure about this casting, but it should be probably safe

homper (approved):

PR #1154, opened 2026-09-11
Add apt fallback for Ubuntu mirror outages

homper:
Canonical's Ubuntu mirrors can intermittently fail or hang on Blacksmith runners. Detect update failures and retry through azure.archive.ubuntu.com before installing CI prerequisites.


homper:
https://status.blacksmith.sh/cmtx0883205f40wn0gg834kz2
Ok, closing for now
And blacksmith said that it should be less possible in the future (I assume some caching or mirror change on their side or something)

PR #1157, opened 2026-09-11
fix: Add WAL_REC_TOAST_CHUNK record

za-arthur:
## Changes

When decoding DELETE and UPDATE with `REPLICA_IDENTITY_FULL` we need to have previous values of a tuple. It works for non-toasted values and it doesn't work for toasted values. To fix this issue we add new record type `WAL_REC_TOAST_CHUNK` which stores a TOAST chunk per record.

Changes:

* New `WAL_REC_TOAST_CHUNK` record and `WALRecToastChunk` struct
* `ORIOLEDB_WAL_VERSION` was bumped to 20

## Reproduction

This can be reproduced by the query:

```
create extension orioledb;
create table t (id int primary key, a text);
alter table t replica identity full;

-- md5 output is incompressible, so the value has to be stored out of line
insert into t select 1, string_agg(md5(random()::text), '') from generate_series(1, 128);

select 1 from pg_create_logical_replication_slot('s', 'test_decoding');

update t set a = 'new';

insert into t select 2, string_agg(md5(random()::text), '') from generate_series(1, 128);
delete from t where id = 2;

select regexp_replace(data, '''[0-9a-f]{64,}''', '''<3.9 kB value>''', 'g')
from pg_logical_slot_get_changes('s', null, null)
where data not like 'BEGIN%' and data not like 'COMMIT%';
```

Expected output:

```
table public.t: UPDATE: old-key: id[integer]:1 a[text]:'<3.9 kB value>' new-tuple: id[integer]:1 a[text]:'new'
table public.t: INSERT: id[integer]:2 a[text]:'<3.9 kB value>'
table public.t: DELETE: id[integer]:2 a[text]:'<3.9 kB value>'
```

But orioledb without fix returns:

```
table public.t: UPDATE: old-key: id[integer]:1 a[text]:unchanged-toast-datum new-tuple: id[integer]:1 a[text]:'new'
table public.t: INSERT: id[integer]:2 a[text]:'<3.9 kB value>'
table public.t: DELETE: id[integer]:2 a[text]:unchanged-toast-datum
```


akorotkov:
Opened #1162 as an alternative for the same bug, doing what heap does — `toast_flatten_tuple()` on the old tuple — instead of logging the removed chunks.

Two things I found while reviewing this PR that led me there, both reproducible:

1. `toastOldValues` is a process-global static reset on any row record with `ix_type != oIndexToast`. A value bigger than one `LOCAL_WAL_BUFFER_SIZE` spans containers, and containers of different backends interleave in the WAL stream, so a concurrent transaction's row record between the chunks and their row record wipes the accumulator. What survives starts mid-data, so `VARSIZE_ANY` reads a length out of arbitrary bytes — out of bounds in a non-cassert build, caught by the new `Assert(VARSIZE_ANY(old_value) == otv.toasted_size)` under cassert. Max chunk is about `O_BTREE_MAX_TUPLE_SIZE` (~2.6 kB), so values above roughly 5–7 kB are exposed; the 3.9 kB value in the test sits just below that.

2. An UPDATE that does **not** touch the TOASTed column still decodes as `unchanged-toast-datum`: `tts_orioledb_update_toast_values()` skips attributes where `o_toast_equal()` holds, so no chunks are removed and nothing is logged. Heap has no such gap, since `toast_flatten_tuple()` inlines every external attribute of the old tuple.

Where this PR is better: chunks have no size ceiling. The flattening approach is bounded by `OTupleHeaderData.len` being 15 bits — above 32767 bytes it falls back to the placeholder. Lifting that would need either a wider tuple header or a chunk-style path for the tail, so the two approaches may well end up combined.

Also worth knowing regardless of which way we go: checking the replay side turned up a pre-existing crash-recovery bug — a table with a varlena primary key and `REPLICA IDENTITY FULL` does not survive a crash, because `o_create_key_tuple()` memsets the destination while the key attributes still point into it. Confirmed on unpatched main; fixed in the second commit of #1162.

PR #1162, opened 2026-09-14
Log the old tuple flattened for REPLICA IDENTITY FULL (alternative to #1157)

akorotkov:
## Changes

An alternative to #1157 for the same bug: logical decoding of an UPDATE or
DELETE on a `REPLICA IDENTITY FULL` table reports `unchanged-toast-datum` for
every TOASTed attribute of the old tuple.

Instead of logging the removed TOAST chunks as a new record type, this does
what heap does: `ExtractReplicaIdentity()` runs `toast_flatten_tuple()` before
WAL-logging the old tuple, so the values travel inside it.

`o_tuple_flatten_toast()` builds that tuple the way `tts_orioledb_form_tuple()`
does — ctid and bridge attributes handed over separately, no `to_toast` map —
with the TOASTed attributes fetched as they are stored, still compressed if
they were compressed. It runs before the chunks are removed, and only for a
logically logged `REPLICA IDENTITY FULL` table that has something TOASTed;
nothing changes at `wal_level = replica`.

### Recovery

Replay never applies a flattened tuple, so no record type has to be taught to
skip anything:

- `WAL_REC_DELETE` and the old half of `WAL_REC_REINSERT` are reduced to a key
  in place before use;
- `WAL_REC_UPDATE`'s second tuple is ignored — it was already documented as
  being there only for logical decoding.

Only the **old** tuple is ever flattened; the new tuple, which is what replay
inserts, is untouched.

### Carrying a tuple with no useful bound

- `OTupleHeaderData.len` is 15 bits. A tuple that does not fit now sets it to
  `O_TUPLE_LEN_LONG` — the largest value the field can hold, so a short tuple
  can never be mistaken for a long one — and carries its real length in a
  `uint32` right after the header; the payload starts `SizeOfOTupleHeaderLong`
  in. The data offset became `OTupleHeaderDataOff()` everywhere it used to be
  a constant. No page-format change: everything stored in a B-tree is capped
  by `o_btree_check_size_of_tuple()` far below the threshold, so only a tuple
  built for WAL can take the long form.
- The length fields of the modify records grew from 16 to 32 bits.
  `ORIOLEDB_WAL_VERSION` is bumped to 20; the reader keeps parsing the narrow
  fields of older containers.
- A record the local WAL buffer cannot hold gets a container of its own: the
  buffer carries only its headers and the tuple bytes are registered straight
  into the XLog record, so a large old tuple is never copied. Anything already
  buffered is flushed first, so WAL order is unchanged.
- The reader copies a tuple that does not fit into the caller's `OFixedTuple`
  into a context of its own, reset per record.

Verified at 16 kB, 32740 B, 32800 B, 100 kB and 400 kB — the value comes
through at every size, instantly.

## Why not the chunk approach

Two findings from reviewing #1157, both measured:

1. **The decoder's accumulator is a process-global static**, reset on any row
   record with `ix_type != oIndexToast`. A value bigger than one
   `LOCAL_WAL_BUFFER_SIZE` spans containers, and containers of different
   backends interleave in the WAL stream, so a concurrent transaction's row
   record landing between the chunks and their row record wipes it. What
   survives starts mid-data, so `VARSIZE_ANY` reads a length out of arbitrary
   bytes — an out-of-bounds read that the new
   `Assert(VARSIZE_ANY == toasted_size)` only catches under cassert. Max chunk
   is about `O_BTREE_MAX_TUPLE_SIZE` (~2.6 kB), so values above roughly
   5–7 kB are exposed; the PR's own 3.9 kB test sits just below.

2. **An UPDATE that does not touch the TOASTed column still loses it.**
   `tts_orioledb_update_toast_values()` skips attributes where
   `o_toast_equal()` holds, so no chunks are removed, nothing is logged, and
   the old tuple still decodes as `unchanged-toast-datum`. Heap has no such
   gap. This branch handles it, because it flattens the whole old tuple.

## Two pre-existing bugs this uncovered

Both stand on their own and could be taken separately.

**A table with a varlena primary key and `REPLICA IDENTITY FULL` does not
survive a crash.** `o_create_key_tuple()` collects the key attributes with
`o_fastgetattr()` — for a varlena that is a pointer into the source tuple —
and then memsets the destination before `o_tuple_fill()` copies them. Replay
does exactly that when it rewrites the old tuple into a key in place:

```
TRAP: failed Assert("(data - tuple->data) == tuple_size")
      src/tuple/format.c, in replay_on_record
```

Reproduced with `(k text primary key, a text)`, a DELETE and `kill -9`, and
confirmed identical on unpatched main. By-value keys are unaffected, which is
why it has gone unnoticed.

**A logical subscriber of a table without a primary key never receives
TOASTed values.** `o_convert_toast_pointers()` keyed the reconstructed pointer
by an index into its own arrays, already shifted past the ctid, while the
chunks reach the reorder buffer keyed by the TOAST tree's attnum, which is
relative to the leaf tuple. They agree only when there is no ctid — that is,
only when the table has a primary key. Measured: `(id int primary key, junk
text)` decodes the value, `(id int, junk text)` reports
`unchanged-toast-datum` for the same INSERT.

This one is why `test_logical_subscription_toast_update_pkey` needed the fix
before it would pass here: it used to work by accident, because the old tuple
of the UPDATE was wrong in exactly the same way as what the subscriber had
stored, so the two mismatches cancelled out.

## Testing

- regress 52/52, isolation 39/39, testgres `logical_test` 50/50
- crash recovery with a varlena PK and a 120 kB TOASTed column: md5 of every
  row identical across `kill -9`
- new cases in `test/sql/toast.sql`: the reproduction from #1157, an UPDATE
  that leaves the TOASTed attribute alone, a table with no primary key
  (`primaryIsCtid`), a value past the local WAL buffer, and a value past the
  15-bit tuple length


pashkinelfe (commented):
While flattening tuple we currently need 2x tuple size memory to store all toast chunks and allocated flattened tuple. This is the same in PG, however. I propose to save on the unneeded 2x allocation, which will make Oriole toast flattening better that PG. @akorotkov please look at the branch https://github.com/orioledb/orioledb/tree/toast-flatten-save-memory and merge it to the current PR if you agree.

akorotkov:
@pashkinelfe The idea is right and I want it — dropping the second copy makes flattening strictly better than heap's `toast_flatten_tuple()`, which does hold both. But the branch as it stands has a memory-safety bug, so I have not merged it as-is; I pushed it with a fix on `akorotkov/toast-flatten-save-memory` (`5662efe5`) for you to look at.

**The bug.** `o_tuple_fill_ex()` takes the four-byte-header branch for every skipped attribute, but `o_new_tuple_size()` gets no `skip_copy` to consult, and it reserves a *short* varlena for anything small enough to pack:

```c
if (ATT_IS_PACKABLE(atti) && VARATT_CAN_MAKE_SHORT(DatumGetPointer(val)))
    data_length += VARATT_CONVERTED_SHORT_SIZE(DatumGetPointer(val));   /* no alignment */
```

`VARATT_CAN_MAKE_SHORT` looks at nothing but VARSIZE, and the cookie's VARSIZE is the real detoasted size — so for any toasted attribute of 130 bytes or less the sizer reserves the short form while the fill writes the long one. That is 3 bytes plus up to 3 of alignment past the reservation *per attribute*, every later attribute at a wrong offset, and `attr_offsets` pointing outside the allocation for the detoast pass to `memcpy` into.

**It is reachable.** The loop in `tts_orioledb_toast_values()` toasts the largest remaining attribute until the tuple fits, with no lower bound on size, and `STORAGE EXTERNAL` skips the compressor entirely. Two dozen 120-byte external attributes do it. A/B on the same script:

| build | result |
|---|---|
| `akorotkov/flat-old-tuple` | UPDATE succeeds, 0 crashes |
| `+ toast-flatten-save-memory` | `TRAP: failed Assert("(data - tuple->data) == tuple_size"), format.c` |
| `+ the fix` | UPDATE succeeds, regress 52/52 |

Only cassert builds get the assert; `normal` and production get a silent heap overrun.

**The fix** keeps the cookie where it pays and drops it where it does not: a value of 130 bytes or less is detoasted up front as before, since holding something that small alongside the result costs nothing and the whole point of the cookie is the large ones. That also keeps the flattened bytes identical to the current PR rather than switching small attributes to a long header. `o_tuple_fill_ex()` additionally asserts that a skipped attribute is not one `o_new_tuple_size()` would have packed, so the next caller of this API gets told instead of overrunning.

I added the 24-attribute case to `test/sql/toast.sql`. It only does INSERT/UPDATE/DELETE and never consumes the slot, for the reason below.

**Separate, pre-existing, and worth its own issue.** While writing that test I found that *decoding* such a row crashes in PostgreSQL's toast reassembly:

```
TRAP: failed Assert("data_done == VARATT_EXTERNAL_GET_EXTSIZE(toast_pointer)"), reorderbuffer.c:5176
```

This is **not** yours and not this PR's — I checked it on `main` (`691b1389`) with the same table and it crashes there too, so it predates both. That is why the new test stops short of `pg_logical_slot_get_changes`; extending it to decode is blocked on that bug.

akorotkov:
Merged into the PR — the branch now fast-forwards to `5662efe5`, three commits on top of `77afb56c`:

```
5662efe5 Alexander Korotkov  Keep sizing and filling in step for a skipped attribute
15b03df4 Pavel Borisov       Halve peak memory of o_tuple_flatten_toast
7e560a1d Pavel Borisov       Add o_tuple_fill_ex with skip_copy and attr_offsets support
```

Kept @pashkinelfe's two commits as they are rather than squashing, so the authorship stays with him and the fix is separately reviewable. The tree is byte-identical to what I tested above: regress 52/52, and the new 24-attribute case in `test/sql/toast.sql` fails on `15b03df4` alone and passes with `5662efe5`.

pashkinelfe:
@akorotkov you probably also need latest commit https://github.com/orioledb/orioledb/pull/1188/changes/a9faab7235d3ea85fe57987428a10af290df7f1c from the branch mentioned previoulsy (silence static warnings). I'm looking at your fixes, thanks!
(Draft PR https://github.com/orioledb/orioledb/pull/1188 for the mentioned branch was just for running CI, not for separate merge)

pashkinelfe:
@akorotkov the fix 5662efe5 looks good to me. A couple of cosmetic things for both yours and my commits:

This block can look better as `if/else` block instead of using `continue`:
```
  if (VARATT_CAN_MAKE_SHORT(cookie))
 355         {
 356             pfree(cookie);
 357             values[attn] = PointerGetDatum(o_detoast((struct varlena *)
 358                                                      DatumGetPointer(values[attn])));
 359             eager_atts[neager++] = attn;
 360             continue;
 361         }
 362
 363         values[attn] = PointerGetDatum(cookie);
 364         skip_copy[attn] = true;
 365         toasted_atts[ntoasted++] = attn;
```

I'm also not sure about semantic clarity of `cookie/eager` names. Maybe, rename them if you prefer.

Overall PR looks good to me with https://github.com/orioledb/orioledb/pull/1162#issuecomment-5714668094 above.

pashkinelfe (approved):

akorotkov:
@pashkinelfe all three done, pushed as `f01da87f`:

```
f01da87f Pavel Borisov       Silence static analyzer false positive in o_create_key_tuple
d6f872c5 Alexander Korotkov  Keep sizing and filling in step for a skipped attribute
15b03df4 Pavel Borisov       Halve peak memory of o_tuple_flatten_toast
7e560a1d Pavel Borisov       Add o_tuple_fill_ex with skip_copy and attr_offsets support
```

**`a9faab72` picked up** — and it was needed: `static` was red on `5662efe5` in all three clang cells with exactly the warning your commit describes.

```
src/tableam/tree.c:232:3: warning: Null pointer passed to 1st parameter expecting 'nonnull' [core.NonNullParamChecker]
```

I checked the `Assert` really is what silences it rather than assume so — locally, with it: no `core.*` findings at all; without it: `tree.c:232` flagged again. (Local clang attributes it to `unix.cstring.NullArg` instead of `core.NonNullParamChecker` — same defect, different checker, so the local reproduction is only near-identical to CI's.) Cherry-picked with your authorship intact.

Incidentally the reason this PR gets a `static` run at all is #1186, which removed the fork-only gate; before that, a same-repo PR branch got no static run on either trigger.

**`continue` → `if/else`** — agreed, it reads better now that both arms are doing comparable work.

**Names** — you were right that they were poor. `cookie` said nothing about what it holds, and `eager` was the odd one out against `toasted` when *both* lists hold TOASTed attributes. Now they are named for what happens to the value:

| was | is |
|---|---|
| `cookie` | `size_stub` |
| `toasted_atts` / `ntoasted` | `deferred_atts` / `ndeferred` |
| `eager_atts` / `neager` | `upfront_atts` / `nupfront` |

That renames variables your commit introduced, so it sits in my commit rather than being squashed into yours.

regress 52/52 and pgindent clean on the final tree.

pashkinelfe:
Other ideas of naming (arguable):  
deferred_atts -> toasted_deferred_atts **or** toasted_long_atts
upfront_atts -> toasted_upfront_atts **or** toasted_short_atts

pashkinelfe (approved):

akorotkov:
@pashkinelfe Done — `ae548ba9`. Took your first variant:

| was | is |
|---|---|
| `deferred_atts` | `toasted_deferred_atts` |
| `upfront_atts` | `toasted_upfront_atts` |

Your reasoning is the right one and I had it backwards in my last commit message: I justified the names by "both lists hold TOASTed attributes, so name them for what happens to the value", which explains dropping `toasted` from the *distinguishing* part but not from the names entirely. Upfront-vs-deferred is only a thing inside this one-by-one flattening, so the prefix is what says which attributes are even under discussion. Commit message says that now.

On `toasted_long_atts`/`toasted_short_atts`: I considered it and it does name the actual criterion — the split is literally `VARATT_CAN_MAKE_SHORT(size_stub)`, and short-vs-long is precisely what the bug was about. I went with upfront/deferred because that is what the two loops below do with each list, and the criterion is already spelled out in the comment at the branch. Happy to switch if you prefer the criterion in the name.

Diff against `f01da87f` is the rename and nothing else — 11 lines, same tree otherwise.

CI on `f01da87f` came back green across the board, including the one that mattered:

| workflow | result |
|---|---|
| `static` | success (was failure before your `a9faab72`) |
| `check` | success |
| `sqlsmith` | success |
| `pg_upgrade` | success |

`ae548ba9` is queued; regress 52/52 and pgindent clean locally.

PR #1164, opened 2026-09-15
Implement the heuristic to decrease merge attempts

bwyltcaxe:
During `ROLLBACK` the undo callback is trying to merge the right leaf (because the UNDO rollback happens at previous-to-change order).  Hence, when it goes through the threshold of is_page_too_sparse() it tries to merge it with sibling (left page) for every row at right page, so that's why we see too many useless-tries when the left page is not changed. The perf probes results:

         3,049,759      probe_orioledb:merge_attempt
           169,940      probe_orioledb:merge_pair
            47,396      probe_orioledb:merge_done

The successfull rate is ~1.5%.  This patch introduces the kind-of-cache approach, on first unsuccessfull merge we keep the watermark:
	watermark = vac(left) + free(left).
This is the threshold until merge (probably) will fail and we won't try to try it again then.

This heuristic may skip the merge of sparsed page if concurrent backend compacted its sibling.


pashkinelfe:
@bwyltcaxe it's interesting optimization, thank you! I tagged it as a performance feature along with some others reflecting decision of OrioleDB team to include only bugfixes into upcoming beta18. After that we'll return to numerous performance features including yours.

PR #1167, opened 2026-09-15
Refuse a page image whose layout cannot fit the page

akorotkov:
Fixes the reported B-tree high-key stack buffer overflow
(`src/btree/merge.c:760-762`, via `page_get_hikey_size()`).

Everything that touches a page image computes offsets and lengths out of the
layout fields in its header and then trusts the result.
`page_get_hikey_size()` is the sharpest case: it returns

```c
hikeysEnd - SHORT_GET_LOCATION(chunkDesc->hikeyShortLocation)
```

and `merge_pages()` copies that many bytes into

```c
char newItem[Max(BTreeLeafTuphdrSize, BTreeNonLeafTuphdrSize) +
             O_BTREE_MAX_TUPLE_SIZE];
```

on the stack. `hikeysEnd` is a plain `uint16`, so an image claiming 65535 there
asks for a ~64 kB copy into a ~2.7 kB buffer, out of a source that runs past
the page as well.

Those bytes are not ours to trust: a page image comes from a file that may have
been damaged, and in S3 mode from an endpoint that answers with whatever it
likes. The checksum does not help — it proves only that these are the bytes
that were stored, and anyone who can choose the bytes can compute the checksum
too. Nothing between the read and the copy establishes that the header
describes a page at all: the structural rules exist only as `Assert()`s in
`o_check_page_struct()`, which is behind `CHECK_PAGE_STRUCT` and compiled out
of exactly the builds that would face a hostile store.

## The fix

**Check the layout where the untrusted bytes enter.** `page_struct_is_valid()`
validates the header and the chunk descriptors — `chunksCount`, `itemsCount`,
`dataSize`, `hikeysEnd`, and per chunk its data and hikey locations, their
monotonicity, and that each hikey is a size a tuple can be —  and
`read_page_from_disk()` returns a new `OReadPageResultStructureInvalid` when
they do not hold. The three callers already turn a bad read into an `ERROR`
naming the file and offset; the specifics go into a `WARNING` first, the way a
checksum mismatch does.

That is the header plus at most `BTREE_PAGE_MAX_CHUNKS` descriptors — a few
dozen comparisons against an 8 kB read — and it covers every consumer of a
loaded page rather than the one sink that was reported (split, iterators and
the rest read hikeys too).

`can_be_merged()` additionally refuses a page whose hikey does not fit a tuple.
With the above, such a page cannot arrive from storage any more, so that guard
is about one that went wrong while in memory; it is the last point before the
caller's critical section, where an error would be a PANIC, so it declines the
merge rather than raising one. **Nothing in the test suite reaches it.**

## Test

`test/t/page_struct_validation_test.py` forges a page the way the reported
attack does rather than merely corrupting one: it patches `hikeysEnd` and then
**recomputes the page checksum**, so the image is accepted as intact and is
refused for its structure alone.

It also recomputes the checksum of the untouched page first and requires it to
match what the server stored — if that self-check ever fails, the field offsets
or the algorithm have moved and the forgery would be testing nothing.

Without the fix the scan reads the forged page without complaint; with it the
read fails, the WARNING names `hikeysEnd`, and the cluster stays up.

Checked: regress 52/52, isolation 39/39, eviction/amcheck/merge/page_pool
testgres suites (23 tests), and a non-`IS_DEV` `-Werror` build. The amcheck
page-corruption tests still report a checksum mismatch, since this check runs
after the checksum.


akorotkov:
## Benchmarks

Taken on the bench VM (64-core arm64, 125 GB RAM, NVMe RAID) against PG 18.6
built `--enable-debug --disable-cassert CFLAGS=-O3` at exactly main's pinned
`patches18_2` — a cassert build would have inflated everything and told us
nothing. Arms alternate (A, B, A, B) with `make clean` per arm, since the fix
touches two headers; `.so` md5 logged per arm.

**Workload chosen to be the worst case for a per-page cost:** a 2298 MB table
(294 119 pages, 9 chunks each) with a 64 MB pool, so every full scan re-reads
every page through `read_page_from_disk()` — ~600k page reads/s, all served
from the OS page cache. A read that cost a device seek would hide the check.

### Where the cost actually is (`perf record`, flat, warm scan)

| symbol | share |
|---|---|
| `__arch_copy_to_user` (the page read) | 13.77% |
| `oriole_checksum_block` (per-page checksum, already there) | 6.66% |
| **`page_struct_is_valid` (this PR)** | **0.58%** |

The new check is ~1/11 of the checksum this code already pays on every page
read, and ~1/24 of the copy.

### Instruction counts (`perf stat -a`, 5 scans per sample, 6 samples per arm)

Wall clock on this box has a ~2.5% per-run spread, which cannot resolve tens of
nanoseconds per page; instruction counts have a ~0.1% spread and can.

| | instructions (median) | Δ per page read | Δ of scan |
|---|---|---|---|
| A: main | 37.099e9 | — | — |
| B: first version of this PR | 37.611e9 | +348 | +1.38% |
| B: after the one-pass rewrite | 37.501e9 | **+274** | **+1.08%** |

The rewrite (second commit) came out of this measurement: the first version
compared each chunk with its neighbours by fetching those neighbours' bitfields
again. ~10 ns per page read at this box's IPC.

### End to end

| | A: main | B: fix | delta |
|---|---|---|---|
| warm scan, page cache (n=14 each) | 507.0 ms mean | 514.7 ms | **+1.5%**, 95% CI −0.2%..+3.2% |
| cold scan, `drop_caches`, real reads (n=6) | 1132.7 ms median | 1142.4 ms | +0.86%, inside a 3% spread |
| point lookups, 8 clients (n=6) | 127 460 tps median | 140 118 tps | B *faster* — the 11-14% spread swamps any cost |

So: ~0.6% of cycles and ~1% of instructions on a scan that does nothing but
read pages out of RAM as fast as it can; not distinguishable from zero once the
reads are real, which is the case that matters for a cluster whose data does not
fit in memory — and doubly so for the S3-backed deployments this issue is about,
where a page read crosses a network.

pashkinelfe (commented):
I think page checksums that are always-on are adequate safeguard against page content corruption. Extra checking page contents is compromising performance in prod builds. @akorotkov I propose to close this PR (like #1174) without merging.

akorotkov:
@pashkinelfe, yes, won't fix.

PR #1168, opened 2026-09-15
Detect corrupted undo chain links

pgnickb:
Undo chain links only ever point to a lower location (the offset monotonically
increases as new undo is written). Disk corruption or a bug could turn a link
into a self-reference or a cycle, and every place that follows undo chains
would then loop forever: startup recovery, MVCC tuple-version lookups, and
page-image reconstruction for historical scans.

Check that the next link, when valid, is strictly less than the current
location, and PANIC otherwise.


src/btree/undo.c
```diff
@@ -2232,14 +2259,22 @@ get_prev_leaf_header_from_undo_if_exists(UndoLogType undoType,
...
 
-	if (!undo_read_if_exists(undoType, tuphdr->undoLocation,
+	if (!undo_read_if_exists(undoType, location,
 							 sizeof(prevTuphdr), (Pointer) &prevTuphdr))
 		return false;
 
+	/* Tuple undo chains only ever move backwards; a cycle means corruption. */
+	if (UndoLocationIsValid(prevTuphdr.undoLocation) &&
+		(UndoLocation) prevTuphdr.undoLocation >= location)
+		elog(PANIC,
```
pashkinelfe, line 2274: Can undo_read_if_exists return true but stale data because of race with concurrent undo truncation? In this case PANIC would crash the entire cluster. Looks like return false is better in this case.

pashkinelfe (changes requested):
See added comment. Also do all the changes address only Startup Recovery (as in PR header?) Is commit message relevant?

src/btree/undo.c
pgnickb: I don't think so:
`undo_read_if_exists` calls `UNDO_REC_EXISTS` check before and after the read returning false if the undo rec doesn't exist. So if `undo_read_if_exists` returned `true` we're definitely dealing with a valid data and hence `PANIC` is warranted.

pashkinelfe: What if 
1. undo_read_if_exists reads prevTuphdr from location, post-read check passes, returns true
  2. Concurrent cleaner advances retain floor past location, slot is recycled
  3. A new undo record is written into the same circular buffer slot
  4. The prevTuphdr we hold now has bytes from the new record. prevTuphdr.undoLocation happens to be >= location — PANIC

pashkinelfe: I think the scenario above can't appear. Resolving thread

pashkinelfe (approved):

src/btree/undo.c
pgnickb: The version we hold is memcpy'd so this shouldn't happen.

PR #1170, opened 2026-09-15
Arm merge stopevent after warming the buffer pool

pashkinelfe:
The warmer's reads require eviction in the 8MB pool, which can trigger a merge that parks on the already-armed stopevent. The warmer backend blocks forever, the presser never starts, test hangs.

Risk: under Valgrind the bgwriter may re-evict warmed pages before the presser starts, making the test fail with "nothing parked".


homper:
I am not sure about merging it, because I assumed that it will have same issue that it have before https://github.com/orioledb/orioledb/pull/1134
And as I understand first assumption of my Navier-Stokes solver was correct:
```
Yes, for the code used by that CI job.

src/orioledb.c:872 defines the GUC with:

\```c
&bgwriter_num_workers,
1,              /* default */
1,              /* minimum */
MAX_BACKENDS,
\```

Therefore 0 is invalid. PostgreSQL ignores the invalid configuration value and retains t
he default 1.

The failed job confirms the runtime behavior:

\```text
WARNING: 0 is outside the valid range for parameter
         "orioledb.bgwriter_num_workers" (1 .. 262143)
LOG: orioledb background writer 0 started
\```

So orioledb.bgwriter_num_workers = 0 definitely did not disable the bgwriter in that run
.

🔢 275983 tokens used, cost $0.000000  5h 81% left

------------------------------------------------
```
Because test was probably expecting that bgwriter is disabled and page eviction should be performed by backend and catched by stopevent
So solution could be just setting a correct option and reverting both of those PRs
Could it be?

PR #1171, opened 2026-09-15
Stabilize merge drop test under Valgrind

homper:
Rework the warming workaround added by #1134: warm candidate pages before arming the stopevent so warming cannot park itself. Also use orioledb.debug_disable_bgwriter because bgwriter_num_workers has a minimum of one, so setting it to zero left the worker active.

Supersedes #1170, which reorders warming but does not fix the bgwriter option.


pashkinelfe (approved):

homper:
Ok, it decided to go with original solution in the end (after spending my 5h quota), waiting for CI

---

> 10000 agents

pashkinelfe:
Looks like the current state can fail merge drop test similar to pre - 0eabb10afa93d1b1 state (https://github.com/orioledb/orioledb/actions/runs/35001402639/job/104490531550?pr=1171) Maybe the simple way to fix it is modifying the test to pass under this condition under valgrind only.

akorotkov:
The `bgwriter_num_workers` minimum is a good catch — I had not noticed, and it explains something concrete: in [run 35085730232](https://github.com/orioledb/orioledb/actions/runs/35085730232) the `normal` cell hung for its whole 30-minute budget with the bgwriter parked in this very window, on a test that believed it had switched the worker off.

Measured on a live cluster, for the record:

```
orioledb.bgwriter_num_workers = 0   ->  WARNING: 0 is outside the valid range (1 .. 262143)
                                        1 orioledb background writer in pg_stat_activity
orioledb.debug_disable_bgwriter = true  ->  none
```

One thing that stack dump also shows is that reordering the warm-up is not quite the whole of it: the same window had the *checkpointer* parked, and it had the warm-up backend parked because it ran after the arming — but the `DROP TABLE` connection runs page pool maintenance too, and it is the step that releases the event, so `'true'` leaves a way for the test to park the one session it cannot afford to.

I have put both halves together in #1180 — your `debug_disable_bgwriter` plus arming the event only for what the test is not driving, by `application_name` — with the stacks quoted. Happy either way: land yours and I will rebase mine down to the scoping part, or take #1180 and close this one.

PR #1172, opened 2026-09-16
Stop a page write from using a meta page the tree no longer owns

### LINKED ISSUE #1113: S_lock stuck crash in bgwriter

akorotkov:
Fixes #1113.

## The failure

```
PANIC: stuck spinlock detected at perform_page_io, src/btree/io.c
```

47 seconds after the write began, in the orioledb bgwriter.

A page write reaches the seq bufs through `desc->nextChkp[]` and `desc->tmpBuf[]`, which live in the tree's meta page. The descriptor that names that page is cached per process, and nothing keeps the tree loaded for the page pool's clock sweep: the meta page can go back to the pool and be handed to another tree, whose data then reads as a seq buf — including a "spinlock" word that is really a block number and so never becomes free.

## Why every existing guard passes

Readers are protected from the same staleness by the root page change count, which `find_page()` compares on every descent, and a backend also has its cached descriptor dropped by the invalidation `walk_page_evict_root()` sends. The clock sweep has neither: it reaches a page directly and matches it to a descriptor by the page's oids, which are *identical across incarnations of the tree*. So the bug needs the one shape in which all of the checks pass — the tree reloaded into its old root block number while its meta block moved on — and a LIFO reload hands back the same root often enough.

## The fix

Give the meta page the guard the root page already has: `BTreeRootInfo` records the meta page's change count, taken where the page is allocated. `o_ppool_free_page()` bumps a page's change count under the page lock before handing it out, and that is the only place a meta page's count ever changes, so `BTREE_META_PAGE_IS_OURS()` is one shared-memory read and an integer compare — no lookup, cheap enough to sit in front of every write.

* `walk_page()` asks with the page locked, which holds for the rest of the walk: a concurrent eviction of the tree would have to take that same page, and the in-flight IO the write leaves behind keeps it out just as well.
* `evict_btree()` asks again, because `walk_page_evict_root()` has to release the lock in between to take the evict locks in order.
* A descriptor that fails the check also forgets its block numbers, mirroring the existing repair idiom in `find.c` and `write_tree_pages()`. Without that, the tree's dirty pages would be skipped for ever.

Two designs were tried and dropped, and the reasons may be worth keeping in mind:

1. Revalidating the shared root info per walked page is a `SYS_TREES_SHARED_ROOT_INFO` descent plus a palloc for every page the sweep touches, and a check made before the page lock can go stale before the write.
2. Stamping the tree's identity onto the meta page's `OrioleDBPageDesc` would have been a regression: `init_meta_page()` deliberately leaves those oids invalid, and that invalidity is the only thing that keeps the clock sweep from calling `walk_page()` on a meta page at all.

## Test

`test/t/evict_stale_descr_test.py` builds the interleaving rather than waiting for it, with two new stop events:

* `after_tree_evict` freezes the pool right after an eviction, so the blocks the tree gave up are the ones its next incarnation is handed;
* `after_tree_root_page_alloc` parks the reload between the two `ppool_alloc_page()` calls in `init_shared_root_info()`, where a third backend takes the block the meta page was about to get.

The writer is the bgwriter, as in the report, and the test retries the shape until the reload comes up on the root block the bgwriter's descriptor names (checked through `orioledb_table_pages()`), under a wall-clock budget.

One honest limit: a detect-only build — log the mismatch, let the write proceed — passes the same test, because `perform_page_io()` on a page that already has a file extent need not touch the seq bufs. The test asserts that the guard fires, not that the unguarded write corrupts.

## Checks

regress 52/52 · isolation 39/39 · eviction, eviction_bgwriter, eviction_compression, checkpoint_eviction, eviction_full_memory, merge suites all green · zero skip messages on legitimate traffic · `-Werror` non-cassert build clean.


akorotkov:
Pushed a rework of the test after the `check (amd64, 16, clang, check_page)` failure.

**What happened.** All ten attempts died at the same step: the bgwriter never wrote the victim's page, so the descriptor it was supposed to keep was never cached and the attempt was spent without ever reaching the interesting part. The run took 203.58 s, which is exactly ten times the 20 s I gave that step — the give-away. The `... FAIL` line for `test_replication_multiple_checkpoints_rollback` in the same log is interleaved output from a parallel `make` job; `replication_test` itself reported OK for all 45 tests, and `evict_stale_descr_test.py` was the only `Error 1`.

**Why that step is fragile.** It needs two transient things at once: the pool in the bgwriter's write mode (more than half dirty, a twentieth free) and a dirty victim page still resident. The second is easy to lose — a backend's own sweep can take that page before the bgwriter's next wakeup. Setting both up once and then waiting long is the wrong shape; the test now re-establishes both and gives the bgwriter a short window each time, and does the same in the exploit phase.

**And the counts are now reported**, so a future failure says which step did not come up:

```
# attempts: {'no_prime': 0, 'no_evict': 0, 'other_root': 0, 'built': 1}
```

**One deliberate change of verdict.** A run that never sees the guard's report now skips instead of failing. I tried to make the negative case an assertion — park the bgwriter at `after_ionum_set`, which is inside `write_page()` and therefore past the guard, and call a park with nothing reported an unguarded write — and it gave false failures 2 runs in 4: after the guard fires it repairs the descriptor, and the bgwriter also reaches that page legitimately whenever its descriptor was refreshed for any other reason. So the absence of the report is not evidence of anything, and the test says so rather than pretending otherwise. The report itself is durable and unambiguous, and that is what it asserts.

Measured after the rework: 8 runs, all green — 5 on the ordinary build (4.9–5.3 s) and 3 with `-DCHECK_PAGE_STRUCT -DCHECK_PAGE_STATS`, the configuration that failed (5.5 s, and 80 s with five attempts, which is the retry loop working). Worst case is bounded by a 180 s budget, after which it skips. regress 52/52 on the rebuild.

---

**On the sqlsmith failure in the same push** ([run 35077139483](https://github.com/orioledb/orioledb/actions/runs/35077139483)) — it is unrelated to this change:

```
TRAP: failed Assert("((BTreePageHeader *)(p))->field2 >= MAXALIGN(sizeof(BTreeLeafTuphdr)) + MAXALIGN(insert_item->tuplen)"), src/btree/insert.c:1488
#13 o_tbl_index_insert          operations.c:2562
#14 o_apply_new_bridge_index_ctid  operations.c:368
#15 o_tbl_update                operations.c:1932
```

A leaf page of a bridged table's primary index whose free space is short of the item being inserted, on the bridge-ctid apply path. Three things place it outside this PR: the guard added here never fired in that run (no `belongs to another tree now` in the log), the changed code is the page *write* and eviction path rather than insert or free-space accounting, and a rerun of the same run passed all four sqlsmith cells. Worth a ticket of its own.

pashkinelfe:
While discussing this PR it was proposed to change naming ChangeCount->Incarnation or so to have more clear semantics

pashkinelfe (commented):
The current fix fixes a test from PR 1114 https://github.com/orioledb/orioledb/pull/1114/changes/264d4b77d394604aa010f2b3f9bb8828daba3793

I propose to merge this test with the current branch

pashkinelfe:
Looks like a test test_write_through_a_reassigned_meta_page added in this branch hangs, but doesn't fail without a fix

akorotkov:
`check (arm64, 17, gcc, debug)` and `check (arm64, 16, gcc, valgrind_1)` on run 35140930088 failed on a bug of mine in the test, not in the code under test:

```
File ".../test/t/evict_stale_descr_test.py", line 321, in _report
    if self.wrote_unguarded:
AttributeError: 'EvictStaleDescrTest' object has no attribute 'wrote_unguarded'
```

A leftover from dropping the "the bgwriter wrote through unguarded" verdict, which I removed as unsound: `_report()` still asked for the flag that `setUp()` no longer sets. Every local run had taken the early return — the report was always found — so the branch was never executed here. Removed, and this time I ran the skip path deliberately (`ATTEMPTS = 0`):

```
skipped "the bgwriter never reached the page while its descriptor was stale; attempts: {'no_prime': 0, 'no_evict': 0, 'other_root': 0, 'built': 0}"
OK (skipped=1)
```

plus three normal runs, all `built: 1`, 6.2–6.9 s.

Worth noting what those two cells were telling us underneath the traceback: `built: 0` on both, so on those machines the interleaving never came up at all and the test would have skipped — which is what it is meant to do there rather than fail.

akorotkov:
@pashkinelfe Both halves of that are right, and thanks — the second one was the more important.

**The hang.** Root-caused and reproduced locally. The prime step armed `after_ionum_set` on the *tree*, not the writer:

```
pg_stopevent_set('after_ionum_set', '$.treeName == "o_victim_pkey"')
```

and then the main thread issued the pressure statement *synchronously*, with the `pg_stopevent_reset` only after the loop. So when the bgwriter parked mid-write of a victim page, the presser blocked behind its in-flight IO, and the one thread that could reset the event was blocked inside that statement. Nothing breaks that:

```
after_ionum_set │ {47910}   <- orioledb bgwriter, parked
47919 │ client backend │ active │ UPDATE o_churn SET v = v || 'x' WHERE id <= 18000;
```

Fixed with an `_armed()` context manager that resets in `finally`, and by putting anything that may park on its own thread — leaving the block is what releases it. Same treatment for `after_tree_evict` and `after_tree_root_page_alloc`, where the grabber could wedge behind the parked reload the same way. Two other harness defects fell out of chasing it: `_victim_root()` hard-failed when the pool evicted the whole tree between attempts (hit in 3 of 5 forced runs), and `orioledb_tbl_check()` failures reported a bare `false` because `check_btree` reports every cause as a NOTICE, which never reaches the server log.

**"Doesn't fail without a fix."** Confirmed, and my own commit message admitted it: the test asserted the guard's LOG message, which by construction cannot appear when the guard is absent. Measured on a build with the guard compiled out — `skipped`, `OK`, 3 runs out of 3. Worthless as a regression test.

Worth noting why watching for damage isn't enough either: I probed the unfixed build with 5 shapes built, then waited 120 s for the reported symptom and ran `orioledb_tbl_check` on all three trees. No crash, everything clean. The stuck spinlock lives under `USE_ASSERT_CHECKING` on one branch of `perform_page_io`, and whether the garbage seq buf looks locked depends on the bytes that happened to land in the reused block — so the reported symptom is not a reliable oracle.

What closes it is asserting the invariant at the use site, where the seq bufs are taken:

```c
Assert(btree_desc_is_local_temp(desc) || BTREE_META_PAGE_IS_OURS(desc));
```

`walk_page()` establishes this under the page lock and `evict_btree()` re-establishes it after `walk_page_evict_root()` drops the lock, so it is a precondition the callers already owe. With the guard removed the test now fails 3/3 in 5–12 s, in the right process:

```
TRAP: failed Assert("btree_desc_is_local_temp(desc) || BTREE_META_PAGE_IS_OURS(desc)"), src/btree/io.c:2036
background worker "orioledb background writer" was terminated by signal 6
```

That is also the first direct evidence that the constructed shape is the genuinely dangerous one and not merely guard-triggering. The assertion stays quiet through the full regress suite (52/52), which is what says the invariant holds on every path reaching a write, not just the one the test builds.

Pushed as `64476be9`. One thing still open and not from either of these: a single `orioledb_tbl_check('o_churn')` failure on arm64/17 debug that I could not reproduce — 2100 checks across 7 runs of the same shape, plus an A/B that disproved my own bgwriter-race theory. The NOTICE plumbing above is there so the next occurrence says which branch of `check_btree` fired.

akorotkov:
Restructured into two commits so that the property @pashkinelfe asked for is literally checkable, rather than resting on my word:

```
50b8a4a7  Stop a page write from using a meta page the tree no longer owns
ac3c051e  Catch a page write that goes through a foreign meta page
```

```
git checkout ac3c051e && make installcheck   # test/t/evict_stale_descr_test.py FAILS
git checkout 50b8a4a7 && make installcheck   # it passes
```

Measured, three runs each: `ac3c051e` alone gives `FAILED (failures=1)` in 4.4 / 7.8 / 4.1 s; with `50b8a4a7` on top, `OK` in 24 / 45 / 7.4 s.

The first commit carries the test, the two stop events, and the detection: `BTreeRootInfo.metaPageChangeCount` plus the assertion in `perform_page_io()` where the seq bufs are taken. It contains no prevention at all, so it is red. The second commit is only the guard in `walk_page()` / `evict_btree()` and `forget_stale_tree_incarnation()`, and it turns the test green.

Why it was worth splitting: my earlier answer conflated two different things. With the guard removed but the assertion kept, the test failed 3/3 — but with the *whole* commit reverted it went back to a green skip, and I wrote that off as an inherent limit of any test whose oracle ships in the same commit. That was a cop-out. Putting the oracle in its own commit removes the excuse: the detection is test infrastructure and belongs with the test, the guard is the fix, and now the fix is exactly what the red-to-green transition measures.

`git diff 64476be9 HEAD` is **empty** — same tree as the previously reviewed single commit, only the history differs. `ac3c051e` also builds clean under `-Werror` without cassert (the helper that only the guard uses lives in the second commit, so there is no unused-function warning in the first).

regress 52/52 and pgindent clean on the tip.

pashkinelfe (approved):

PR #1173, opened 2026-09-16
Validate seven decoded quantities before they are trusted

akorotkov:
Seven independent findings from an Antithesis triage batch (ORI-269, ORI-274, ORI-285, ORI-291, ORI-295, ORI-297, ORI-301), one commit each. Six turn a value that arrives from disk, shared memory or a WAL record into a checked value before it is used as a size, an index or a loop bound; one is a plain correctness bug with a wrong query result.

## The one with a wrong answer

**Ask a bitmap scan's qualifier once per row** (ORI-295). `o_exec_bitmap_fetch()` evaluated the qual twice per row — once to count and once as the loop condition — so a volatile qual that rejected a row on the first call and accepted it on the second **returned that row**. The test drives an `OR` over a secondary index, because the single-primary streaming path asks once and is unaffected: with the fix 1 row and 2 calls, without it 2 rows and 4 calls, and the extra row is the rejected one.

## The other six

**Refuse a recovery message that does not fit the worker queue** (ORI-285). `worker_send_modify()` flushed when a message did not fit, but a flush does not resize the queue: an oversized message then wrote past the 8 kB buffer. It is now a FATAL, which is what an unroutable record deserves.

**Validate the undo item header before using it** (ORI-297). `undo_item_buf_read_item()` computed `itemSize - sizeof(header)`, which underflows to roughly 4 GB for a small `itemSize`, and `item_type_get_descr()` used the item's type to index a callback table without bounding it. Both come straight off the undo log.

**Bound a free extent by the data file it describes** (ORI-274). `foreach_free_extent()` turns a 64-bit length into one callback per 65535 blocks, so a corrupt length means an effectively endless loop. The extent is now bounded by the length of the file it belongs to.

**Check a relnode undo item's tree counts against its size** (ORI-301). `btree_relnode_undo_callback()` walked `oldNumTrees` and `newNumTrees` entries without asking whether the item was big enough to hold them.

**Check a decoded length before copying it into a fixed buffer** (ORI-269). `copy_fixed_tuple()`, `copy_fixed_key()` and `copy_fixed_shmem_key()` bounded the length with an `Assert` alone, so a non-assert build copied whatever length the page offered into a fixed buffer. The two shared-memory variants PANIC rather than ERROR: they run with a spinlock held.

**Require SELECT on the relation for the structure diagnostics** (ORI-291). `orioledb_tbl_structure()` and `orioledb_idx_structure()` print page contents, including key and tuple data, to any caller. They now require SELECT on the relation and refuse when row-level security applies to it, since the page image cannot honour a row filter.

A note on ORI-291's suggested fix: a per-column `pg_attribute_aclcheck_all(..., ACLMASK_ALL)` check denies even the owner, because it only inspects per-column ACLs and the default `attacl` is NULL — caught by an owner-path test. Table-level SELECT plus the RLS refusal is what this implements.

## Checks

regress 52/52 · isolation 39/39 · recovery_test 59 · recovery_worker_test 3 · undo_eviction_test 3 · checkpointer_test 2 · pgindent clean.

Two findings from the same batch are not here because they need a format decision first: ORI-280 needs a durable bridge flag in `RelnodeUndoStackItem` and therefore an `ORIOLEDB_BINARY_VERSION` bump (the guard suggested in that ticket crashes — `RelidByRelfilenumber()` asserts `IsTransactionState()` on the first commit that drops a bridge), and ORI-267 needs either a per-backend queue in shared memory or location-plus-count in `XidFileRec`.


akorotkov:
The 16 failing cells all had one cause, and it was the test rather than the fix: the `EXPLAIN (COSTS OFF)` I added for ORI-295 prints the children of the `BitmapOr` in the order the planner picks them, and PG 16 and 17 pick the other order than PG 18, where the expected output was generated.

```
-   Recheck Cond: ((id = 7) OR (v = 12))
+   Recheck Cond: ((v = 12) OR (id = 7))
     ->  BitmapOr
-         ->  Bitmap Index Scan on o_qual_once_pkey
          ->  Bitmap Index Scan on o_qual_once_v_idx
+         ->  Bitmap Index Scan on o_qual_once_pkey
```

That hunk was the entire diff in every cell — PG 16 and PG 17, amd64 and arm64, all check types; no PG 18 cell failed and no isolation diff appeared anywhere.

The plan was in the test to show that the query lands on the key-bitmap fetch rather than the single-primary streaming path, which asks the qualifier once and would make the test pass for the wrong reason. Dropping it would lose that; printing it means depending on a planner detail. So it now asks for the shape instead of printing the plan:

```sql
SELECT count(*) FILTER (WHERE l ~ 'BitmapOr') AS bitmap_or,
       count(*) FILTER (WHERE l ~ 'o_qual_once_pkey') AS pkey_scan,
       count(*) FILTER (WHERE l ~ 'o_qual_once_v_idx') AS v_idx_scan
	FROM qual_once_plan() l;
```

`EXPLAIN` without `ANALYZE` executes nothing, so the sequence the qualifier counts is untouched and the call count below is still exact. Amended into the ORI-295 commit rather than stacked on top.

Verified: regress 52/52, isolation 39/39, and with the `bitmap_scan.c` hunk reverted the test still convicts the bug — 2 rows and 4 qualifier calls instead of 1 and 2.

PR #1174, opened 2026-09-16
Refuse a page from disk whose downlink is not on disk

akorotkov:
ORI-262.

## The hole

`DOWNLINK_IS_IN_MEMORY()` is the *absence* of a bit, and `O_GET_IN_MEMORY_PAGE(blkno)` indexes the page pool with the low 31 bits of whatever it is handed:

```c
(O_PAGE_IS_LOCAL(blkno) ? local_ppool_pages[(blkno) & O_BLKNO_MASK] :
 (Page) (o_shared_buffers + (uint64) ((blkno) & O_BLKNO_MASK) * ORIOLEDB_BLCKSZ))
```

with only the all-ones sentinel rejected. So a forged downlink on a durable page becomes a pointer anywhere past the pool, followed by the descent, by page cleanup, and by `orioledb_tbl_check()` — which any role with EXECUTE on it can call.

## The invariant

An in-memory downlink names a block of *this cluster's* page pool and an in-IO one names an IO slot; a file can know about neither. A page that has just come off disk therefore carries on-disk downlinks only — `prewrite_image_check()` already checks that on the way out, and `load_page()` asserted it on the way in. This makes it a real check on the way in, which closes the class at the boundary instead of at the 28 places that follow a downlink.

The bounds of the pool itself are not the useful test: a blkno inside the pool but belonging to another tree is just as wrong, and the cheap invariant — nothing from a file may claim to be in memory at all — rules that out too.

## What it costs

Measured separately (the benchmark is not part of this PR). The pool at its minimum, 8 MB, against a 400k-row tree of 1004 leaves and 7 non-leaf pages with ~144 items each; three phases — random primary key lookups, one sequential scan, and evict-the-tree-then-look-one-key-up repeated, which is the highest rate of check work a query can ask for. Three builds, 3 runs each:

| phase | as shipped | check off | check ×256 |
|---|---|---|---|
| point (random PK lookups) | 344.7 ms | 349.7 ms | 332.6 ms |
| scan (one seq scan) | 140.8 ms | 140.7 ms | 141.5 ms |
| descend (evict, look up, repeat) | 391.3 ms | 413.7 ms | 680.3 ms |

Standard deviations: 5–10 ms for point, ~1 ms for scan, 16–34 ms for descend.

A direct A/B cannot resolve the check at all: 391 ms against 414 ms with a standard deviation of 16–20 ms, and with the sign inverted. Hence the third build, with the call wrapped in a loop of 256, whose slope gives `(680.3 − 391.3) / 255 / 600 = 1.9 µs` per descent — i.e. **~0.9 µs per non-leaf page load, ~6 ns per item**. The ×256-vs-off pair independently gives 1.7 µs and 0.87 µs, so the two estimates agree.

At that price one pass is **~0.3%** of the 652 µs a descent costs *in the phase built to maximize it*. The scan phase is flat across all three builds, as it should be — a leaf page returns before an item is looked at. And in the tree measured, 7 pages of 1011 are non-leaf, so under an ordinary read workload 99.3% of page loads pay nothing at all.

## Checks

regress 52/52 · amcheck_test 10 · eviction_test 16 — no legitimate page trips the check.


pashkinelfe:
It might be unneeded resource consumption to check all page downlinks. On-disk pages are protected by checksums and this check is supposedly only for integrity of logic, page corruption looks unlikely. Maybe ori-262 could be closed without code changes.

akorotkov:
@pashkinelfe, agreed.

PR #1180, opened 2026-09-16
Park only the process merge_test is waiting for

akorotkov:
Builds on #1171, which found the half of this I did not have. Either that PR or this one can go first; as it stands this contains both fixes and #1171 becomes redundant.

## Two ways the test parks the wrong process

`test_merge_into_concurrently_dropped_tree` arms its stop event for everyone:

```python
ctrl.execute("SELECT pg_stopevent_set('merge_after_target_unlock', 'true');")
```

Everything that runs page pool maintenance reaches that window, so `'true'` parks whoever gets there first — including the test's own sessions. A test session that parks is a deadlock: the steps that would release the event are the ones that never run.

That is what hung [run 35085730232](https://github.com/orioledb/orioledb/actions/runs/35085730232) (`check (amd64, 17, clang, normal)`, on an unrelated branch) for its whole 30-minute budget. `ci/list_stuck.sh` caught three processes in the same window:

```
#4  handle_stopevent
#5  btree_try_merge_and_unlock
#6  walk_page
#7  o_ppool_run_maintenance
#8  o_ppool_reserve_pages
#9  checkpoint_init_new_seq_bufs        <- checkpointer
...
#8  bgwriter_main                       <- orioledb bgwriter
...
#9  load_page
#10 find_page
#11 o_btree_iterator_create
#14 o_index_scan_getnext
#18 ExecLimit                           <- the test's own warm-up loop
```

`ExecLimit` identifies the last one: the `SELECT 1 FROM o_mNN ORDER BY id LIMIT 1` warm-up, which in the current code runs *after* the arming.

## What this changes

* The event is armed only for what the test is not driving, by `application_name`, and the warm-up moves ahead of the arming. What is left to park is the presser — the scan this test means to catch in the window — and the checkpointer, which reaches it just as legitimately.
* The bgwriter is taken out of the picture with `orioledb.debug_disable_bgwriter`, **from #1171**: `bgwriter_num_workers` has a minimum of 1, so the `= 0` the test used only logs

  ```
  WARNING:  0 is outside the valid range for parameter "orioledb.bgwriter_num_workers" (1 .. 262143)
  ```

  and leaves the worker running — which is how it came to be parked in that window at all. Measured on a live cluster: with `bgwriter_num_workers = 0` one `orioledb background writer` is in `pg_stat_activity`; with `debug_disable_bgwriter = true`, none.

The application-name exclusion is still worth having with the bgwriter gone, because the warm-up is not the only driver session that can reach the window — the `DROP TABLE` connection runs page pool maintenance too, and it is the step that releases everything.

The same defect has the other failure mode as well: the deadline reporting "nothing parked in the merge window" when the process that parked was one the test was not waiting for. An earlier stand run measured 283 stalls in 6044 iterations before the scoping change and 0 in 1568 after.

## Checks

`merge_test` as a suite (15 tests) and this test three times on its own, 37–42 s each. No local reproduction of the stall is offered: at the measured rate it needs hundreds of iterations, and the stack dump above is the evidence.

Unrelated to that run's other state: its four valgrind cells were still running normally, and only the `normal` cell hit the timeout.


homper (approved):
Ok) Hope this stabilizes it

akorotkov:
The valgrind cell on this branch failed, and it is not a hang — it is the *other* failure mode of the same test:

```
FAIL: test_merge_into_concurrently_dropped_tree
AssertionError: unexpectedly None : nothing parked in the merge window
```

Nothing reached the window within the 90-second deadline. The whole `merge_test` module took 1841 s in that cell, so 90 s of polling is a fraction of what one step costs there; the presser was still working through its first scan of `o_filler` (60 MB against an 8 MB pool) when the deadline expired. Disabling the bgwriter makes it stricter, too: that worker used to reach the window as well, and the poll accepted whichever process got there.

So the timeout now follows `USE_VALGRIND` — 90 s normally, 900 s under Valgrind, the same idiom `multi_insert_test.py` uses — and when it does expire it prints what every process was doing, which is what tells "the presser has not got there yet" apart from "the sparse pages are gone".

`(arm64, 18, gcc, valgrind_1)` was the only failing cell of that run. One more thing that run makes clear: `test_rewind_xid_oriole_restart ... FAIL` in the same log is interleaved output from a parallel `make` job — `rewind_xid_test` reported OK for all of its tests, and `merge_test.py` was the only `Error 1`.

I cannot run Valgrind on this machine (arm64 macOS), so the timeout is reasoned from that cell's own numbers rather than measured. Locally: the suite (15 tests), this test three times on its own, and once with `USE_VALGRIND=1` to take the other branch of the constant.

PR #1181, opened 2026-09-16
Resolve a key's collation before a comparison can need a page

akorotkov:
Recovery aborts, and the cluster will not start again, when the first WAL record it replays for a table is a DELETE or an UPDATE and the table's key has a collation of its own:

```
TRAP: failed Assert("!have_locked_pages()"), File: "src/utils/page_pool.c", Line: 182
```

```
#0  o_ppool_reserve_pages                src/utils/page_pool.c:182
#1  load_page                            src/btree/io.c
#2  find_page                            src/btree/find.c      <- INNER find, on the class cache sys tree
#3  o_sys_cache_search                   src/catalog/o_sys_cache.c
#4  o_SearchCatCacheInternal_hook
#5  lookup_collation_cache / pg_newlocale_from_collation
#6  varstr_cmp / bttextcmp / FunctionCall2Coll
#7  o_call_comparator                    src/tableam/descr.c
#8  btree_page_search                    src/btree/find.c
#9  find_page                            src/btree/find.c      <- OUTER find, leaf page LOCKED (MODIFY)
#10 o_btree_normal_modify                src/btree/modify.c
#11 apply_modify_record                  src/recovery/worker.c
```

A comparison function for a collatable type asks pg_locale.c for the collation, and pg_locale.c reads pg_collation the first time a process uses it. Under recovery that read is answered from OrioleDB's own system trees, which can mean reading a page — and the comparison runs inside `find_page()`, on a leaf page locked for modification, where reserving a page is forbidden. Recovery then fails, the postmaster shuts down "due to startup process failure", and the next start replays the same record: a crash loop. Reported as ORI-328, found by Antithesis on the Jepsen `internal-sim` workload, whose tables have a `TEXT collate unicode` key.

## Why a backend never gets there

`o_find_opclass_comparator()` asks the type's **sort support** function for a comparator, and the sort support function resolves the collation right there, while nothing is locked. But it only does so when the process has a database of its own:

```c
	if (MyDatabaseId == opclass->key.common.datoid && OidIsValid(opclass->ssupOid))
```

Single-process recovery — what the postmaster runs after a backend crash — has no database, falls through to the plain comparison function, and arrives at its first comparison with nothing resolved. Recovery *workers* are connected to a database and are therefore safe, which is why `pg_ctl -m immediate` does not show this.

So resolve the collation where the comparator is built, on that same fallback path. pg_locale.c caches the result per process, making this one lookup per collation per process, and the page pool is never asked for a page under a lock again.

The insert path's own warm-up — the `HACK` in `apply_tbl_insert()` that compares every key bound before `o_btree_modify()` — is what kept INSERTs safe all along, and is now redundant. I left it alone to keep this change to the one mechanism.

## The checkpointer has the same problem

It compares index keys while writing a tree and has no transaction to read pg_collation in, so an ICU-collated key aborts it on every checkpoint of a tree with internal pages. `pg_newlocale_from_collation()` requires a transaction and exempts only recovery, so the call here is skipped where neither holds and the checkpointer is left as it was. Fixing it needs a PostgreSQL-side change (skip that assertion for a lookup an extension answers itself); that is prepared separately and will come with its own patchset tag.

## Test

`test/t/recovery_collation_test.py` drives all four conditions the failure needs, and the two that are not obvious cost the most:

* an ICU collation on the key, and a DELETE or an UPDATE as the first replayed record;
* **a one-page tree** — in a taller one `find_page()` compares on internal pages first, and those are read without a lock, which resolves the collation by accident. A 400-row table does not reproduce at all;
* **a killed backend**, which is what makes the recovery single-process (`recovery_single = IsFatalError()`). It also needs `restart_after_crash = on`, which BaseTest turns off.

Without this fix both cases leave the node unable to start.

## Checks

On a stock patchset (`pppa18`, no PostgreSQL-side change): regress 52/52 · isolation 39/39 · recovery_collation_test 2 · collate_test 9 (1 skipped).


akorotkov:
Two cells failed; one is mine, the other is not.

## `check (arm64, 16, gcc, debug)` — my test, and a race in its harness

```
FAIL: test_first_replayed_record_is_a_delete
AssertionError: 'Unable to make multiprocess recovery' not found in '...'
```

The log it printed ends like this:

```
12:16:11.940 [28835] LOG:  statement: SELECT 1;
12:16:11.941 [28799] LOG:  server process (PID 28833) was terminated by signal 9: Killed
12:16:11.941 [28799] LOG:  terminating any other active server processes
```

That `SELECT 1` is the test's own wait for the node to come back, and it is logged **one millisecond before** the postmaster noticed the killed backend. Waiting for a connection to succeed does not mean the crash has been dealt with: for the first milliseconds the old backends are still running and serve it. The helper returned immediately, the test looked for a recovery that had not started yet, and the tearDown then stopped a node in the middle of reinitialising — which is where the `pg_ctl: server does not shut down` in the same traceback comes from. The stack dump taken later shows that cluster idle and healthy: a postmaster in `ServerLoop` and a checkpointer in its hour-long wait, nothing stuck.

The helper now waits for the postmaster's own account of what it is doing — `terminating any other active server processes`, then `database system is ready to accept connections` — before touching the node, and the assertions read only the part of the log this crash wrote, not whatever was there before.

The engine change is not implicated: `test_first_replayed_record_is_an_update` passed in the same cell, and it needs the same code path.

Six consecutive local runs pass. (Two of my own runs failed first with the real TRAP — because I had rebuilt the extension from a different branch; `make clean` after switching branches, every time.)

## `check (arm64, 17, gcc, sanitize)` — not this PR

The step hit its 40-minute budget. `recovery_collation_test` ran in that cell and passed, in 2.4 s and 1.4 s; the cell simply did not finish its work in time. The same cell is green on #1173, #1178 and #1182, so it is the runner rather than anything in these three files. I will rerun it.

pashkinelfe (approved):
LGTM

PR #1182, opened 2026-09-16
Hash the conflicting tuple before releasing its page

akorotkov:
`o_btree_modify_handle_conflicts()` unlocks the leaf page and then hands the tuple to `wait_for_tuple()`:

```c
	unlock_page(blkno);

	Assert(cbAction <= OBTreeCallbackActionXidExit);

	if (cbAction == OBTreeCallbackActionXidWait)
		wait_for_tuple(desc, curTuple, oxid, ...);
```

`curTuple` points *into* that page — it came from `BTREE_PAGE_READ_LEAF_ITEM()` at the top of the function — and `wait_for_tuple()` hashes it to build the heavyweight tuple lock tag. By then the page is unlocked, and the transaction being waited for is the one working on it.

## Measured, not argued

A build that computes the hash on both sides of the unlock and compares logged a mismatch:

```
DBGSTALEHASH 2042858748 != 2043140225
```

under 12 pgbench clients doing conflicting `UPDATE ... WHERE k = 'key1..3'` plus 6 clients inserting and deleting around them on the same page — a few thousand conflict waits is enough to catch one. So the window is live on ordinary hardware, not only under a fault injector.

## What it costs

Two things.

The lock tag identifies the wrong tuple, so the row lock protocol silently guards something else.

And the hash function of a collatable type *copies* the value it is handed: a copy sized by one read of the value's length and filled by another writes past its own buffer when the length changes in between. That is a heap overflow inside `memcpy` — the shape of the SIGSEGV reported as ORI-329, which faulted on a write at the end of the heap inside `memcpy`, in a backend parked in `wait_for_oxid()`.

## The fix

Hash the tuple where it is still ours — before the unlock — and pass the `uint32` down. The lookup is on the slow path already, and the hash is computed only when the wait is actually going to take the tuple lock, which is the same condition `wait_for_tuple()` used.

## What this does not claim

That it is all of ORI-329. The crash site there was never identified, and the wake-up without a lock grant that the ticket describes is still unexplained. The same workload also reproduces a separate `ERROR: compressed pglz data is corrupt` on the read side, on unfixed main as well — untriaged, and the read-side twin of the same "tuple bytes moved underneath us" problem.

## Checks

regress 52/52 · isolation 39/39 · recovery_test 59 · commit_window_test 2.


akorotkov:
Rebased onto current main. Both failures were the same thing, and neither was this change:

```
##[error]The action 'Check' has timed out after 30 minutes.
```

in `check (arm64, 18, gcc, debug)` and `check (amd64, 18, clang, check_page)` — and in both, `ci/list_stuck.sh` found the same cluster alive at the end:

```
postgres -D .../test/tmp_check_t/merge-merge_into_concurrently_dropped_tree_tgsn/data
python3 -m unittest -v test/t/merge_test.py
```

That is the `merge_test` deadlock: the test armed `merge_after_target_unlock` for every process, including its own, so whichever session got there first could be one the test needed in order to release the event. #1180 fixed it and has since landed, and this branch predated it — hence the rebase.

Checks after the rebase: regress 52/52, and `merge_test` (15 tests) green against the merged fix.

PR #1183, opened 2026-09-16
Reject foreign metadata on a relnode collision in O_TABLES fetch

homper:
The O_TABLES system tree comparator orders by (datoid, relnode, chunknum) and omits reloid, so a lookup whose reloid differs from the chunk it lands on still reaches that chunk when (datoid, relnode, chunknum) collides.  The fetch callback's fallback unconditionally returned OTupleFetchMatch for such an unmatched tuple, so the iterator returned the foreign relation's chunk and o_tables_get_extended() deserialized another table's descriptor -- a cross-relation metadata leak reachable through a normal describe query.  Reject the chunk instead: returning OTupleFetchNotMatch lets the iterator convert it into a NULL result, matching the (reloid, relnode) key against the full OIDs before any deserialization.  Added a testgres test that queries o_t1's reloid with o_t2's relnode and asserts the description is not found.


pashkinelfe (approved):
LGTM

PR #1187, opened 2026-09-17
Gather statistics of orioledb's blocks

bwyltcaxe:
Orioledb relations didn't report page accessess to 'pg_statio_all_indexes' and 'pg_statio_all_tables' before, so tools based on statistics (e.g. pg_profile) didn't cover orioledb relations.

This patch introduces implementation of such statistics gathering:

	* read_page_from_disk => read page
	* find_page and seq_scan => hit page

Counters are implemented just like Postgres's usual statistics schedule: pg_stat_force_next_flush() makes them visible immediately.


bwyltcaxe:
This patch also needs core's changes: https://github.com/orioledb/postgres/pull/88. This is my first try to push smth to orioledb's core. I am not sure that I've done everything correctly...

homper:
Was slightly confused, but yes, it makes sense, yes

LLM output was this:
```
Ordinary in-memory access produces:
read = 0
hit  = 1
Storage load produces:
read = 1
hit  = 0
The exception is the PR’s root-page bug: when a root is loaded from storage
read = 1
hit  = 1
```

And for root it makes sense for PR not to be that intrusive

---

> This is my first try to push smth to orioledb's core. I am not sure that I've done everything correctly...

There are two options: we backport PR to one branch or it could be 3 PR's with different pathces** branches as base
I will backport to all other versions and update .pgtags tomorrow

bwyltcaxe:
> Was slightly confused, but yes, it makes sense, yes
> 
> LLM output was this:
> 
> ```
> Ordinary in-memory access produces:
> read = 0
> hit  = 1
> Storage load produces:
> read = 1
> hit  = 0
> The exception is the PR’s root-page bug: when a root is loaded from storage
> read = 1
> hit  = 1
> ```
> 
> And for root it makes sense for PR not to be that intrusive

Oh, thanks. I think we can fix it in the following way:
```
find_page()
{
...
root_was_at_page_pool = ORootPageIsValid(desc) && OMetaPageIsValid(desc);
if (!tryFlag)
     o_btree_load_shmem(desc);
...
if (root_was_at_page_pool)
    o_btree_count_page_hit(desc);
...
}
```

What do you think?

> 
> > This is my first try to push smth to orioledb's core. I am not sure that I've done everything correctly...
> 
> There are two options: we backport PR to one branch or it could be 3 PR's with different pathces** branches as base I will backport to all other versions and update .pgtags tomorrow

Big thanks.

By the way, are we ok with such edits on core's side? Currently there is no way for custom AMs to gather statistics -- everything MUST go through postgres' own buffer read, so it's kinda hopeless :(. Should we try to push this to community or what's the plan to utilize the core's side patch?

PR #1195, opened 2026-09-19
Keep a tree loaded while somebody reaches into its shared state (#1133, #1196)

### LINKED ISSUE #1133: Assert in 027_stream_regress
On 25a550ac0e81a627fbc8ae4409d71c3ab308231f  PG18/pg_tests test fails:

https://github.com/orioledb/orioledb/actions/runs/34279598204/job/102240976397

```
Run bash ./orioledb/ci/check_output.sh
dumping /tmp/cores-25a550ac0e81a627fbc8ae4409d71c3ab308231f-1788902194/1788903344_50361.core for /home/runner/_work/orioledb/orioledb/pgsql/bin/postgres

warning: Can't open file /dev/zero (deleted) during file-backed mapping note processing

warning: Can't open file /dev/shm/PostgreSQL.962547850 during file-backed mapping note processing

warning: Can't open file /SYSV00366ca3 (deleted) during file-backed mapping note processing
[New LWP 50361]

warning: 44	./nptl/pthread_kill.c: No such file or directory
[Thread debugging using libthread_db enabled]
Using host libthread_db library "/lib/aarch64-linux-gnu/libthread_db.so.1".
Core was generated by `postgres: standby_1: orioledb background writer 0                             '.
Program terminated with signal SIGABRT, Aborted.
#0  __pthread_kill_implementation (threadid=281239661329568, signo=signo@entry=6, no_tid=no_tid@entry=0) at ./nptl/pthread_kill.c:44

Thread 1 (Thread 0xffc9361c3ca0 (LWP 50361)):
#0  __pthread_kill_implementation (threadid=281239661329568, signo=signo@entry=6, no_tid=no_tid@entry=0) at ./nptl/pthread_kill.c:44
        tid = 50361
        ret = 0
        pd = 0xffc9361c3ca0
        old_mask = {__val = {281239661413200}}
        ret = <optimized out>
#1  0x0000ffc9359b7540 in __pthread_kill_internal (signo=6, threadid=<optimized out>) at ./nptl/pthread_kill.c:78
No locals.
#2  0x0000ffc93596cb3c in __GI_raise (sig=sig@entry=6) at ../sysdeps/posix/raise.c:26
        ret = <optimized out>
#3  0x0000ffc935957e00 in __GI_abort () at ./stdlib/abort.c:79
        save_stage = 1
        act = {__sigaction_handler = {sa_handler = 0x20, sa_sigaction = 0x20}, sa_mask = {__val = {281239653764744, 1, 281239653788328, 3, 281474280766092, 12, 281239653788336, 2, 3558233349350752256, 3473176235313488997, 0, 3473177153597865984, 3982807437535588096, 281474280767040, 188711961205160, 12}}, sa_flags = -695944424, sa_restorer = 0xffc935ae1600 <stderr>}
#4  0x0000aba1ed6999ac in ExceptionalCondition (conditionName=conditionName@entry=0xffc933326340 "meta_page_get_num_seq_scans(desc->rootInfo.metaPageBlkno) == 0", fileName=fileName@entry=0xffc933325068 "src/btree/io.c", lineNumber=lineNumber@entry=2767) at assert.c:66
No locals.
#5  0x0000ffc9331b3c4c in evict_btree (checkpoint_number=<optimized out>, desc=0xaba2206c46e0) at src/btree/io.c:2767
        rootPageBlkno = <optimized out>
        metaPage = 0xffc92d4fbb80
        evicted_tree_data = {key = {datoid = 0, relnode = 0, tablespace = 0}, file_header = {ctid = 0, bridgeCtid = 0, rootDownlink = 0, datafileLength = 0, numFreeBlocks = 0, leafPagesNum = 0}, maxLocation = {0, 0}, freeBuf = {offset = 0, tag = {key = {oids = {datoid = 0, reloid = 0, relnode = 0, spcoid = 0}}, num = 0, type = 0 '\000'}}, nextChkp = {offset = 0, tag = {key = {oids = {datoid = 0, reloid = 0, relnode = 0, spcoid = 0}}, num = 0, type = 0 '\000'}}, tmpBuf = {offset = 0, tag = {key = {oids = {datoid = 0, reloid = 0, relnode = 0, spcoid = 0}}, num = 0, type = 0 '\000'}}, dirtyFlag1 = false, dirtyFlag2 = false, punchHolesChkpNum = 0}
        img = "\377\3777\001\000\0000\000\005\000\000\000\005\000\000\000\000\000\000\000\000\000\000 \003\000\000\000\000\000\000\000\377\377\377\377\377\377\377\377'\000\000\000\b\000\000\000\001\000\001\000@\000(\001@\000\000$", '\000' <repeats 196 times>...
        notModified = <optimized out>
        evict_key = {datoid = 16566, relnode = 16694, tablespace = 1663}
        evict_lockNo = <optimized out>
        root_blkno = <optimized out>
        root_desc = <optimized out>
        new_downlink = 9223653511831486464
        was_dirty = <optimized out>
        chkpNum = 859808832
        file_header = {ctid = 0, bridgeCtid = 0, rootDownlink = 0, datafileLength = 0, numFreeBlocks = <optimized out>, leafPagesNum = <optimized out>}
        hasMetaLock = <optimized out>
        root_blkno = <optimized out>
        rootPageBlkno = <optimized out>
        root_desc = <optimized out>
        metaPage = <optimized out>
        file_header = <optimized out>
        evicted_tree_data = <optimized out>
        new_downlink = <optimized out>
        img = <optimized out>
        was_dirty = <optimized out>
        chkpNum = <optimized out>
        notModified = <optimized out>
        hasMetaLock = <optimized out>
        evict_key = <optimized out>
        evict_lockNo = <optimized out>
        __func__ = "evict_btree"
        not_used = <optimized out>
        __errno_location = <optimized out>
        __errno_location = <optimized out>
#6  walk_page_evict_root (oids=..., p=<optimized out>, page_desc=<optimized out>, blkno=760200064, desc=<optimized out>) at src/btree/io.c:3176
        copy_blkno = false
        locksState = {indexRegularLock = false, indexCheckpointerLock = true, tableRegularLock = false, tableCheckpointerLock = false, tableOids = {datoid = 16566, reloid = 16694, relnode = 16694, spcoid = 1663}}
        checkpoint_number = 5
        result = false
        ionum = -1
        locksState = <optimized out>
        checkpoint_number = <optimized out>
        copy_blkno = <optimized out>
        result = <optimized out>
        ionum = <optimized out>
#7  walk_page (blkno=blkno@entry=2106, evict=evict@entry=true) at src/btree/io.c:3319
        page_desc = <optimized out>
        context = {desc = 0xaba2206c46e0, img = 0x0, parentImg = 0x0, imgData = "\377\377\243\033\000\000`\000\000\000\000\000\002\000\000\000\000\000\000\000\000\000\000 \003\000\000\000\000\000\000\000\377\377\377\377\377\377\377\377\243\000\000\000\b\000\000\b\001\000\002\000@\000 \002\200\000\000$", '\000' <repeats 452 times>..., parentImgData = '\000' <repeats 6184 times>..., partial = {src = 0xffc92dacdb80 "\377\377\243\033", isPartial = false, hikeysChunkIsLoaded = true, chunkIsLoaded = {true, false <repeats 37 times>}}, parentPartial = {src = 0xaba2205cf270 " \361\\ \242\253", isPartial = false, hikeysChunkIsLoaded = false, chunkIsLoaded = {133, 214, 255, 255, false, false, 61, false <repeats 16 times>, 175, 23, 59, 67, 198, 69, 55, 208, 4, 133, 214, 255, 255, false, false}}, csn = 0, imgReadCsn = 188712816014219, imgUndoLoc = 2305843009213693952, index = 0, items = {{blkno = 6946, pageChangeCount = 3, locator = {chunkOffset = 21, itemOffset = 2, chunkItemsCount = 18, chunkSize = 472, chunk = 0xffc92faca850}}, {blkno = 3284, pageChangeCount = 0, locator = {chunkOffset = 36, itemOffset = 11, chunkItemsCount = 12, chunkSize = 216, chunk = 0xffc92de2f7b8}}, {blkno = 1865378674, pageChangeCount = 1819240818, locator = {chunkOffset = 25701, itemOffset = 12130, chunkItemsCount = 29295, chunkSize = 28521, chunk = 0x7367702f6264656c}}, {blkno = 1932487793, pageChangeCount = 1701994856, locator = {chunkOffset = 29743, itemOffset = 28009, chunkItemsCount = 31333, chunkSize = 28271, chunk = 0x54552f6374452f65}}, {blkno = 67, pageChangeCount = 0, locator = {chunkOffset = 0, itemOffset = 0, chunkItemsCount = 0, chunkSize = 0, chunk = 0x0}}, {blkno = 0, pageChangeCount = 0, locator = {chunkOffset = 0, itemOffset = 0, chunkItemsCount = 0, chunkSize = 0, chunk = 0x0}} <repeats 27 times>}, insertXactInfo = 0, insertTuple = {data = 0x0, formatFlags = 0 '\000'}, lokey = {tuple = {data = 0x0, formatFlags = 0 '\000'}, fixedData = '\000' <repeats 96 times>, "@\004\205\326\377\377\000\000\370\215\2415\311\377\000\000p\004\205\326\377\377\000\000L:\2345\311\377\000\000P\n\2565\311\377\000\000P2\001\000\000\000\000\000\000\020\002", '\000' <repeats 13 times>, "\240\004\205\326\377\377\000\000\344I\2345\311\377\000\000\000\360\\ \242\253\000\000Xk\0356\311\377", '\000' <repeats 11 times>...}, leafLokey = {tuple = {data = 0x0, formatFlags = 0 '\000'}, fixedData = '\000' <repeats 2687 times>}, undoLokey = {tuple = {data = 0x0, formatFlags = 0 '\000'}, fixedData = '\000' <repeats 1504 times>...}, parentImgDeferred = false, flags = 192}
        desc = <optimized out>
        p = <optimized out>
        parent_page = <optimized out>
        oids = {datoid = 16566, reloid = 16694, relnode = 16694, spcoid = 1663}
        int_hdr = <optimized out>
        checkpoint_number = 5
        copy_blkno = false
        merge_tried = <optimized out>
        findResult = <optimized out>
        ionum = -1
        img = '\000' <repeats 401 times>...
        is_root = <optimized out>
        checkResult = <optimized out>
        retry = <optimized out>
#8  0x0000ffc9333153d8 in o_ppool_run_maintenance (pool=0xffc9333d1c48 <page_pools>, evict=true, shutdown_requested=0xaba1eda9e258 <ShutdownRequestPending>) at src/utils/page_pool.c:445
        blkno = 2106
        undoRegularSize = 0
        undoSystemSize = 0
        haveRetainRegularLoc = false
        haveRetainSystemLoc = false
        o_pool = 0xffc9333d1c48 <page_pools>
        skippedLocalEvictions = 0
        skippedLocalEvictionsLimit = 57344
        lastPageEvictSharedCount = 220
        exhausted = false
#9  0x0000ffc933311354 in ppool_run_maintenance (shutdown_requested=<optimized out>, evict=true, pool=0xffc9333d1c48 <page_pools>) at /home/runner/_work/orioledb/orioledb/orioledb/include/utils/page_pool.h:121
No locals.
#10 bgwriter_main (main_arg=<optimized out>) at src/workers/bgwriter.c:165
        i = 14
        poolType = OPagePoolMain
        lastUsedLocation = <optimized out>
        j = <optimized out>
        _save_exception_stack = 0xffffd68566e0
        _save_context_stack = 0x0
        _local_sigjmp_buf = {{__jmpbuf = {281239613314512, 188712816776144, 188712816767936, 281239613314544, 0, 188711965206968, 1000, 1, 188711965310976, 188712816478408, 281474280810096, 3610304268102632872, 4299262264297, 3610304344800489888, 0, 0, 0, 0, 0, 0, 0, 0}, __mask_was_saved = 0, __saved_mask = {__val = {32728308, 0, 0, 0, 0, 188712816483208, 0, 72339068318738000, 0, 0, 3982807437535588096, 281474280810096, 281239652865584, 281239661408256, 0, 281474280810096}}}}
        _do_rethrow = <optimized out>
        pool = 0xffc9333d1c48 <page_pools>
        rc = <optimized out>
        wake_events = 25
        need_eviction = true
        need_write = <optimized out>
        appname = "orioledb background writer 0\000\253\000\000 B[ \242\253\000\000\220f\205\326\377\377\000\000\024+j\355\241\253\000\000\024\01613\311\377\000"
        __func__ = "bgwriter_main"
#11 0x0000aba1ed4585fc in BackgroundWorkerMain (startup_data=startup_data@entry=0xaba22058a750, startup_data_len=startup_data_len@entry=1472) at bgworker.c:843
        local_sigjmp_buf = {{__jmpbuf = {188711965523112, 1472, 1472, 75, 0, 188711965206968, 1000, 1, 188711965310976, 188712816478408, 281474280810528, 3610256266772942852, 5, 3610304344800490480, 0, 0, 0, 0, 0, 0, 0, 0}, __mask_was_saved = 1, __saved_mask = {__val = {18446744066192964099, 0, 3982807437535588096, 281474280810448, 3982807437535588096, 281474280810496, 281239653240212, 281474280810480, 281474280810560, 281474280810560, 281474280810544, 4294967288, 3982807437535588096, 281474280810560, 188711961281804, 5}}}}
        worker = 0xaba220584450
        entrypt = <optimized out>
        __func__ = "BackgroundWorkerMain"
#12 0x0000aba1ed45b0c4 in postmaster_child_launch (child_type=<optimized out>, child_slot=75, startup_data=startup_data@entry=0xaba22058a750, startup_data_len=startup_data_len@entry=1472, client_sock=client_sock@entry=0x0) at launch_backend.c:291
        pid = <optimized out>
#13 0x0000aba1ed45cf28 in StartBackgroundWorker (rw=0xaba22058a750) at postmaster.c:4102
        bn = 0xaba2206408c8
        worker_pid = <optimized out>
        bn = <optimized out>
        worker_pid = <optimized out>
        __func__ = "StartBackgroundWorker"
        __errno_location = <optimized out>
        __errno_location = <optimized out>
        __errno_location = <optimized out>
#14 maybe_start_bgworkers () at postmaster.c:4273
        rw = 0xaba22058a750
        num_launched = 0
        now = 0
        iter = {cur = 0xaba22058ad28, next = 0xaba22058a228, end = <optimized out>}
#15 0x0000aba1ed4608a8 in PostmasterMain (argc=argc@entry=4, argv=argv@entry=0xaba220582f60) at postmaster.c:1381
        opt = <optimized out>
        status = <optimized out>
        userDoption = <optimized out>
        listen_addr_saved = <optimized out>
        output_config_variable = <optimized out>
        __func__ = "PostmasterMain"
#16 0x0000aba1ed0dafa8 in main (argc=4, argv=0xaba220582f60) at main.c:227
        do_check_root = <optimized out>
        dispatch_option = <optimized out>
$1 = {{lock = 0xffc92be301dc, mode = LW_EXCLUSIVE}, {lock = 0xffc92be2ff54, mode = LW_SHARED}}
$2 = 0xffc9333b2f38 <myLockedPages>
#16 0x0000aba1ed0dafa8 in main (argc=4, argv=0xaba220582f60) at main.c:227
227				PostmasterMain(argc, argv);
$3 = 0xaba220582f90 "/home/runner/_work/orioledb/orioledb/pgsql/bin/postgres"
$4 = 0xaba220582fd0 "-D"
$5 = 0xaba220582ff0 "/home/runner/_work/orioledb/orioledb/postgresql/src/test/recovery/tmp_check/t_027_stream_regress_standby_1_data/pgdata"
$6 = 0xaba220583070 "--cluster-name=standby_1"
tar: Removing leading `/' from member names
tar: Removing leading `/' from hard link targets
```

### LINKED ISSUE #1196: Users of a tree's shared state trust a cached descriptor: safe on a primary only by the eviction lock, unsafe on a standby
Splitting this out of #1133, which is one instance of it.

A `BTreeDescr` is cached per backend and nothing keeps its tree loaded. The clock sweep can evict the tree and hand the root and meta pages to somebody else while the descriptor goes on naming them. Callers fall into two groups, and the second one is not protected.

## Group A — load, then descend: no problem, and the load is redundant

`find_page()` loads the tree itself on entry, compares the root's change count on the way down, and zeroes `desc->rootInfo` and reloads when it does not match. Every page it touches is validated. Callers that pre-load and then descend gain nothing from the pre-load:

- `src/recovery/recovery.c:530`, `:681`, `:2846`
- `src/catalog/o_indices.c:2402`

(Worth removing separately, and worth checking first whether any of them relies on a side effect such as `ppool_reserve_pages()`.)

## Group B — load, then read or write shared state directly

These take a word out of the meta page, or start from `rootInfo.rootPageBlkno`, without anything re-validating afterwards. `o_btree_load_shmem()` cannot help them: it only guarantees the tree *exists*, and its answer expires the moment it returns — `ORootPageIsValid() && OMetaPageIsValid()` tests that block numbers are set, not that they are still this tree's.

| site | what it touches | if the page is foreign |
|---|---|---|
| `src/recovery/worker.c:991`, `:1000` → `btree_ctid_update_if_needed()` | **CAS write** to `metaPage->ctid` | silently corrupts another tree's ctid counter |
| `src/catalog/free_extents.c:604` | takes `metaPage->metaLock` | an LWLock inside another tree's page — the #1113 damage shape |
| `src/catalog/indices.c:2471` | reads `bridge_ctid` to seed with | seeds from garbage |
| `src/indexam/handler.c:1147`, `src/catalog/indices.c:1948` | `TREE_NUM_LEAF_PAGES` | wrong plan estimate |
| `src/orioledb.c:2365` | root page → `tree_height` | wrong value |
| `src/btree/io.c:3585`, `:3595`; `src/btree/undo.c:2401`, `:2498`; `src/btree/check.c:805` | walk from `rootInfo.rootPageBlkno` | walks another tree |
| `src/btree/scan.c` `init_checkpoit_number()` | registers into `numSeqScans` | **issue #1133** — fixed in #1195 |

## Why it only bites on a standby today

On a primary, `get_evict_btree_locks()` takes an ordinary `AccessExclusiveLock` on the relation, which conflicts with the `AccessShareLock` any reader holds until its transaction ends. A tree whose table someone has open simply is not evicted, so group B never meets a foreign page.

During recovery that lock is skipped — `LockAcquireExtended()` refuses anything above `RowExclusiveLock` on a `LOCKTAG_RELATION` while `RecoveryInProgress() && !InRecovery` — and nothing is asked in its place:

- `src/recovery/worker.c:439` `o_fetch_table_descr(oids)` and `:445` `o_fetch_index_descr(oids, ix_type, /* lock = */ false, NULL)` — the recovery worker holds **no** orioledb tree lock, of either flavour;
- an ordinary backend with a warm `rel->rd_amcache` takes none either, measured on the #1133 repro: the evictor held two `userlock` entries and the reader only ordinary `relation` locks — disjoint namespaces, nothing conflicts.

The worker case is the sharp one: the descriptor is fetched once (after `AcceptInvalidationMessages()`, so fresh then), the tree can be evicted while the worker keeps applying, and the next `btree_ctid_update_if_needed()` writes into whatever tree holds that page now. No assertion covers it.

## Consequence for the pool-filling problem

That heavyweight lock is also why a tree stays in the pool while a session merely has the table open, so it is a thing we want to remove. Removing it turns every row of the table above into a primary bug as well. **Group B has to be fixed first** — the order matters.

(Removing it was attempted in #1195 and reverted for a separate reason: `orioledb_tbl_check` then reported leaked extents, suggesting the lock also excludes writers from eviction's seq-buf and free-extent finalisation.)

## Fixing

- **Reads of a single word** can use validate → read → re-validate, retrying on a mismatch; the page's change count is bumped before it is reused, so the second look catches a theft.
- **The write** (`btree_ctid_update_if_needed`) and **the lock acquisition** (`free_extents.c:604`) cannot: by the time a re-check notices, the counter has been written or the lock taken inside somebody else's page. Those need real exclusion, or a home for the counter that survives eviction.
- Group A should simply lose its `o_btree_load_shmem()` calls, so that ownership of the problem is visible: the descent owns it, and nobody else should look like they are covered.

akorotkov:
Fixes #1133: `Assert(meta_page_get_num_seq_scans(desc->rootInfo.metaPageBlkno) == 0)` in `evict_btree()`, reported from a standby's bgwriter.
Fixes #1196: the general form of the same bug — every caller that reaches into a tree's shared state through a cached descriptor.

## What it is

A `BTreeDescr` is cached per backend and nothing keeps its tree loaded, so the clock sweep can evict it and hand the root and meta pages to somebody else while the descriptor goes on naming them. `find_page()` has always known that — it compares the root's change count on the way down and reloads when it does not match — but a sequential scan **registering itself** never asked.

That registration lives in the meta page and says "this scan is still reading the files of checkpoint N", which is what `can_use_checkpoint_extents()` reads. Counting it into a page the pool has already handed away, or into one on its way out, is the assertion in the report: `evict_btree()` reads that counter twice — once to decline the eviction, once seventy lines and a whole page write later to assert nobody appeared — and finds a scan that crept in between.

A **standby** is where it shows, because `get_evict_btree_locks()` cannot take the ordinary relation locks during recovery (`LockAcquireExtended()` refuses anything above `RowExclusiveLock` on a `LOCKTAG_RELATION` there), so nothing else stands between an eviction and a reader. On a primary the eviction path's `AccessExclusiveLock` hides the whole class: a tree whose table anyone has open is simply not evicted.

## The #1133 fix (commit 2)

1. **Give the count back to the page that took it.** Release read `BTREE_GET_META(desc)`, which after a reload is a *different* page: the count would come off a tree that never gave it, and the page that did would keep a phantom one for good. Latent until now, because the relation lock kept reloads from happening under a scan.
2. **Move the registration to the new incarnation**, since the question is asked of whatever meta page the tree has now.
3. **`evict_btree()` stops deciding the question**: the soft check and the assertion give way to the bargain `cleanup_btree()` already strikes — a meta page carrying registrations is marked `toBeFreedOnSeqScanRelease` rather than freed, and the last scan to leave frees it.
4. **`refind_page()` drops a hint from a replaced incarnation.** It descends from a hint the caller captured earlier and calls `o_btree_load_shmem()` after taking it; if that load brings in a different incarnation, the hint names a page the pool has handed to somebody else and the descent walks another tree.

## The general fix (commit 3): a fast-path pin

Descending is safe — `find_page()` checks every page by change count. The callers that take a word out of the meta page, or take an LWLock that lives *inside* it, have no such check, and `o_btree_load_shmem()` cannot help them: it says only that the tree exists, and that answer expires as it returns.

So give them a pin, in the shape of PostgreSQL's fast-path locks:

```
user:                             evictor:
  slot = metaPageBlkno              evictClaim = 1 (CAS)
  pg_memory_barrier()               pg_memory_barrier()
  read claim; check META_IS_OURS    scan all max_procs slots
  refused -> reload tree, retry     any slot names my page -> decline
```

Whichever went second sees the other, and it is the **user** that stands down — a reload costs it little, and an evictor has other pages to get on with. Entry point is `o_btree_load_shmem_pinned()` / `btree_unpin_meta_page()`.

Wired: `init_checkpoit_number()` (this is #1133's other half), `btree_ctid_update_if_needed()` (a CAS write a recovery worker can land on a foreign page), `add_free_extents_from_tmp()` and `try_to_punch_holes()` (both take LWLocks out of the meta page), the planner and parallel-worker estimates, and `bridge_ctid` at index rebuild.

**Not wired, because they already answer the question** — the table in #1196 is broader than it needed to be:

| site | why it is already fine |
|---|---|
| `check_btree_compression()` | takes the checkpoint flavour of the relation lock, which conflicts with the evictor's |
| `write_tree_pages()` | carries the root's change count into its walk and reloads on failure |
| `orioledb_test_corrupt_*_undo()` | `#ifdef IS_DEV`, under a relation lock, never in recovery |

One site #1196 **missed** and this covers: `try_to_punch_holes()`.

### Three things worth calling out in review

- **Eviction reaches back into the tree it is evicting.** `perform_page_io()` allocates disk space, which drains that same tree's free-extent and punch-holes files — so both critical-section sites are reachable from inside `evict_btree()`, where this backend already holds the claim. A claim we hold therefore counts as permission, or the retry loop waits for itself. This one is a hard hang, and **regress and isolation passed twice with it in**; only testgres catches it, because only testgres calls `orioledb_evict_pages()`.
- **The claim has to outlive `ppool_free_page()`**, or a reader can pin in the gap, pass the ownership check (the page has not been reused yet) and then have the page handed away under it. It cannot simply be written back afterwards either — by then the page may be another tree's and we would clear *their* claim — so it is dropped locally instead. Reuse re-runs `init_meta_page()`, which zeroes the flag.
- **The retry loop drains sync requests when the checkpointer is the one waiting.** The evictor being waited out can be blocked in `RegisterSyncRequest()` waiting for the checkpointer to drain; one side waits on a latch, the other on a lock, and nothing sees a cycle. Same idiom as `o_tables_rel_lock_extended()`. The sleep is skipped on the first refusal, which almost always just means a stale descriptor.

`BTreeMetaPage` is memory-only (it holds LWLocks), so `evictClaim` needs no `ORIOLEDB_PAGE_VERSION` bump.

## The pin belongs at the call site, not in `o_btree_load_shmem()`

Folding it into the shared function covers every caller in one stroke, which is tempting and wrong: reloading reserves pages, and the other callers are entitled to assume that reaching an already-loaded tree does not. With the check there, `test_two_concurrent_transactions_pool_exhaustion` — which exists to catch exactly this — hung for the full 30-minute cell timeout. That version is not in this PR. (I could not reproduce the hang locally, so the attribution rests on the mechanism, on it landing precisely in that test, and on eight recent main runs showing no such timeout — not on a local repro.)

`refind_page()` was the one caller that captured state before the load (point 4 above). Found as a crash in a standby's recovery worker: `Assert(pageOids.datoid == desc->oids.datoid)` at `page_contents.c:205`, the detector for exactly this. The other nineteen callers read `desc->rootInfo` after the load and are unaffected — audited individually.

## Also tried and reverted: removing the heavyweight lock

The ordinary `AccessExclusiveLock` in `get_evict_btree_locks()` is why a tree stays in the pool while any transaction merely has the table open (`AccessShareLock` lives to the end of a transaction). Removing it made `evict_stale_descr_test` report `Extent 0 1 is neither free or busy` — leaked extents. A/B is clean: with the removal it fails, without it passes.

Hypothesis, not proven: that lock also excluded **writers** (`RowExclusiveLock` conflicts with it), and eviction's seq-buf and free-extent finalisation is not ready for a concurrent modification. The pin is the prerequisite for that removal — it covers the readers — but the writer side still needs covering, so the removal is not in this PR.

## Commits

1. **the repro** — a stop event inside the gap plus `test/t/evict_seq_scan_race_test.py`. Red on that commit: 3 runs, 3 assertions, ~6 s each, with the bgwriter as the evictor in one run and a recovery worker in the others.
2. **the #1133 fix** — green.
3. **the pin** — the general statement of the same bug.

## Testing

Commit 2 in isolation: repro green (3/3), regresscheck 52/52.

Full branch:
- regresscheck 52/52, isolationcheck 39/39, testgres parts 1/2/3 all green
- the #1133 repro 5/5 against 4/4 on the pre-pin base
- the assert on taking an already-held pin slot never fired across the whole suite, which is the evidence that one slot per backend is enough
- `-Werror` non-cassert build clean; pgindent clean

### Known limitation

Pin-inside-pin is caught only by an `Assert`; in a build without cassert the inner unpin would clear the outer pin. Never observed, and the `normal` CI cell is the one without cassert. A depth counter would close it if reviewers would rather not rely on the assert.


akorotkov:
⚠️ Not ready — CI found a real crash, and it looks like mine.

`check (arm64, 18, gcc, pg_tests)` failed in the standby's recovery worker:

```
TRAP: failed Assert("O_PAGE_IS_LOCAL(blkno) || (pageOids.datoid == desc->oids.datoid && pageOids.relnode == desc->oids.relnode)"),
File: "src/btree/page_contents.c", Line: 205
background worker "orioledb recovery worker" (PID 36621) was terminated by signal 6
PANIC: could not write page 1979 to file orioledb_data/1/20 with offset 0: No such file or directory
```

That assert is the detector for "a stale downlink onto an evicted-and-reused page, e.g. on a hot standby" — precisely the area this PR changes. It fired during CIC replay, right after two deferred BUILDING index builds and two "parallel index build DSM segment no longer mapped" messages.

All twelve pg_tests cells are green on main, and only this one cell failed here, so I am treating it as mine until shown otherwise rather than as a flake. Diagnosing; the prime suspect is `btree_seq_scans_reregister()` running inside `o_btree_load_shmem()`, where it can reach `ppool_free_page()` at a point callers never used to expect one.

Please hold off reviewing until this is resolved.

akorotkov:
Reworked after two CI findings — the earlier ⚠️ is resolved, and the description above now matches what is in the branch. Ready for review once this round of CI is green.

akorotkov:
Filed #1196 for the wider pattern this PR touches one corner of: about a dozen callers read or write a tree's shared state through a cached descriptor, protected on a primary only by the eviction lock and not at all during recovery. Notably it also says why that lock cannot be removed for the pool-filling problem until those are fixed.

akorotkov:
Third CI finding, and this one was the test, not the code.

`check (arm64, 17, gcc, valgrind_2)` died on the 150-minute cell timeout with my test hung. The reset of the stop event sat on the success path: if no eviction reached the gap inside the deadline — which under valgrind, twenty times slower, is the likely run — the test raised before disarming, the next process to reach the event parked for good, and the node never shut down.

That is exactly the defect I fixed in #1172 and then wrote again here. The test now holds the event in a `_armed()` context manager whose `finally` always resets it, waits 300 s rather than 120, and **skips** rather than fails when nothing reaches the gap — an eviction that never came tested nothing, and a green tick would be a lie about that.

The reworked test is in the first commit, so the red/green split still measures the fix and not the harness: repro commit fails, fix commit passes, verified again after the rework. The C side is byte-identical to the previous push.

akorotkov:
It hung under valgrind a second time, at the same place, with the stop event already guaranteed to be reset — so the disarm was not the whole story, and I have not diagnosed the rest.

Rather than keep spending 150-minute cells guessing, the test is now skipped under valgrind the way the s3 suite already is, with a TODO saying plainly that this is unfinished:

```python
# TODO: find out what this waits on under valgrind and re-enable it there.
@unittest.skipIf(os.environ.get('USE_VALGRIND') == '1',
                 'standby eviction race does not line up under valgrind')
```

The test needs three timing-dependent things to coincide — a standby whose pool is starved by replay, an eviction of one particular root, and a reader getting its cursor open while that eviction is parked — and in a cell where everything runs twenty times slower they stop coinciding. It takes about six seconds in every other cell, and that is where it guards the fix: it still fails on the repro commit and passes on the fix commit, re-verified.

This is a workaround, not an explanation. If you would rather I chase the valgrind hang properly before this merges, say so and I will.

akorotkov:
Correction, with the evidence I should have fetched before calling it a flaky test: **the hang is not this test's**, and it is not a valgrind artefact either. Filed as #1197.

The replica's own log — from the run's uploaded artifact, which I had not looked at — shows the test doing exactly what it is meant to and then the server failing to stop:

```
15:11:53.133  statement: SELECT pg_stopevent_reset('after_tree_root_page_write');
15:12:02.432  received fast shutdown request
15:12:02.489  [41478] LOG: recovery on exit: 2
15:12:02.820  background worker "orioledb recovery worker" (PID 41478) exited with exit code 1
                                  <- log ends
```

Park, scan, disarm and a shutdown request, all inside 27 seconds even under valgrind. Then the standby never finishes shutting down: 1h44m later its startup process is still in `shm_mq_send_bytes(nowait=false)` → `worker_queue_flush(worker_id=2)` → `orioledb_redo` → `PerformWalRecovery`, blocked on the queue of the worker that exited at 15:12:02.820. With `PGCTLTIMEOUT=3000` under valgrind that is up to 50 minutes per node, which is where the cell went.

The skip stays for now, but its comment no longer blames the test — it points at #1197.

Two harness gaps this exposed, both noted in that issue: `ci/list_stuck.sh` only tails a server log for data directories under `/tmp/`, so testgres hangs arrive with no PostgreSQL log; and in the first of the two runs its `pkill` raced the vgdb capture, so that run produced no stacks at all — which is why I spent a round guessing instead of reading.

Issue #1145, opened 2026-09-10
promotion_xid_horizon_switchover test stucked


homper:
https://github.com/orioledb/orioledb/actions/runs/34401734534/job/102635135611?pr=1140

Could be just a weird test

[Stucked jobs from run with backtraces](https://github.com/user-attachments/files/32054836/orioledb_stuck_stack.txt)

akorotkov:
Recurred, and this time the logs say exactly what is wrong. It is not a weird test — the standby has replayed **past** the point at which the new timeline forked, so it can never follow it.

Seen on https://github.com/orioledb/orioledb/actions/runs/35523937655/job/106112729094 (`check (amd64, 18, clang, check_page)`), on a PR that touches nothing in this area. The `Check` step died on its 30-minute timeout; the test itself had been stuck for 27 of those minutes.

Standby log, repeating every five seconds until the job was killed:

```
LOG:  restarted WAL streaming at 0/3000000 on timeline 3
LOG:  replication terminated by primary server
DETAIL:  End of WAL reached on timeline 3 at 0/3000268.
LOG:  new timeline 4 forked off current database system timeline 3 before
      current recovery point 0/3000350
LOG:  waiting for WAL to become available at 0/3000368
```

`0/3000350 > 0/3000268`: the standby's recovery point is **ahead of the fork**. PostgreSQL will not rewind, so this is permanent — the standby waits for WAL that will never exist on timeline 4.

Meanwhile the test polls forever:

```
statement: select pg_catalog.pg_last_wal_replay_lsn() >= '0/3000470'::pg_lsn
```

once a second, for the whole 27 minutes. Everything else in the dump is idle: the startup process sits in `WaitForWALToBecomeAvailable`, the walreceiver and walsender are idle, no process is spinning.

So there are two things here, and the second is the one that costs CI time:

1. **The divergence itself.** The old primary was promoted at an LSN below what the standby had already replayed. Worth understanding on its own — it is the shape of a lost-write window, not merely a test artefact.
2. **The test cannot fail.** `pg_last_wal_replay_lsn() >= …` is polled with no deadline, so a standby that is permanently stuck burns the cell's entire budget instead of failing in seconds with a legible message. A bounded wait would turn a 30-minute timeout into a real failure that names the problem — worth doing regardless of (1).

Reopening: the original close was "could be just a weird test", and that reading does not survive the log above.

🤖 Generated with [Claude Code](https://claude.com/claude-code)

PR #1198, opened 2026-09-20
Do not wait for ever on a recovery worker that may be gone (#1197)

### LINKED ISSUE #1197: Standby never finishes a fast shutdown: startup process stuck in shm_mq_send() to an exited recovery worker
Seen twice on valgrind cells of #1195, each time costing the cell its full 150-minute timeout. It is not the test that hangs — the test finishes its work in 27 seconds — it is the standby that never stops.

## What the logs show

Replica log (`evict_seq_scan_race-..._tgsb/logs/postgresql.log`), run 35448498721, `check (arm64, 18, gcc, valgrind_2)`:

```
15:11:36.153  orioledb recovery started.
15:11:36.233  orioledb recovery worker 2 started.          [pid 41478]
...
15:11:53.133  statement: SELECT pg_stopevent_reset('after_tree_root_page_write');
15:12:02.432  received fast shutdown request
15:12:02.488  [41478] ERROR: terminating orioledb worker due to administrator command
15:12:02.489  [41478] LOG:  recovery on exit: 2
15:12:02.820  background worker "orioledb recovery worker" (PID 41478) exited with exit code 1
15:12:03.016  background worker "orioledb recovery worker" (PID 41479) exited with exit code 1
                                    <- log ends here
```

All six recovery workers are terminated and exit. The log stops. The postmaster never reports shutting down.

One hour forty-four minutes later, the harness dumps stacks through vgdb, and the startup process is here:

```
#5  shm_mq_send_bytes (mqh=0x73c5ff8, nbytes=7920, nowait=false)
#6  shm_mq_sendv
#7  shm_mq_send
#8  worker_queue_flush (worker_id=2)              src/recovery/recovery.c:5962
#9  worker_send_modify (worker_id=2, recType=RecoveryMsgTypeUpdate)
#10 spread_idx_modify
#11 replay_on_record                              src/recovery/recovery.c:5371
#12 wal_parse_container                           src/recovery/wal_reader.c:785
#13 replay_container / orioledb_redo              src/recovery/recovery.c:1426
#15 ApplyWalRecord
#16 PerformWalRecovery                            xlogrecovery.c:1850
```

It is blocking on the queue of worker 2 — the worker that exited at 15:12:02.820. Every other process in the dump (16 of them, master and replica) is idle in a latch wait; the orioledb bgwriter is in its ordinary 200 ms loop. The master finished its own work and went on checkpointing normally.

## Why that should not happen

`worker_queue_flush()` handles a detached counterparty correctly:

```c
result = shm_mq_send(state->queue, state->queue_buf_len, state->queue_buf, false, true);
if (result == SHM_MQ_DETACHED) { unexpected_worker_detach = true; return; }
```

and the worker's `PG_CATCH` detaches on the way out:

```c
if (recovery_worker_queue != NULL)
    shm_mq_detach(recovery_worker_queue);
```

So a worker dying by ERROR should release a sender blocked on its queue. Something in that handshake did not happen for worker 2, and the sender waits forever. I have not established which half failed — the dump carries stacks but not `mq_detached`.

## Why it costs so much

`PGCTLTIMEOUT=3000` under valgrind (Makefile:401), so a node that will not stop costs up to 50 minutes, twice for a master/standby pair. That is where the 150-minute cell goes.

## Two things that made this hard to see

- `ci/list_stuck.sh` tails the server log only when the data directory matches `-D /tmp/([a-z0-9_]+)/`. testgres puts its data directories under `test/tmp_check_t/…`, so for any testgres test the hang dump arrives with no PostgreSQL log at all. The logs above came from the uploaded artifact instead.
- In the first of the two runs the vgdb dump produced no frames: `list_stuck.sh` line 64-66 `pkill -KILL tail/gdb/postgres` raced the capture, and the job log has only gdb's connection preamble.

Both are worth fixing on their own; the second one cost a full round of guessing.

akorotkov:
Fixes #1197 — a standby that never finishes a fast shutdown, twice costing a valgrind cell its full 150-minute timeout.

## What happens

The test finishes its work in 27 seconds; it is the node that will not stop. The startup process is here:

```
shm_mq_send_bytes (nbytes=7920, nowait=false)
shm_mq_send
worker_queue_flush (worker_id = 2)
worker_send_modify
spread_idx_modify
replay_on_record
...
PerformWalRecovery
```

blocked on the queue of the worker the postmaster logged as exited half a second into the shutdown. Every other process in the dump is idle.

## What this changes

`worker_queue_flush()` waited with `nowait = false`, so shm_mq's detach handshake was the *only* thing that could ever release it. Whatever kept that handshake from arriving, the sender waits for ever — and on a standby the sender is the startup process, which does not act on a shutdown request from `CHECK_FOR_INTERRUPTS()`. It acts on it in the redo loop it is no longer in. So nothing can stop the node.

Wait in steps instead, and at each step ask the question **the loop immediately above this one in the same file already asks** while waiting for a worker to catch up: is the worker still there? `GetBackgroundWorkerPid()` answers it without depending on the queue handshake at all. That existing polling loop is itself the evidence that this path was already known not to be trustworthy on its own.

The wait stays a latch wait, which the reader sets on every read, so ordinary backpressure costs no more than before; the 100 ms timeout only decides how soon a dead worker is noticed.

## What this does *not* do — please read before merging

**It does not explain why the detach did not arrive, and is not meant to.** It makes the sender's progress independent of that question.

What I did rule out, by reading:

- the queues are created once at shmem init and never re-created, so `mq_detached` is not being cleared underneath;
- `"recovery on exit: 2"` really is the worker id, so worker 2 is the process that exited — and it went through the `PG_CATCH` that calls `shm_mq_detach()`;
- the two worker pools use disjoint index ranges, so this is not a mixed-up id.

**I could not reproduce the hang locally** — it needs a recovery worker as the evictor, which I have not been able to arrange. So this change is verified only in the sense that it breaks nothing (see Testing); it is *not* verified against the failure it addresses. The valgrind cells are where #1197 actually appeared, which is the main reason this is a PR rather than a local branch.

The second commit is about that: the next occurrence should be able to answer the question this one leaves open. `ci/list_stuck.sh` now finds the server log under the testgres layout (previously any testgres hang dumped with no PostgreSQL log at all), waits for the vgdb captures instead of racing them with `pkill` (the first of the two #1197 runs produced nothing but gdb's connection preamble), and prints the `shm_mq` the sender is blocked on.

## Testing

regresscheck 52/52; testgres part 1 (replication, recovery, recovery workers) and part 3 green. Nothing here is covered by a new test — the behaviour it changes only shows up when a worker dies while a flush is in flight.


akorotkov:
Pushed a fix for the PG16 build: `WAIT_EVENT_MESSAGE_QUEUE_SEND` is the PG17+ name, and PG16 calls it `WAIT_EVENT_MQ_SEND` (PostgreSQL 17 generated the wait-event names and renamed this one). That single identifier is what failed **every** PG16 cell — `check`, `static`, `sqlsmith`, the docker build and all of `pg_upgrade`, which builds 16. Folded into the first commit so no broken commit is left in the history, and verified by compiling `recovery.c` against the real PG16 headers rather than by inspection.

The one non-PG16 failure, `check (amd64, 18, clang, check_page)`, is **not from this PR**: it is #1145, which I have reopened with the root cause. The standby had replayed past the point where the new timeline forked (`new timeline 4 forked off ... before current recovery point 0/3000350`, fork at `0/3000268`), so it can never follow it, and the test polls `pg_last_wal_replay_lsn()` with no deadline. The startup process in that dump is idle in `WaitForWALToBecomeAvailable`, not in the queue send this PR touches.

Incidentally the second commit earned its keep straight away: the server log in that dump was found by the new testgres-aware lookup. The old script would have printed nothing.

pashkinelfe (approved):

Issue #90, opened 2022-10-03
NO ERROR:  could not serialize access due to read/write dependencies among transactions


Tanyaevm:
sql/test.sql:

**Error with orioledb:** test stuck
```sql
CREATE TABLE o_test_1 (val_1 VARCHAR(10))USING orioledb;

INSERT INTO o_test_1 VALUES ('aaa');

SELECT gid FROM pg_prepared_xacts;

BEGIN TRANSACTION ISOLATION LEVEL SERIALIZABLE;
INSERT INTO o_test_1 VALUES ('ddd');

PREPARE TRANSACTION 'foo2';

COMMIT PREPARED 'foo2';

BEGIN TRANSACTION ISOLATION LEVEL SERIALIZABLE;

UPDATE o_test_1 SET val_1 = 'eee' WHERE val_1 = 'ddd';

PREPARE TRANSACTION 'foo4';

SELECT gid FROM pg_prepared_xacts;

BEGIN TRANSACTION ISOLATION LEVEL SERIALIZABLE;
SELECT * FROM o_test_1;

INSERT INTO o_test_1 VALUES ('fff');
PREPARE TRANSACTION 'foo5';

SELECT gid FROM pg_prepared_xacts;

ROLLBACK PREPARED 'foo4';

SELECT gid FROM pg_prepared_xacts;

DROP TABLE o_test_1;
```

**Error for postgres:**
```sql
CREATE TABLE o_test_1 (val_1 VARCHAR(10));--USING orioledb;

INSERT INTO o_test_1 VALUES ('aaa');

SELECT gid FROM pg_prepared_xacts;

BEGIN TRANSACTION ISOLATION LEVEL SERIALIZABLE;
INSERT INTO o_test_1 VALUES ('ddd');

PREPARE TRANSACTION 'foo2';

COMMIT PREPARED 'foo2';

BEGIN TRANSACTION ISOLATION LEVEL SERIALIZABLE;

UPDATE o_test_1 SET val_1 = 'eee' WHERE val_1 = 'ddd';

PREPARE TRANSACTION 'foo4';

SELECT gid FROM pg_prepared_xacts;

BEGIN TRANSACTION ISOLATION LEVEL SERIALIZABLE;
SELECT * FROM o_test_1;

INSERT INTO o_test_1 VALUES ('fff');

--ERROR:  could not serialize access due to read/write dependencies among transactions
--DETAIL:  Reason code: Canceled on identification as a pivot, during write.
[...]

PR #1208, opened 2026-09-21
Resolve a key's collation for the checkpointer too

### LINKED ISSUE #1193: Checkpointer aborts comparing an ICU-collated index key while writing a tree
Split out of ORI-328. The recovery half of that ticket is fixed and merged (#1181); this is the second site with the same root cause, which #1181 deliberately left alone because closing it needs a PostgreSQL-side change.

## Symptom

Writing a tree walks it level by level, and above the leaves that means comparing keys (`checkpoint_internal_pass()` → `o_btree_cmp()`). For a key with a collation of its own the comparison asks `pg_locale.c` for that collation, which reads `pg_collation` the first time a process uses it. The checkpointer has no transaction to read it in, so on an assertion-enabled build **every** checkpoint of such a tree aborts the checkpointer:

```
TRAP: failed Assert("IsTransactionState()"), File: "relcache.c"
#2  pg_newlocale_from_collation
#3  varstr_cmp / bttextcmp / FunctionCall2Coll
#7  o_call_comparator
#11 checkpoint_internal_pass
#21 CheckpointerMain
```

Without assertions the lookup is served, but from the middle of a tree walk — which is exactly the place #1181 stopped reading a system-tree page from, since doing so can reserve a pool page under a held page lock.

## Reproduction

An ICU-collated text key and enough rows that the tree has internal pages; a one-page tree is written without comparing anything.

```sql
CREATE TABLE o_coll (
  k text COLLATE "unicode" NOT NULL,
  v int NOT NULL,
  PRIMARY KEY (k)
) USING orioledb;
INSERT INTO o_coll SELECT md5(g::text), g FROM generate_series(1, 2000) g;
CHECKPOINT;   -- checkpointer aborts on an --enable-cassert build
```

## Why #1181 could not cover it

#1181 resolves a key's collation once, while the comparator is built, so no catalog lookup can happen later under a page lock. The same call cannot be made for the checkpointer, and `o_resolve_collation()` says so at the point where it gives up:

```c
/*
 * pg_newlocale_from_collation() insists on a transaction to read
 * pg_collation in, and makes an exception only for recovery.  The
 * checkpointer therefore cannot be warmed up here even though its
 * comparisons have the same problem; it stays as it was.
 */
if (!IsTransactionState() && !RecoveryInProgress())
    return;
```

The lookup itself was never the problem — `o_set_syscache_hooks()` is already in force there, so OrioleDB answers it from its own system trees and no relation is ever opened. What is missing is permission to make the lookup, and a safe moment to make it in.

## Fix, and what blocks it

Two halves, both written and tested locally but not proposed yet:

1. **PostgreSQL fork** — `Let an extension answer the collation lookup without a transaction`: `pg_newlocale_from_collation()` exempts only recovery from its transaction assertion; extend the exemption to a lookup served through `SearchCatCacheInternal_hook`, where no relation is opened. Branch `akorotkov/collation-catcache-hook-assert`, commit `386e6ac2e0b`.
2. **OrioleDB** — resolve the collation while the comparator is built for the checkpointer too, as recovery now does, and let it in. Branch `akorotkov/ori-276`-era work sits on `akorotkov/ori-328-checkpointer-collation`, commit `a296bd6c8`, with `test/t/checkpoint_collation_test.py`.

The OrioleDB half depends on the fork half, so the ordering question is really about the fork change — whether exempting a hook-served lookup from the transaction assertion is the shape we want, or whether the checkpointer should instead be given a transaction (or the collation pre-resolved somewhere a transaction exists).

## Notes

- The local OrioleDB branch predates a fair amount of `main` and needs a rebase before it becomes a PR.
- Scope is wider than the checkpointer in principle: any process that compares collatable keys without a transaction and without `RecoveryInProgress()` hits the same wall. The checkpointer is the one reproduced.
- Related: ORI-328 (recovery half, fixed in #1181).

akorotkov:
Fixes #1193.

## What was wrong

Writing a tree walks it level by level, and above the leaves that means comparing keys (`checkpoint_internal_pass()` → `o_btree_cmp()`). For a key with a collation of its own the comparison asks `pg_locale.c` for that collation, which reads `pg_collation` the first time a process uses it. The checkpointer has no transaction to read it in, so on an assertion-enabled build every checkpoint of such a tree aborts the checkpointer:

```
TRAP: failed Assert("IsTransactionState()"), File: "relcache.c"
#2  pg_newlocale_from_collation
#3  varstr_cmp / bttextcmp / FunctionCall2Coll
#7  o_call_comparator
#11 checkpoint_internal_pass
#21 CheckpointerMain
```

Two statements reproduce it: an ICU-collated text key and enough rows for internal pages.

Without assertions the lookup is served, but from the middle of a tree walk — and the page **is** locked there (`checkpoint_fix_split_and_lock_page()` runs immediately before the call), which is exactly the situation #1181 stopped reading a system-tree page from, since doing so can reserve a pool page under a held page lock. So this is not a cassert-only problem.

## The fix

The lookup itself was never the problem — `o_set_syscache_hooks()` is already in force there, so OrioleDB answers it from its own system trees and no relation is ever opened. What was missing is permission to make the lookup and a safe moment to make it in. So resolve the collation while the comparator is built, as recovery already does since #1181, and let the checkpointer in.

The safe moment exists: the checkpointer builds its comparators in `checkpoint_tables_callback()` → `o_fetch_index_descr()`, before the walk, with nothing locked. Verified from the abort stack of the intermediate state.

## Why not keep it all on our side

Measured, not assumed. The obvious alternative is to give the checkpointer a transaction:

1. `StartTransactionCommand()` trips `Assert("s->prevSecContext == 0")` — our own syscache hooks, which must stay in force during the lookup, set `SECURITY_LOCAL_USERID_CHANGE`.
2. With that masked it trips `Assert("MyProc->vxid.procNumber == vxid.procNumber")` in `VirtualXactLockTableInsert()`. `InitAuxiliaryProcess()` sets `vxid.procNumber = INVALID_PROC_NUMBER` on purpose: auxiliary processes have no virtual xid.

Pre-warming the locale cache cannot work either: on 18 the assertion sits **before** the cache check, so every comparison aborts, not just the first.

## Testing

Built PostgreSQL 18.6 + orioledb/postgres#90 and ran this branch against it:

| PostgreSQL | result |
|---|---|
| with orioledb/postgres#90 | `checkpoint_collation_test` passes; regress 52/52, isolation 39/39 |
| #90 reverted | fails — the abort moves from the tree walk to `o_resolve_collation()` at comparator build, exactly as expected |

## The PostgreSQL half is in

- [x] orioledb/postgres#90 (patches18)
- [x] orioledb/postgres#91 (patches17) — the assertion lives in `lookup_collation_cache()` there
- [x] orioledb/postgres#92 (patches16) — same
- [ ] swap the commit hashes in `.pgtags` for patchset tags once those are cut — not needed for CI, which takes a commit as readily as a tag

`.pgtags` points at the three merge commits (`7d0ad66c7` / `e3c4c4324` / `eb3aa7061`). The first CI run, against the old patchset, failed 30 cells in three test files — the new one, `recovery_collation_test`, and the `COLLATE "POSIX"` temp table in `replication_test` — all of them checkpointing a table with a non-default collation and all dying the same way, `TRAP: failed Assert("IsTransactionState()")`. With the patchset pinned here they pass.

One test comes back with this: `collate_test.py` skipped `test_all_databases_need_one_encoding` with the reason *"lookup_collation_cache() asserts before the syscache hook"* — this ticket's assertion, worked around by switching the test off. The skip is gone and the module is 9 for 9.


pashkinelfe (approved):
LGTM unless there is some reason not to update patchset at the moment

